PR #1, opened 2020-09-14
facilitator: generate sample ingestion data

tgeoghegan:
This PR adds a Rust target for the facilitator binary that will implement the map and reduce jobs for the facilitator server. At the moment, it only implements generation of sample ingestion data from random values.


tgeoghegan:
I have made more progress beyond what's in this PR, including substantial refactorings of the code in here, in the timg/ingestion-validation branch. To avoid redundant reviews and rebases, I'm going to rescind this PR and work with yuriks to review a larger changeset.

PR #2, opened 2020-09-18
Avro schema updates

tgeoghegan:
In today's eng sync meeting, we agreed on a few changes to the Avro
schema:

 - remove signature_of_packets from signature file and instead store a
   SHA-256 hash of the .avro file in the corresponding header, to
   preclude mix and match attacks.
 - remove ingestion-signature.avsc and store signature bytes directly into a file.
 - add invalid-packet.avsc to represent packets that failed validation
 - rename 'aggregaton-share.avsc' to 'sum-part.avsc' and digest of invalid packets file


avro-schema/batch-signature.avsc
```diff
@@ -0,0 +1,12 @@
+{
```
gurayAlsac, line 1: I suppose there isn't a lot of structure to the batch header signature. Is it simpler to have it as a raw signature file?

tgeoghegan: That's a good idea and would save us some serialization ceremony. We could just store the signature bytes in the .sig files.

yuriks: Since we're now only signing one file, that seems like a fine idea to me too.

avro-schema/sum-part.avsc
```diff
@@ -61,6 +61,17 @@
             "type": "long",
             "logicalType": "timestamp-millis",
             "doc": "time range information for the shares in this batch."
+        },
+        {
+            "name": "invalid_packet_uuids",
+            "type": {
+                "type": "array",
```
yuriks, line 68: As we (me and @tgeoghegan) talked in Slack, my concern with this is that it requires the entire list of uuids to be parsed whenever you want to read this header, it also prevents writing or reading this file effectively in a streaming way, since the entire message needs to be (de)serialized in memory at once. This is fine if we don't have many invalid UUIDs (and I don't think we expect that), however, it is a potential DoS vector (especially for the N-hourly reduction steps, which might be reading many of these files in serial without parallelization).

If we do it like this, then I think we should agree on truncating the number of items in this list to some maximum value. We can still keep track of the total number of failures in an overflow counter. Afaik this list is mostly intended as an eventual debugging aid?

@winstrom since you were the one who originally proposed this in the schema doc

yuriks (changes requested):

avro-schema/sum-part.avsc
tgeoghegan: We could also put the invalid UUIDs back into a separate Avro encoded file (which is how it was laid out in the IDL doc) and instead put `invalid_packet_uuids_file_digest` here.

tlepoint (approved):
How do we want to specify which public key was used when generating the signature, as it may happen that during a rotation the ingestion service issues some batches using the old key if is has not been updated yet?

avro-schema/sum-part.avsc
```diff
@@ -43,12 +48,12 @@
             "doc": "If specified, the hamming weight of the vector will be verified during the validity check on the server."
         },
         {
-            "name": "sums",
+            "name": "sum",
```
manu-raghunath, line 51: The cardinality of uuids and sum have to be same, is it not ?

avro-schema/ingestion-header.avsc
```diff
@@ -52,6 +52,11 @@
             "type": "long",
             "logicalType": "timestamp-millis",
             "doc": "time range information for the shares in this batch."
+        },
+        {
+            "name": "packet_file_digest",
+            "type": "bytes",
+            "doc": "SHA-256 digest of the .avro file containing packets in this batch."
```
manu-raghunath, line 59: As pointed out in another comment, should there be a field to identify the key used to sign the header?

avro-schema/sum-part.avsc
manu-raghunath: Should this an array of (sub) records [(uuid, sum)] ?

tlepoint: We actually want to compute only one "sum" over many `batch_uuid`'s, and this is important for the privacy story of Prio. In other words, we want to sum the `sum`'s of your subrecords and only reveal the total.

avro-schema/sum-part.avsc
tlepoint: Having the invalid packets in another file, hashed in the header, is a good design decision 👍  Thanks.

avro-schema/sum-part.avsc
manu-raghunath: Then should this be an array? The change from sums to sum seemed to indicate what you mentioned but the schema still refers to an array?

avro-schema/sum-part.avsc
tgeoghegan: I think that's a good idea, but I have a few more questions about error handling at all stages of the pipeline, which I am going to ask in an email to prio-implementers today in anticipation of resolving them during Thursday's meeting. We can implement whatever schema changes result from that conversation in a subsequent PR.

avro-schema/sum-part.avsc
tgeoghegan: @yuriks and I also found this confusing when we hashed it out on Friday afternoon. Here's how I think about it: the basic atom of data produced by any client in this system is a vector of _bins_ fields (_bins_ is also _dim_ or _dimension_ in some contexts). When we sum up data, the result is still a _bins_-dimensional vector. I changed this to `sum`, singular to indicate it's a single vector, but by dint of being a vector, it's still an array of longs.

avro-schema/ingestion-header.avsc
tgeoghegan: > How do we want to specify which public key was used when generating the signature, as it may happen that during a rotation the ingestion service issues some batches using the old key if is has not been updated yet?

Not a bad idea. What I wrote up in https://docs.google.com/document/d/1MdfM3QT63ISU70l63bwzTrxr93Z7Tv7EDjLfammzo6Q/edit#heading=h.4jef7v1c76vl is that clients should attempt to use all the keys listed in the peer's keyfile, but including a key fingerprint of some sort is better. We haven't yet settled on a plan of record for key exchanges so I think what I'd like to do is incorporate this suggestion into the design doc and then implement the corresponding schema changes once we've all agreed on how this will work. WDYT @manu-raghunath and @tlepoint?

avro-schema/sum-part.avsc
manu-raghunath: Thanks for the clarification and the addition in the doc field for the same.

avro-schema/ingestion-header.avsc
manu-raghunath: Sounds good.

avro-schema/sum-part.avsc
```diff
@@ -4,10 +4,15 @@
...
-            "doc": "Universal unique identifier to link with data share batch sent to other server(s) participating in the aggregation."
+            "name": "batch_uuids",
+            "type": {
+                "type": "array",
+                "items": {
+                    "type": "string",
+                    "logicalType": "uuid"
+                }
+            },
+            "doc": "Universal unique identifiers of data share batches included in this sum."
```
yuriks, line 15: Abbreviate to "UUID" for consistency with other comments and to make it easier to read.

yuriks (approved):
Wording suggestion, but LGTM

Issue #25, opened 2020-09-24
Include key identifier in signed headers


tgeoghegan:
Reviewers of #2 pointed out we should include an explicit key identifier in headers so that message recipients can gracefully handle key rotations instead of having to try all the keys listed in a peer's manifest.

tgeoghegan:
The annoying thing about this is that the header of a batch is the wrong place to put the key-identifier: one shouldn't try to even parse the header unless its signature is valid, much less read a value out of it. We got rid of the Avro structure for signatures because eventually that object contained nothing but a `bytes` of the signature, but now it'd be nice to be able to put a `string` for `key_identifier` next to it.

tgeoghegan:
Resolved in #67

PR #34, opened 2020-09-29
avro: Add identifiers

tlepoint:
- Add the key identifiers to the avro schemas.
- Make the description of the digest more specific in the sum-part.

Addresses #25.


tlepoint:
This will require more modifications. Closing for now.

Issue #19, opened 2020-09-21
Facilitator should read and write objects from S3


tgeoghegan:
In order to read and write messages from S3, the facilitator needs to implement an `S3Transport` alongside its existing `FileTransport`. [Rusoto](https://github.com/rusoto/rusoto) appears to be the crate of choice for working with AWS APIs.

tgeoghegan:
If we were in AWS, then I think we'd get suitable creds injected into our running environment by ~* cloud magic *~ but as POR is to deploy the facilitator to GCP, so we have to figure out how to make the necessary AWS credential(s) available to pods running in GCP.

tgeoghegan:
We still need #45 for this.

tgeoghegan:
#45 got merged. The workflow manager needs to be able to list buckets and delete content, but that's tracked in #13. Closing.

Issue #28, opened 2020-09-28
facilitator: std::path::Path is wrong type for Transport::{get, put}


tgeoghegan:
In #4, @yuriks notes:

> I think using `Path` here is actually not the right call: `Path` inherently deals with OS-native paths. While that's the natural key for `FileTransport`, it doesn't really apply for something like S3. If I run this on Windows, for example, I actually still want to keep using `/` as the separator when I upload to S3, not `\`. So I think the right way of doing this is to have key be a generic path value (can either just use `str` with `/` as the separator, or create your own `newtype` over it) and then inside `FileTransport` you can parse it and re-convert to a `Path` when accessing the `filesystem`.

tgeoghegan:
Fixed in #36

PR #36, opened 2020-10-01
facilitator: Add S3Transport

tgeoghegan:
Adds a Transport implementation which reads objects from and writes
objects to S3. The various subcommands on the facilitator tool can now
take arguments like `s3://bucket-name` as well as local paths for input
and output storage.

Addresses #19

### DIFF
--- a/facilitator/Cargo.toml
+++ b/facilitator/Cargo.toml
@@ -12,12 +12,13 @@ base64 = "0.12.3"
 chrono = "0.4"
 clap = "2.33.3"
 derivative = "2.1.1"
+hyper = "0.13.8"
+hyper-rustls = "0.21.0"
 prio = "0.2"
 rand = "0.7"
 ring = { version = "0.16.15", features = ["std"] }
-rusoto_core = { version="0.45.0", default_features=false, features=["rustls"] }
-rusoto_mock = { version="0.45.0", default_features=false, features=["rustls"] }
-rusoto_s3 = { version="0.45.0", default_features=false, features=["rustls"] }
+rusoto_core = { version = "0.45.0", default_features = false, features = ["rustls"] }
+rusoto_s3 = { version = "0.45.0", default_features = false, features = ["rustls"] }
 serde = { version = "1.0", features = ["derive"] }
 tempfile = "3.1.0"
 thiserror = "1.0"
@@ -26,3 +27,6 @@ uuid = { version = "0.8", features = ["serde", "v4"] }
 
 [build-dependencies]
 vergen = "3"
+
+[dev-dependencies]
+rusoto_mock = { version = "0.45.0", default_features = false, features = ["rustls"] }
--- a/facilitator/Cargo.toml
+++ b/facilitator/Cargo.toml
@@ -11,12 +11,17 @@ avro-rs = "0.11.0"
 base64 = "0.12.3"
 chrono = "0.4"
 clap = "2.33.3"
+derivative = "2.1.1"
 prio = "0.2"
 rand = "0.7"
 ring = { version = "0.16.15", features = ["std"] }
+rusoto_core = { version="0.45.0", default_features=false, features=["rustls"] }
+rusoto_mock = { version="0.45.0", default_features=false, features=["rustls"] }
+rusoto_s3 = { version="0.45.0", default_features=false, features=["rustls"] }
 serde = { version = "1.0", features = ["derive"] }
 tempfile = "3.1.0"
 thiserror = "1.0"
+tokio = { version = "0.2", features = ["rt-core", "io-util"] }
 uuid = { version = "0.8", features = ["serde", "v4"] }
 
 [build-dependencies]
--- a/facilitator/src/bin/facilitator.rs
+++ b/facilitator/src/bin/facilitator.rs
@@ -1,5 +1,3 @@
-use std::{path::Path, str::FromStr};
-
 use anyhow::{anyhow, Context};
 use chrono::{prelude::Utc, NaiveDateTime};
 use clap::{App, Arg, ArgMatches, SubCommand};
@@ -8,6 +6,8 @@ use ring::signature::{
     EcdsaKeyPair, KeyPair, UnparsedPublicKey, ECDSA_P256_SHA256_FIXED,
     ECDSA_P256_SHA256_FIXED_SIGNING,
 };
+use rusoto_core::Region;
+use std::{path::Path, str::FromStr};
 use uuid::Uuid;
 
 use facilitator::{
@@ -19,7 +19,7 @@ use facilitator::{
         DEFAULT_INGESTOR_PRIVATE_KEY, DEFAULT_PHA_ECIES_PRIVATE_KEY,
         DEFAULT_PHA_SIGNING_PRIVATE_KEY,
     },
-    transport::LocalFileTransport,
+    transport::{LocalFileTransport, S3Transport, Transport},
     DATE_FORMAT,
 };
 
@@ -68,8 +68,10 @@ fn main() -> Result<(), anyhow::Error> {
                         .value_name("DIR")
                         .default_value(".")
                         .help(
-                            "Directory to write sample data for the PHA (aka \
-                            first server) into",
+                            "Storage to write sample data for the PHA \
+                            (aka first server) into. May be either a local \
+                            filesystem path or an S3 bucket, formatted as \
+                            \"s3://bucket-name\"",
                         ),
                 )
                 .arg(
@@ -78,8 +80,10 @@ fn main() -> Result<(), anyhow::Error> {
                         .value_name("DIR")
                         .default_value(".")
                         .help(
-                            "Directory to write sample data for the \
-                            facilitator (aka second server) into",
+                            "Storage to write sample data for the \
+                            facilitator (aka second server) into. May be \
+                            either a local filesystem path or an S3 bucket, \
+                            formatted as \"s3://bucket-name\"",
                         ),
                 )
                 .arg(
@@ -292,7 +296,11 @@ fn main() -> Result<(), anyhow::Error> {
                         .long("ingestion-bucket")
                         .value_name("DIR")
                         .default_value(".")
-                        .help("Directory containing ingestion data"),
+                        .help(
+                            "Directory containing ingestion data. May be \
+                            either a local filesystem path or an S3 bucket, \
+                            formatted as \"s3://bucket-name\"",
+                        ),
                 )
                 .arg(
                     Arg::with_name("validation-bucket")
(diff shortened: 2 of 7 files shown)


facilitator/src/lib.rs
```diff
@@ -56,6 +56,8 @@ pub enum Error {
     CryptographyUnspecifiedError(String, ring::error::Unspecified),
     PeerValidationError(String),
     ConversionError(String),
+    RusotoError(String),
```
tgeoghegan, line 59: Rusoto's APIs (e.g. [get_object](https://rusoto.github.io/rusoto/rusoto_s3/trait.S3.html#tymethod.get_object)) all return errors from distinct enums (e.g. [GetObjectError](https://rusoto.github.io/rusoto/rusoto_s3/enum.GetObjectError.html) vs. [CompleteMultipartUploadError](https://rusoto.github.io/rusoto/rusoto_s3/enum.CompleteMultipartUploadError.html)) which makes it difficult to wrap them without an explosion of enum cases in our own `Error` type. For now I `format!` underlying errors into a `String` and was planning to revisit this once @yuriks's error handling work lands.

facilitator/src/transport.rs
```diff
@@ -88,11 +100,314 @@ impl Transport for NullTransport {
     }
 }
 
-// TODO: S3Transport; https://github.com/rusoto/rusoto
+/// Implementation of Transport that reads and writes objects from Amazon S3.
+/// Credentials for authenticating to AWS are automatically sourced from
+/// environment variables or ~/.aws/credentials.
+/// https://github.com/rusoto/rusoto/blob/master/AWS-CREDENTIALS.md
```
tgeoghegan, line 106: I punted on more sophisticated credential handling for now because I think we don't yet know how to get our AWS credential to our jobs running in GCP. [There are several ways to provide credentials in Rusoto](https://docs.rs/rusoto_credential/0.45.0/rusoto_credential/index.html).

yuriks (changes requested):
This is a partial review. I figured I'd submit it so you can work on these things in the meantime since you're going to be rebasing on top of my branch after it's merged anyway.

facilitator/src/transport.rs
```diff
@@ -88,11 +100,314 @@ impl Transport for NullTransport {
...
+            region: region,
+            bucket: bucket,
+            client_provider: client_provider,
+        }
+    }
+
+    /// Convert Path to key suitable for use in S3 API.
+    fn s3_key(key: &Path) -> Result<String, Error> {
+        Ok(key
+            .to_str()
```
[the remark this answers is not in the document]
tgeoghegan: I wrote #28 to track the inadequacy of `Path` here. I could incorporate a fix into this change.

facilitator/src/transport.rs
```diff
@@ -88,11 +100,314 @@ impl Transport for NullTransport {
...
+        // enough content.
+        self.buffer.extend_from_slice(buf);
+        if self.buffer.len() >= self.buffer_capacity {
+            self.upload_part()?;
+        }
+
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
```
tgeoghegan, line 374: One consequence of this implementation is that callers _must_ call `std::io::Write::flush` on the `Write` they get or the upload will (1) never be completed and (2) we will be billed indefinitely for any upload parts Amazon is holding on to. We control all the callers in this same codebase, so we can make sure we flush() appropriately (hence the changes I made in `batch.rs`) but I wonder if `MultipartUploadWriter` could defend itself against this better. For instance, if we added an `std::ops::Drop` implementation that panics if `self.completed_parts` is non-empty, to help programmers detect what I think is a programming error rather than something that could reasonably happen at runtime.

yuriks: Rust's type system has no way of "preventing a type from being dropped" and thus forcing you to consume it somehow, so that's not really possible. The closest approximation to that is to have a function accepting a closure, which acquires the resource, runs the closure and then immediately cleans up the resource afterwards. This might be something we can do here but it would require a lot of changes and would probably be awkward to use.

Now of course you're free to impl `Drop` and then make a blocking call to the network to finalize the chunk, but that will not extend to async network calls, so it feels like a bad solution to have that be the main API and not just a fallback.

In addition, using `flush` on the writer (as you're doing now) is also a bad option, because it means that you're breaking the contract from the trait that `flush` can be called multiple times.

So in the end I think a reasonable thing to do is have a dedicated method in `MultipartUploadWriter` to finish the upload, and another to cancel it. As for the cleanup, panicing in `Drop` if none of the aforementioned methods were called is a possibility, but I don't know if it's actually a good idea in production. I suppose there are no cases in which we'd rather execution finish with a forgotten file which was not written (rather than aborting entirely), so this might be valid for our use-case (but probably not others, which is fine). An alternative is logging an error and then, in the background, calling the cancellation method as a last-ditch effort.

As for encouraging users to not forget to call the functions in the first place, a good way of doing this is to make the finalization call return some value you're going to need later in the program. I don't know if there's anything that would fit the bill here, but it's something to do if you can think of something that'd fit.

Regarding the costs of not finalizing an upload, we should set a [bucket lifecycle policy](https://docs.aws.amazon.com/AmazonS3/latest/dev/mpuoverview.html#mpu-stop-incomplete-mpu-lifecycle-config) which will expire unfinished uploads after a certain amount of time. We'd need to do this either way because there's always the chance that your machine/process randomly die at any point during the upload.

tgeoghegan: Ooh, good point about the `flush` contract. The only problem with extending the `MultipartUploadWriter` method surface as you describe is that `Transport::put` would no longer return an `std::io::Write` but we control all the callers so that's not so bad.

facilitator/Cargo.toml
```diff
@@ -11,12 +11,19 @@ avro-rs = "0.11.0"
...
 chrono = "0.4"
 clap = "2.33.3"
+derivative = "2.1.1"
+hyper = "0.13.8"
+hyper-rustls = "0.21.0"
 prio = "0.2"
 rand = "0.7"
 ring = { version = "0.16.15", features = ["std"] }
+rusoto_core = { version="0.45.0", default_features=false, features=["rustls"] }
+rusoto_mock = { version="0.45.0", default_features=false, features=["rustls"] }
```
yuriks, line 21: This can be moved to `[dev-dependencies]`

facilitator/src/bin/facilitator.rs
```diff
@@ -67,19 +102,25 @@ fn main() -> Result<(), anyhow::Error> {
...
                         .value_name("DIR")
                         .default_value(".")
+                        .validator(path_validator)
                         .help(
-                            "Directory to write sample data for the PHA (aka \
-                            first server) into",
+                            "Storage to write sample data for the PHA \
+                            (aka first server) into. May be either a local \
+                            filesystem path or an S3 bucket, formatted as \
+                            \"s3://{region}-{bucket-name}\"",
```
yuriks, line 110: The examples in the help all use `-` as a separator instead of `/`

facilitator/src/transport.rs
```diff
@@ -130,33 +476,339 @@ mod tests {
...
+                    MultipleMockRequestDispatcher::new(requests),
+                    MockCredentialsProvider,
+                    region.clone(),
+                )
+            },
+        )
+        .expect("failed to create multipart upload writer");
+
+        // First write will fail due to HTTP 401
+        writer.write(&[0; 51]).expect_err("expected error");
```
yuriks, line 713: These messages are all quite superfluous. For tests just use `unwrap()` and `unwrap_err()`, the error will have the actual code line.

yuriks (changes requested):
Finished looking at the rest of the code I hadn't before, and also left some new things I noticed.

facilitator/src/bin/facilitator.rs
```diff
@@ -646,3 +710,16 @@ fn public_key_from_arg(arg: &str, matches: &ArgMatches) -> UnparsedPublicKey<Vec
         Err(_) => UnparsedPublicKey::new(&ECDSA_P256_SHA256_FIXED, key_bytes),
     }
 }
+
+fn transport_for_output_path(arg: &str, matches: &ArgMatches) -> Result<Box<dyn Transport>> {
+    let path = parse_path(matches.value_of(arg).unwrap())?;
```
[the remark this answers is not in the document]
tgeoghegan: I think just to hide an `unwrap` call behind this function, and put less noise at the callsite. That said a subsequent PR does end up pulling more than one value from `matches` here (to configure whether the `S3Transport` should fetch an auth token from instance metadata or use AWS creds from the environment) so this will end up being useful soon.

facilitator/Cargo.toml
```diff
@@ -17,9 +17,8 @@ hyper-rustls = "0.21.0"
 prio = "0.2"
 rand = "0.7"
 ring = { version = "0.16.15", features = ["std"] }
-rusoto_core = { version="0.45.0", default_features=false, features=["rustls"] }
-rusoto_mock = { version="0.45.0", default_features=false, features=["rustls"] }
-rusoto_s3 = { version="0.45.0", default_features=false, features=["rustls"] }
+rusoto_core = { version = "0.45.0", default_features=false, features=["rustls"] }
+rusoto_s3 = { version = "0.45.0", default_features=false, features=["rustls"] }
```
yuriks, line 21: You still missed spaces around the rest of the `=`s (for `default_features` and `features`, sorry)

facilitator/src/transport.rs
```diff
@@ -24,47 +63,325 @@ impl LocalFileTransport {
...
+                        )
+                        .into(),
+                    ),
+                    ..Default::default()
+                }),
+            )
+            // Can't use anyhow here because we must return std::io::Result
+            .map_err(|e| {
+                // Clean up botched uploads
+                if let Err(cancel) = self.cancel_upload() {
```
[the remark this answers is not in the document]
yuriks: Ah, this doesn't work because it'd move out `e` in the closure.

facilitator/src/lib.rs
```diff
@@ -64,55 +63,26 @@ impl Write for DigestWriter {
...
-    fn cancel_upload(&mut self) -> Result<()> {
-        Ok(())
-    }
-}
-
-impl MemoryWriter for Vec<u8> {}
-
 /// SidecarWriter wraps an std::io::Write, but also writes any buffers passed to
 /// it into a MemoryWriter. The reason sidecar isn't simply some other type that
 /// implements std::io::Write is that SidecarWriter::write assumes that short
```
yuriks, line 68: This comment is outdated now.

yuriks (approved):
Left 2 trivial comments but LGTM otherwise, feel free to merge after fixing them/rebasing!

PR #37, opened 2020-10-02
Use error handling libraries

yuriks:
See commit messages for more info. I recommend reviewing each commit individually.

### DIFF
--- a/facilitator/src/aggregation.rs
+++ b/facilitator/src/aggregation.rs
@@ -7,6 +7,7 @@ use crate::{
     transport::Transport,
     Error,
 };
+use anyhow::{anyhow, Context, Result};
 use chrono::NaiveDateTime;
 use prio::{encrypt::PrivateKey, server::VerificationMessage};
 use ring::signature::{EcdsaKeyPair, KeyPair, UnparsedPublicKey, ECDSA_P256_SHA256_FIXED};
@@ -42,7 +43,7 @@ impl<'a> BatchAggregator<'a> {
         share_processor_signing_key: &'a EcdsaKeyPair,
         peer_share_processor_key: &'a UnparsedPublicKey<Vec<u8>>,
         share_processor_ecies_key: &'a PrivateKey,
-    ) -> Result<BatchAggregator<'a>, Error> {
+    ) -> Result<BatchAggregator<'a>> {
         Ok(BatchAggregator {
             is_first: is_first,
             aggregation_name: aggregation_name,
@@ -67,10 +68,7 @@ impl<'a> BatchAggregator<'a> {
 
     /// Compute the sum part for all the provided batch IDs and write it out to
     /// the aggregation transport.
-    pub fn generate_sum_part(
-        &mut self,
-        batch_ids: &Vec<(Uuid, NaiveDateTime)>,
-    ) -> Result<(), Error> {
+    pub fn generate_sum_part(&mut self, batch_ids: &Vec<(Uuid, NaiveDateTime)>) -> Result<()> {
         let share_processor_public_key = UnparsedPublicKey::new(
             &ECDSA_P256_SHA256_FIXED,
             Vec::from(self.share_processor_signing_key.public_key().as_ref()),
@@ -137,7 +135,7 @@ impl<'a> BatchAggregator<'a> {
         &mut self,
         batch_id: &Uuid,
         batch_date: &NaiveDateTime,
-    ) -> Result<IngestionHeader, Error> {
+    ) -> Result<IngestionHeader> {
         let ingestion_batch: BatchReader<'_, IngestionHeader, IngestionDataSharePacket> =
             BatchReader::new_ingestion(
                 self.aggregation_name,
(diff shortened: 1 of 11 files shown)


facilitator/src/lib.rs
```diff
@@ -13,27 +13,36 @@ pub mod transport;
...
     MalformedSignatureError(String),
+    #[error("malformed data packet: {0}")]
     MalformedDataPacketError(String),
+    #[error("end of file")]
     EofError,
-    IoError(String, std::io::Error),
-    LibPrioError(String, Option<EncryptError>),
+    #[error("I/O error: {0}")]
+    IoError(String, #[source] std::io::Error),
+
```
yuriks: Redundant whitespace which I should remove before merging

facilitator/src/bin/facilitator.rs
```diff
@@ -614,11 +601,10 @@ fn main() -> Result<(), Error> {
                 .unwrap()
                 .map(|s| NaiveDateTime::parse_from_str(&s, DATE_FORMAT).unwrap())
                 .collect();
-            if batch_ids.len() != batch_dates.len() {
-                return Err(Error::IllegalArgumentError(
-                    "must provide same number of batch-id and batch-date values".to_owned(),
-                ));
-            }
+            ensure!(
```
yuriks, line 604: I'd decided not to use `ensure!` but I must've missed this one. Expand to an explicit if before merging.

facilitator/src/test_utils.rs
```diff
@@ -0,0 +1,78 @@
+use ring::signature::{
```
tgeoghegan, line 1: nit: these values are also used to provide default arguments in `src/main/facilitator.rs` so `test_utils.rs` isn't _quite_ right. `default_keys.rs`?

facilitator/src/bin/facilitator.rs
```diff
@@ -528,18 +537,11 @@ fn main() -> Result<(), Error> {
...
-                Ok(priv_key) => priv_key,
-                Err(e) => {
-                    return Err(Error::CryptographyError(
-                        "failed to parse value for share-processor-private-key".to_owned(),
-                        e,
-                    ))
-                }
-            };
+            )
+            .context("failed to parse value for share-processor-private-key")?;
```
tgeoghegan, line 544: I love all the places where you replace seven lines of matching with a one line call.

facilitator/src/batch.rs
```diff
@@ -167,30 +168,29 @@ impl<'a, H: Header, P: Packet> BatchIO<'a, H, P> for BatchReader<'a, H, P> {
 impl<'a, H: Header, P: Packet> BatchReader<'a, H, P> {
     /// Return the parsed header from this batch, but only if its signature is
     /// valid.
-    pub fn header(&self, key: &UnparsedPublicKey<Vec<u8>>) -> Result<H, Error> {
+    pub fn header(&self, key: &UnparsedPublicKey<Vec<u8>>) -> anyhow::Result<H> {
         let mut signature = Vec::new();
         self.transport
             .get(self.batch.signature_key())?
             .read_to_end(&mut signature)
-            .map_err(|e| Error::IoError("failed to read signature".to_owned(), e))?;
```
tgeoghegan, line 175: I'm so glad to see all these `to_owned()`es go away. I hated typing each one.

tgeoghegan (approved):
I have a nit or two you should feel free to ignore. I love this change! `anyhow` is a great find.

facilitator/src/aggregation.rs
```diff
@@ -42,7 +43,7 @@ impl<'a> BatchAggregator<'a> {
         share_processor_signing_key: &'a EcdsaKeyPair,
         peer_share_processor_key: &'a UnparsedPublicKey<Vec<u8>>,
         share_processor_ecies_key: &'a PrivateKey,
-    ) -> Result<BatchAggregator<'a>, Error> {
+    ) -> anyhow::Result<BatchAggregator<'a>> {
```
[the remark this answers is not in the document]
yuriks: No, nothing terrible happens. It's just that since the current code already uses `std::Result`, importing both at random across the codebase would be confusing. After everything is converted we can do a pass and just import it everywhere.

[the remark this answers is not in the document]
yuriks: In retrospect I could have actually imported it, since if you pass both parameters to it, it works exactly the same. (The definition is just `pub type Result<T, E = Error> = core::result::Result<T, E>;`)

facilitator/src/test_utils.rs
yuriks: Well, my point in the commit message is that they *shouldn't* be defaults there. We shouldn't have things be set up such that things are accidentally encrypted using a compromised key if you forget to specify the flag.

facilitator/src/bin/facilitator.rs
yuriks: There's a lot more places I could've done that, but they'd require some restructuring of the code usually, so I decided to just not do any of that at all in order to avoid going into rabbit holes. So, future work!

facilitator/src/aggregation.rs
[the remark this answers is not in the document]
yuriks: I said "in retrospect" but I can just put this in the change right now, so I'll do that later.

facilitator/src/test_utils.rs
tgeoghegan: Yes, sorry, I really should have read your commits first. I think you make a good point about safety here, and we should work to make it as hard as possible to use this tool wrong. I just did command line flags for everything because as I was writing this, the intended audience was engineers driving it from their shells, but as you work through the execution manager, we'll have to figure out how we want to provide parameters to this program (environment variables? config files mounted into the container? params passed through the `args` array in a k8s job template?) and that strikes me as a good time to decide how to make this safe.

tgeoghegan (approved):

PR #40, opened 2020-10-06
workflows: Run clippy during CI

yuriks:
I recommend reviewing commits individually.

This PR fixes or silences all the outstanding clippy lints in the default configuration, and enable clippy to run as a GitHub action. This also includes a few other unrelated cleanups I spotted along the way.

One trade-off here is that I configured clippy to run as part of our existing "build" job (just like the `cargo fmt` check currently does). This is nice in that is re-uses the compilation cache that clippy generates (since it needs to actually do full type-checking with the compiler to work, it just doesn't do any codegen), but it also means that a clippy failure will abort the job before it attempts to do an actual compile. While clippy does report most error that the compiler itself does, it does miss a few things checked by the compiler in later stages. In this case the PR author would need to a edit-push-wait for CI cycle before the other errors will be surfaced by CI.

To fix this we could either:
 - Move clippy (and probably also fmt) to its own dedicated job, which would run in parallel with the build job. This uses more CI resources since it's essentially compiling things twice, but makes the two jobs fail independently.
 - Modify the workflows definition file so that it continues executing later steps even if the clippy step fails. This however complicates the workflow definition since GitHub doesn't seem to support this use case well. (You need to insert explicit `if:` conditions on every single later build step.)

If anyone thinks it would be worth it, I can implement either of these options and update the PR.


tgeoghegan (approved):

yuriks:
Filed #48 to track future improvements to the GitHub Actions fail-fast behavior which I mentioned in the PR description.

PR #42, opened 2020-10-07
Remove Some Clones

gsquire:
This small patch removes some calls to `clone` for types that are either pointers or implement `Copy`.


yuriks (approved):
Thanks for the contribution!

PR #50, opened 2020-10-13
Update ingestion-data-share-packet.avsc

winstrom:
Changing doc string of "device_nonce" to be a more accurate and informative description.


tgeoghegan:
This seems fine to me. @winstrom you're aware that the data share processors do nothing with this value, and that it doesn't get included in any form in any of the subsequent messages in the data pipeline?

Issue #20, opened 2020-09-22
Deploy tool(s)


tgeoghegan:
Once we have a Terraformized application, we still need tools around that which can do things like issue keys, construct keyfiles, post them to S3 buckets, get certificates, etc. This tool would fit into a PHA onboarding workflow in which we obtain a minimal set of parameters from new PHAs (e.g., S3 bucket URLs, keyfile location) and let the tool do the rest accordingly.

tgeoghegan:
Pulling into the pha-integration milestone, because I intend to use the deploy tool to get X.509 certificates over packet decryption keys, and we need those for that test milestone.

tgeoghegan:
We have deploy-tool, our janky little pal, and it suffices for our Narnia deployment needs. Further work (namely the automated key rotator) will be tracked elsewhere.

Issue #15, opened 2020-09-21
Terraform-ize cloud deployment


tgeoghegan:
Once we have settled on a public cloud to use, write a Terraform module that can spin up a facilitator instance.

tgeoghegan:
Calling this a P1 for the pha-integration milestone because I really, _really_ want to do this but depending on how many instances we have at that stage, we could conceivably spin things up by hand.

tgeoghegan:
The Terraform stuff we merged today is a lot of this work, but we still need stuff like creation and configuration of S3 buckets, possibly construction and posting of manifests, so I won't close this yet.

tgeoghegan:
We're capable of deploying all the stuff we need for the Narnia test. Any further work on terraform modules will be tracked in other issues, in other milestones.

Issue #24, opened 2020-09-24
Automated message signing and packet encryption key rotation


tgeoghegan:
Once we have all participants in the system advertising public keys and other params from keyfiles or manifests, we can automatically rotate keys used to sign messages written into S3 buckets.

tgeoghegan:
Further, we should run the key generation and rotation tasks in the Kubernetes cluster as a cronjob, so that secrets never need to be exposed to our operator computers. This will involve building a Docker image for the key rotator and putting a `kubectl run` invocation in a makefile, I think.

tgeoghegan:
Retitled; this is about validating and finishing our implementation of key rotation (`deploy-operator`, `manifest-updater`, etc.), and then deploying it in all the `prio-server` deployments (ISRG/prod-us, ISRG/prod-intl, NCI/prod-us, Google/prod-intl).

tgeoghegan:
When we do this, we should also consider automating key backups (relatedly, see #935). For operational smoothness, it'd be nice if key backups were automated. On the other hand, if we empower some Kubernetes job to do this, then we have to be careful to configure ACLs on the backups so that they cannot be deleted or tainted with by a bug somewhere in `prio-server`. Maybe the versioning features offered by the each cloud platform's secrets storage products suffices.

BranLwyd:
Some thoughts on validations that the key rotation infrastructure will need to implement:
* Keys must either have no versions, or exactly one primary version.
* Post-rotation, keys must have at least one version.
* When being used to update keys in a manifest, keys must have at least one version.
* When being used to update keys in a manifest, keys must have versions with distinct creation times.
* When being used to update keys in a manifest, existing key versions' public portions should match with the public keys written to the manifest, matching by creation time/key ID (both of which would provide the same matching).
* Post-update, manifests must have exactly one packet encryption key version.
* Post-update, manifests must have at least one batch signing key version.
* Post-update, manifests' non-key data must match the pre-update manifest data exactly.

BranLwyd:
From discussion in https://github.com/abetterinternet/prio-server/pull/1114:
> Specifically, if a packet encryption key got so far ahead of its related manifests that the key did not include the version advertised as primary in the manifest, we would be asking clients to encrypt with a key we no longer consider valid for decryption. And if a batch signing key got so far ahead of its related manifests that the primary key was not included in the manifest, we would be trying to sign data with a key that we do not recognize as valid for verification.

[...]

PR #51, opened 2020-10-13
 Terraform: Kubernetes secrets, renames, manifests, construction of all necessary data share processors, deploy tooling

tgeoghegan:
This PR contains a series of commits that enables us to deploy data share processors in support of arbitrarily many ingestor servers and PHAs. The specific features are discussed in each commit message. There are several missing pieces here (tests on deploy-tool, support in deploy-tool for certifying packet decryption keys, Terraform management of manifest domain and its DNS, inclusion of key-identifier in batch signatures, hooking up peer manifest discovery to arguments on facilitator) but I wanted to get all these changes visible to our team because they might inform how the workflow manager will be built, and also so that the remaining tasks could be tackled in parallel.

Partially addresses: #20 #15

### DIFF
--- a/deploy-tool/main.go
+++ b/deploy-tool/main.go
@@ -1,0 +1,224 @@
+package main
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"time"
+)
+
+// This tool consumes the output of `terraform apply`, generating keys and then
+// populating specific manifests and Kubernetes secrets with appropriate values.
+// We do this in this tool because if we generated secrets via Terraform
+// resources, the secret values would appear in the Terraform state file. The
+// struct definitions here MUST be kept in sync with the output variable in
+// terraform/modules/facilitator/facilitator.tf and the corresponding structs in
+// facilitator/src/manifest.rs.
+
+// BatchSigningPublicKey represents a public key used for batch signing.
+type BatchSigningPublicKey struct {
+	// PublicKey is the PEM armored base64 encoding of the ASN.1 encoding of the
+	// PKIX SubjectPublicKeyInfo structure. It must be an ECDSA P256 key.
+	PublicKey string `json:"public-key"`
+	// Expiration is the ISO 8601 encoded UTC date at which this key expires.
+	Expiration string `json:"expiration"`
+}
+
+// PacketEncryptionCertificate represents a certificate containing a public key
+// used for packet encryption.
+type PacketEncryptionCertificate struct {
+	// Certificate is the PEM armored X.509 certificate.
+	Certificate string `json:"certificate"`
+}
+
+// SpecificManifest represents the manifest file advertised by a data share
+// processor. See the design document for the full specification.
+// https://docs.google.com/document/d/1MdfM3QT63ISU70l63bwzTrxr93Z7Tv7EDjLfammzo6Q/edit#heading=h.3j8dgxqo5h68
+type SpecificManifest struct {
+	// Format is the version of the manifest.
+	Format float64 `json:"format"`
+	// IngestionBucket is the region+name of the bucket that the data share
+	// processor which owns the manifest reads ingestion batches from.
+	IngestionBucket string `json:"ingestion-bucket"`
+	// PeerValidationBucket is the region+name of the bucket that the data share
+	// processor which owns the manifest reads peer validation batches from.
+	PeerValidationBucket string `json:"peer-validation-bucket"`
+	// BatchSigningPublicKeys maps key identifiers to batch signing public keys.
+	// These are the keys that peers reading batches emitted by this data share
+	// processor use to verify signatures.
+	BatchSigningPublicKeys map[string]BatchSigningPublicKey `json:"batch-signing-public-keys"`
+	// PacketEncryptionCertificates maps key identifiers to packet encryption
+	// certificates. The values are PEM encoded X.509 certificates, which
+	// contain the public key corresponding to the private key that the data
+	// share processor which owns the manifest uses to decrypt ingestion share
+	// packets.
+	PacketEncryptionCertificates map[string]PacketEncryptionCertificate `json:"packet-encryption-certificates"`
+}
+
+// TerraformOutput represents the JSON output from `terraform apply` or
+// `terraform output --json`. This struct must match the output variables
+// defined in terraform/main.tf, though it only need describe the output
+// variables this program is interested in.
+type TerraformOutput struct {
+	ManifestBucket struct {
+		Value string
+	} `json:"manifest_bucket"`
+	SpecificManifests struct {
+		Value map[string]SpecificManifest
+	} `json:"specific_manifests"`
+}
+
+func generateKeyPair(dataShareProcessorName, keyName string) (*ecdsa.PrivateKey, error) {
+	ecdsaKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate ECDSA P256 key: %w", err)
+	}
+
+	pkcs8PrivateKey, err := x509.MarshalPKCS8PrivateKey(ecdsaKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal ECDSA key to PKCS#8 document: %w", err)
+	}
+
+	// Put the private keys into Kubernetes secrets. There's no straightforward
+	// way to update a secret value using kubectl, so we use the trick from[1].
+	// There's no particular reason to use `-o=json` but we must set some output
+	// format or nothing is written to stdout.
+	// [1] https://blog.atomist.com/updating-a-kubernetes-secret-or-configmap/
+	//
+	// We can't provide the base64 encoding of the key to --from-literal because
+	// then kubectl would base64 the base64, so we have to write the b64 to a
+	// temp file and then provide that to --from-file.
+	log.Printf("updating Kubernetes secret %s/%s", dataShareProcessorName, keyName)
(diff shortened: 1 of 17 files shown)


tgeoghegan:
Force pushed some fresh changes to reflect the latest consensus on manifest file formats and that both Apple and Google ingestors use a single packet encryption key.

deploy-tool/main.go
```diff
@@ -0,0 +1,227 @@
...
+	} `json:"manifest_bucket"`
+	SpecificManifests struct {
+		Value map[string]struct {
+			KubernetesNamespace string           `json:"kubernetes-namespace"`
+			SpecificManifest    SpecificManifest `json:"specific-manifest"`
+		}
+	} `json:"specific_manifests"`
+}
+
+func generateKeyPair(namespace, keyName string) (*ecdsa.PrivateKey, error) {
```
aaomidi, line 84: Don't know how I feel about this method also interfacing with kubernetes. Makes testing this code a bit harder too from what I can gather.

tgeoghegan: You mean `os.Exec("kubectl")`? I agree, it's pretty bad. It's also pretty bad that we run this locally on a deployer's personal machine, because it exposes the keys to us individually. What I think we should do for real is run this key generation job in the k8s cluster, and as a cronjob so it can automatically rotate keys, which is tracked by #24. The goal here is to enable us to stand up an instance for Narnia and for Atredis to look at on the 26th.

However I agree the the function name could clarify that it also populates a k8s secret.

aaomidi: That's fair yeah, we're taking plenty shortcuts :D

facilitator/src/manifest.rs
```diff
@@ -0,0 +1,336 @@
...
+            ));
+        }
+        if pem.contents.len() < ECDSA_P256_SPKI_PREFIX.len() {
+            return Err(anyhow!("PEM contents not long enough to contain ASN.1 encoded ECDSA P256 SubjectPublicKeyInfo"));
+        }
+        if &pem.contents[..ECDSA_P256_SPKI_PREFIX.len()] != ECDSA_P256_SPKI_PREFIX {
+            return Err(anyhow!(
+                "PEM contents are not ASN.1 encoded ECDSA P256 SubjectPublicKeyInfo"
+            ));
+        }
```
yuriks, line 121: Suggestion: Use `split_at` which lets you avoid repeating the lenght a few times.
```rust
let (prefix, key) = pem.contents.split_at(ECDSA_P256_SPKI_PREFIX.len());
// Check if prefix != ECDSA_P256_SPKI_PREFIX
// Otherwise, go ahead and return key.
```

Unfortunately `strip_prefix` isn't stable yet otherwise that'd be an even better fit.

facilitator/src/manifest.rs
```diff
@@ -0,0 +1,336 @@
...
+            .context(format!("no value for key {}", identifier))?;
+
+        let pem = pem::parse(&key.public_key)
+            .context(format!("failed to parse key entry {} as PEM", identifier))?;
+        if pem.tag != "PUBLIC KEY" {
+            return Err(anyhow!(
+                "key for identifier {} is not a PEM encoded public key"
+            ));
+        }
+        if pem.contents.len() < ECDSA_P256_SPKI_PREFIX.len() {
```
yuriks, line 114: I did some checking and figured out that the size of the signature is alway 520 bits (or 65 bytes):

```py
>>> import base64
>>> prefix = bytes([0x30, 0x59, 0x30, 0x13, 0x06, 0x07, 0x2a, 0x86, 0x48, 0xce, 0x3d, 0x02, 0x01, 0x06, 0x08, 0x2a, 0x86, 0x48, 0xce, 0x3d, 0x03, 0x01, 0x07, 0x03, 0x42, 0x00,])
>>> base64.b64encode(prefix + bytes([0xFF]*65))
b'MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgD//////////////////////////////////////////////////////////////////////////////////////w=='
```

Which can be [fed into asn1js](https://lapo.it/asn1js/#MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgD______________________________________________________________________________________w) to verify it parses. asn1js also warns if there's excess data at the end, so this is the right number of bytes.

So this way you can also check that the entire SPKI has a length of 91 bytes, and the signature 65 bytes, to be a bit more strict.

deploy-tool/main.go
yuriks: +1 on renaming this to reflect that it also uploads the secret

yuriks (changes requested):
Thanks for separating the commits, especially putting some of the renames in a commit of their own, it makes things easier/faster to review.

One major issue I have with this change is I don't like the approach of having the jobs hit the manifest url directly when they run. This mixes data/control-plane concerns, makes us repeatedly re-request the manifest for every batch, forces us to allow the jobs to hit arbitrary internet endpoints (they can't be firewalled/be in a private cluster), and adds additional potential points of failure.

That said, since we're in a hurry, this implementation is fine for now. If we have time I think we should overhaul this later on along with our deployment story though.

tgeoghegan:
> One major issue I have with this change is I don't like the approach of having the jobs hit the manifest url directly when they run. This mixes data/control-plane concerns, makes us repeatedly re-request the manifest for every batch, forces us to allow the jobs to hit arbitrary internet endpoints (they can't be firewalled/be in a private cluster), and adds additional potential points of failure.

We don't have to fetch peer manifests from the batch jobs. The workflow manager (control plane) will have to consult peer specific manifests no matter what, in order to find peer batch signing keys so it can verify batch signatures, so we could have it extract all peer config values from manifests and then pass those to individual batch processing jobs in the Kubernetes job template (as env vars, as args to the container, as a config file).

terraform/modules/data_share_processor/data_share_processor.tf
```diff
@@ -0,0 +1,235 @@
...
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "Federated": "accounts.google.com"
+      },
+      "Action": "sts:AssumeRoleWithWebIdentity",
+      "Condition": {
+        "StringEquals": {
+          "accounts.google.com:sub": "${var.ingestor_google_service_account_id}"
```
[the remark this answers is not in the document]
tgeoghegan: I think that makes sense, but it means we have to specify an `aud` value that Google should place into their OIDC auth token. Off the top of my head I'm not sure what we'd ask them to do (i.e. whether we might attach it to a particular PHA instance) but we could talk to the Google folks about what a sensible `aud/oaud` value to check would be. #59

terraform/modules/data_share_processor/data_share_processor.tf
```diff
@@ -0,0 +1,235 @@
...
+      ],
+      "Resource": [
+        "arn:aws:s3:::${local.ingestion_bucket_name}/*",
+        "arn:aws:s3:::${local.ingestion_bucket_name}"
+      ]
+    },
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "AWS": "${aws_iam_role.bucket_owner_role.arn}"
```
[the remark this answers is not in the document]
tgeoghegan: I apologize, that should be `aws_iam_role.bucket_role.arn`, which is what it changed to in more recent commits. I think I just forgot to retest this commit during an interactive rebase.

terraform/main.tf
```diff
@@ -78,24 +88,70 @@ provider "kubernetes" {
...
+# (peer, ingestor) pair.
+# First, we fetch the ingestor global manifests, which yields a map of ingestor
+# name => HTTP content.
+data "http" "ingestor_global_manifests" {
+  for_each = var.ingestors
+  url      = "https://${each.value}/global-manifest.json"
+}
+
+# Then we fetch the single global manifest for all the peer share processors.
+data "http" "peer_share_processor_global_manifest" {
```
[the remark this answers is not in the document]
tgeoghegan: I agree with Yuri on this. We could make `var.peer_share_processor_names` a list of `pha_name, global manifest reference` tuples, or maybe have a map structure as described in #60.

deploy-tool/main.go
```diff
@@ -0,0 +1,224 @@
...
+	// Certificate is the PEM armored X.509 certificate.
+	Certificate string `json:"certificate"`
+}
+
+// SpecificManifest represents the manifest file advertised by a data share
+// processor. See the design document for the full specification.
+// https://docs.google.com/document/d/1MdfM3QT63ISU70l63bwzTrxr93Z7Tv7EDjLfammzo6Q/edit#heading=h.3j8dgxqo5h68
+type SpecificManifest struct {
+	// Format is the version of the manifest.
+	Format float64 `json:"format"`
```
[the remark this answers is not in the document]
tgeoghegan: I think at some point because it's the _Javascript_ Object Notation, I thought you had to declare the field as a float, but no, Go does the sensible thing with `int64` fields, so I changed this to `int64`.

deploy-tool/main.go
```diff
@@ -0,0 +1,224 @@
...
+			}
+
+			block := pem.Block{
+				Type:  "PUBLIC KEY",
+				Bytes: []byte(pkixPublic),
+			}
+
+			newBatchSigningPublicKeys[name] = BatchSigningPublicKey{
+				PublicKey:  string(pem.EncodeToMemory(&block)),
+				Expiration: time.Now().UTC().Format(time.RFC3339),
```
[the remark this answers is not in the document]
tgeoghegan: Yes, I remain uncertain that this field is actually useful given how it looks like key rotation will actually work. I just wanted to plug in a semantically valid value here. I'll change it to be 90 days from creation so it's more plausible (90 days chosen to match the lifetime of the packet encryption key certificates but we should talk about that figure more widely) but I increasingly think we shouldn't have this field at all.

facilitator/src/manifest.rs
tgeoghegan: I wanted to use `strip_prefix`, too. However `split_at` is _much_ nicer and removes a few slice operators, which makes me happy.

yuriks (approved):

PR #53, opened 2020-10-14
Create a valid cert for the private key used for the packet encryption certificate

aaomidi:
A few concerns I have with this PR immediately:

- I've tested the components separately, but not together in the code. Will need help figuring out how to test them.
- The user key gets stored in `./deploy-tool-output/acme/....`, this is probably not what we want for long term and would like this to just run inside Kubernetes and use k8s' various APIs directly (can configure this pod to have a service account and just use the k8s APIs directly without configuring anything)
-  I'm currently configuring a bunch of stuff for this tool using environment variables. This isn't __terrible__ and might actually be nice once inside a k8s cluster, but it's just something to keep in mind.
- The value that finally ends up in `newCertificates[name]` is going to be something like:

```
-----BEGIN CERTIFICATE-----
cert
-----END CERTIFICATE-----

-----BEGIN CERTIFICATE-----
cert
-----END CERTIFICATE-----

```

From what I read in the documentation this is intended, but I wanted to point it out.


deploy-tool/cert/cert.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// IssueCertificate asks the Let's Encrypt CA to validate and sign a certificate for a site using the DNS01 strategy
+// This issue depends on multiple environment variables being set before being invoked:
+// D_CA_AGREE, expecting a value of "true" - checks to see if the user has agreed to the CA terms
+// D_CA_EMAIL, expecting an email address - this is used for the ACME account
+// D_DNS_PROVIDER, expecting a DNS provider, e.g. "cloudflare" - this is used for the DNS challenge
+// D_API_TOKEN, expecting an API token for the DNS provider - this is used to authenticate to the DNS provider
+func IssueCertificate(site string, privKey *ecdsa.PrivateKey) (string, error) {
+
```
tgeoghegan, line 23: nit: delete blank line

deploy-tool/dns/dns.go
```diff
@@ -0,0 +1,35 @@
...
+		return provider, nil
+	}
+
+	return nil, fmt.Errorf("no valid provider selected")
+}
+
+func getDNSProvider() string {
+	dnsProvider := os.Getenv("D_DNS_PROVIDER")
+
+	return dnsProvider
```
tgeoghegan, line 28: nit: just directly `return os.Getenv("D_DNS_PROVIDER")`

deploy-tool/cert/cert.go
```diff
@@ -0,0 +1,95 @@
...
+	acme := certmagic.NewDefault()
+	acme.Storage = &certmagic.FileStorage{Path: "./deploy-tool-output"}
+	csr, err := getCSR([]string{site}, privKey)
+	if err != nil {
+		return "", fmt.Errorf("error when getting the certificate signing request: %v", err)
+	}
+
+	issued, err := acme.Issuer.Issue(context.Background(), csr)
+	if err != nil {
+		return "", fmt.Errorf("error when getting the certificate signing request: %v", err)
```
tgeoghegan, line 57: I think this error string was copy pasted from the one above, but it should refer to the issuance failing.

deploy-tool/cert/cert.go
```diff
@@ -0,0 +1,95 @@
...
+	"os"
+	"strings"
+)
+
+// IssueCertificate asks the Let's Encrypt CA to validate and sign a certificate for a site using the DNS01 strategy
+// This issue depends on multiple environment variables being set before being invoked:
+// D_CA_AGREE, expecting a value of "true" - checks to see if the user has agreed to the CA terms
+// D_CA_EMAIL, expecting an email address - this is used for the ACME account
+// D_DNS_PROVIDER, expecting a DNS provider, e.g. "cloudflare" - this is used for the DNS challenge
+// D_API_TOKEN, expecting an API token for the DNS provider - this is used to authenticate to the DNS provider
```
tgeoghegan, line 21: Why use environment variables for this? Even if we decide to go that way, I'd say the main function or module should handle details like that, and this module should take these parameters in the `IssueCertificate` method or perhaps as fields of a struct you define. Doing it this way would make it harder to write tests on this module, because each test function would have to populate the environment appropriately.

deploy-tool/main.go
```diff
@@ -177,13 +178,16 @@ func main() {
...
 			log.Printf("generating and certifying P256 key %s", name)
-			_, err := generateKeyPair(dataShareProcessorName, name)
+			privKey, err := generateKeyPair(dataShareProcessorName, name)
 			if err != nil {
 				log.Fatalf("%s", err)
 			}
 
-			// TODO(timg) get certificate over the key, insert it here
-			newCertificates[name] = "TODO get certificate"
+			certificate, err := cert.IssueCertificate(name, privKey)
```
tgeoghegan, line 186: "name" here is not what we want to issue the certificate for, but we need to think about how to get the right value. I think we would buy some domain `foo.org`, and then use `${manifestWrapper.KubernetesNamespace}.foo.org`. Of course you'll need to rebase on main after #51 lands to get the version of deploy-tool that plumbs that value in.

deploy-tool/dns/dns.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func getAPIToken() string {
+	apiToken := os.Getenv("D_API_TOKEN")
```
tgeoghegan, line 32: We gotta talk about how to manage this API token.

tgeoghegan (changes requested):

deploy-tool/cert/cert.go
aaomidi: Yeah, I think that makes sense. I followed the design pattern that Lego had created, but I wasn't really happy with it.

This comes down to my question. How is the user going to execute this? Should it be a CLI based utility with flags? Should it be a configuration file?

deploy-tool/dns/dns.go
aaomidi: Why would we need an API token for the LE ACME endpoint?

deploy-tool/main.go
aaomidi: I'll wait on that merge to resolve this issue.

deploy-tool/dns/dns.go
tgeoghegan: There is a notion of an account and a user in the ACME API: https://tools.ietf.org/html/rfc8555#section-7.3

I think maybe certmagic might be abstracting that away from you, and maybe what they're doing is a brand new user for each certificate request. We should check what's going on, and whether it'd be better to manage a persistent ACME credential (maybe ask the SRE and boulder teams for opinions).

aaomidi: Oh you mean the user private key. Yeah since when I wrote this we were planning to only run this on our local machines I assumed we're just going to handle the key.

I think if the intention of this (or another tool that reuses some of the logic) to run automatically in k8s the proper way to do this would be to use the k8s secret storage or just mount it as a volume for the container to see.

terraform/modules/data_share_processor/data_share_processor.tf
```diff
@@ -1,257 +0,0 @@
-variable "data_share_processor_name" {
```
aaomidi, line 1: Will fix this... was an accident :D

deploy-tool/main.go
```diff
@@ -193,13 +194,17 @@ func main() {
...
 			log.Printf("generating and certifying P256 key %s", name)
-			_, err := generateAndDeployKeyPair(manifestWrapper.KubernetesNamespace, name)
+			privKey, err := generateAndDeployKeyPair(manifestWrapper.KubernetesNamespace, name)
 			if err != nil {
 				log.Fatalf("%s", err)
 			}
 
-			// TODO(timg) get certificate over the key, insert it here
-			newCertificates[name] = PacketEncryptionCertificate{Certificate: "TODO get certificate"}
+			certificate, err := cert.IssueCertificate(manifestWrapper.KubernetesNamespace, privKey)
```
aaomidi, line 202: This might have to be updated to manifestWrapper.KubernetesNamespace + the domain

deploy-tool/config/config.go
```diff
@@ -0,0 +1,42 @@
...
+	CloudflareConfig *CloudflareConfig `toml:"Cloudflare"`
+}
+
+type CloudflareConfig struct {
+	APIKey string `toml:"api_key"`
+}
+
+type ACMEConfig struct {
+	Email               string
+	CA                  string
```
tgeoghegan, line 24: `CA` is a little ambiguous. Going by the value in the sample config file, this is the ACME API endpoint, so I think the field should be named accordingly.

deploy-tool/config/config.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+type StorageConfig struct {
+	Driver string
+}
+
+func Read(path string) (DeployConfig, error) {
+	var config DeployConfig
+
+	x, err := toml.LoadFile(path)
```
tgeoghegan, line 35: suggest naming this variable `tree` or something

deploy-tool/cert/k8_storage.go
```diff
@@ -0,0 +1,151 @@
...
+	secretsApi := s.getSecretsAPI()
+
+	secret, err := secretsApi.Get(context.Background(), cleanKey(key), metav1.GetOptions{})
+	if err != nil {
+		return certmagic.KeyInfo{}, err
+	}
+
+	return certmagic.KeyInfo{
+		Key:        key,
+		Modified:   secret.GetCreationTimestamp().UTC(),
```
tgeoghegan, line 133: Is the creation timestamp going to be the right value if the secret was ever updated by `secretApi.Update`?

deploy-tool/cert/k8_storage.go
```diff
@@ -0,0 +1,151 @@
...
+		Size:       int64(len(secret.Data[dataKey])),
+		IsTerminal: true,
+	}, nil
+}
+
+func (s *SecretStorage) Lock(ctx context.Context, key string) error {
+	// Do we need this? If so how should we implement this?
+	return nil
```
tgeoghegan, line 141: My guess is that the Kubernetes secret API should provide reasonable ordering guarantees, so I don't think locking is needed. It's worth doing some research on what guarantees the Kubernetes API provides, though.

deploy-tool/cert/k8_storage.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (s *SecretStorage) Exists(key string) bool {
+	secrets, err := s.getSecretsAPI().List(context.Background(), metav1.ListOptions{
+		FieldSelector: fmt.Sprintf("metadata.name=%v", cleanKey(key)),
+	})
+
+	if err != nil {
+		//TODO where should we log this?
```
tgeoghegan, line 89: Logging errors to `stderr` is fine. When this runs as a cronjob, GKE will capture container stdout/stderr and forward it into Google's managed logs product.

deploy-tool/cert/k8_storage.go
```diff
@@ -0,0 +1,151 @@
...
+func (s *SecretStorage) Delete(key string) error {
+	secretsApi := s.getSecretsAPI()
+
+	err := secretsApi.Delete(context.Background(), cleanKey(key), metav1.DeleteOptions{})
+
+	if err != nil {
+		return err
+	}
+
+	return nil
```
tgeoghegan, line 80: I think you can just `return secretsApi.Delete(...)` and save a few lines here.

deploy-tool/cert/k8_storage.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (s *SecretStorage) Load(key string) ([]byte, error) {
+	secretsApi := s.getSecretsAPI()
+
+	secret, err := secretsApi.Get(context.Background(), cleanKey(key), metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	return secret.Data[dataKey], nil
```
tgeoghegan, line 68: This should check if `secret.Data` contains `dataKey` and return an error if not.

deploy-tool/cert/k8_storage.go
```diff
@@ -0,0 +1,151 @@
...
+
+var matchLabels = map[string]string{
+	"manager": "magiccert",
+}
+var labelSelector = "manager=magiccert"
+
+var specialChars = regexp.MustCompile("[^a-zA-Z0-9_.-]+")
+
+func cleanKey(key string) string {
+	return fmt.Sprintf("cm.k8s.%v", specialChars.ReplaceAllString(key, ""))
```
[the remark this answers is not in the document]
tgeoghegan: Where does the `key` value come from? Do we have any control over it? I'm a little concerned about being clever and trying to sanitize keys because there's a risk of unexpected collisions, though in this case we only ever expect one ACME client credential per Kubernetes cluster so it's a small risk.

deploy-tool/cert/cert.go
tgeoghegan: It seems like all these values now come via config file, so should this comment be updated?

deploy-tool/cert/k8_storage.go
```diff
@@ -0,0 +1,151 @@
...
+	for _, item := range secrets.Items {
+		if item.ObjectMeta.Name == cleanKey(key) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (s *SecretStorage) List(prefix string, _ bool) ([]string, error) {
```
tgeoghegan, line 102: What does the `bool` argument we ignore do?

deploy-tool/cert/k8_storage.go
```diff
@@ -0,0 +1,151 @@
...
+	return fmt.Sprintf("cm.k8s.%v", specialChars.ReplaceAllString(key, ""))
+}
+
+var dataKey = "value"
+
+func (s *SecretStorage) Store(key string, value []byte) error {
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   cleanKey(key),
+			Labels: matchLabels,
```
tgeoghegan, line 37: Do we need to insert the ACME account ID here, or is that stored separately as a secret by certmagic?

deploy-tool/cert/cert.go
```diff
@@ -0,0 +1,84 @@
...
+		DNSProvider: dnsProvider,
+	}
+
+	certmagic.DefaultACME.Agreed = true
+	certmagic.DefaultACME.Email = userEmail
+	// TODO change this to default LE server
+	certmagic.DefaultACME.CA = certmagic.LetsEncryptStagingCA
+
+	acme := certmagic.NewDefault()
+	acme.Storage = &certmagic.FileStorage{Path: "./deploy-tool-output"}
```
tgeoghegan, line 48: Should this specify `SecretStorage` instead of `FileStorage`?

deploy-tool/cert/k8_storage.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/caddyserver/certmagic"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"regexp"
+	"strings"
+)
+
+type SecretStorage struct {
```
tgeoghegan, line 15: `KubernetesSecretStorage` seems like a better name to me. `Secret` is the thing we're storing, but as it is the name doesn't give any insight into _where_ we store the secrets.

deploy-tool/cert/k8_storage_test.go
```diff
@@ -0,0 +1,39 @@
...
+			}
+		})
+	}
+}
+
+func TestStorageImplementationComplete(t *testing.T) {
+	acme := certmagic.NewDefault()
+
+	// Just a compile time check to see if we can use SecretStorage as a certmagic storage system.
+	acme.Storage = &SecretStorage{
```
tgeoghegan, line 35: I think this test is unnecessary if `IssueCertificate` is updated, right?

deploy-tool/dns/dns.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/libdns/cloudflare"
+	"strings"
+)
+
+// GetACMEDNSProvider configures an ACMEDNSProvider value to be used in cert generation
+func GetACMEDNSProvider(deployConfig config.DeployConfig) (certmagic.ACMEDNSProvider, error) {
+	switch strings.ToLower(deployConfig.DNS.Provider) {
+	case "cloudflare":
+		cloudflareConfig := deployConfig.DNS.CloudflareConfig
+		if cloudflareConfig == nil {
```
tgeoghegan, line 16: nit: don't bother defining `cloudflareConfig`, just do `if deployConfig.DNS.CloudflareConfig == nil`

deploy-tool/sample.config.toml
```diff
@@ -0,0 +1,15 @@
+# DNS provider settings
+[dns]
+provider = "cloudflare" # options: cloudflare,
+
+    [dns.cloudflare]
```
tgeoghegan, line 5: Is the indentation significant?

deploy-tool/main.go
```diff
@@ -142,10 +144,25 @@ func generateAndDeployKeyPair(namespace, keyName string) (*ecdsa.PrivateKey, err
 	return ecdsaKey, nil
 }
 
+// getConfigFilePath gets the configuration file path from the DEPLOY_CONFIG_PATH environment variable.
+// this function defaults to ./config.toml if that env variable is empty.
+func getConfigFilePath() string {
+	val := os.Getenv("DEPLOY_CONFIG_PATH")
+	if len(val) == 0 {
+		return "./config.toml"
```
tgeoghegan, line 152: Help me understand where the config file comes from. I don't think any of the values in there are emitted by prior stages of the deployment so I don't think it's emitted from the Terraform module. Looking at what's defined in there, and keeping in mind that imminently we'll be using Google Cloud DNS, it seems like the point of the config file is to allow overriding the default behavior, which would be to use the production Let's Encrypt, use Google Cloud DNS, use Google Cloud Project API credentials from `~/.config/gcloud`, and the Kubernetes driver. I think that's useful for testing this, but now I wonder if in order to absolve every deployer of `prio-server` from being required to make a well-formatted `config.toml`, `deploy-tool/config/config.go` should define default values for the common case.

tgeoghegan (changes requested):

deploy-tool/cert/cert.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	certmagic.DefaultACME.DNS01Solver = &certmagic.DNS01Solver{
+		DNSProvider: dnsProvider,
+	}
+
+	certmagic.DefaultACME.Agreed = true
+	certmagic.DefaultACME.Email = userEmail
+	// TODO change this to default LE server
+	certmagic.DefaultACME.CA = certmagic.LetsEncryptStagingCA
```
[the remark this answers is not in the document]
aaomidi: Thanks for catching this - forgot to pipe it through :)

deploy-tool/cert/cert.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return x509.ParseCertificateRequest(csrDER)
+}
+
+// getUserEmail retrieves the email the user want's to associate with this certificate
+func getUserEmail(deployConfig config.DeployConfig) (string, error) {
+	email := deployConfig.ACME.Email
+	if strings.Contains(email, "@") == false {
+		return "", fmt.Errorf("invalid email address %s", email)
+	}
```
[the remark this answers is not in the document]
aaomidi: Fair, it's not even really a proper validation. Was just thinking of making it fail faster than it would normally, but the extra code is probably not worth it.

deploy-tool/cert/k8_storage.go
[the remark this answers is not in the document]
aaomidi: I don't think the long `letsencrypt-acme-client-credential` is a good idea here considering the keys can only be 63 characters long. Do you think we can make that a bit smaller?

The key is essentially a file path from my understanding. It's essentially created by certmagic and in our storage layer we're mapping these file paths to secrets in k8s.

deploy-tool/cert/k8_storage.go
```diff
@@ -0,0 +1,151 @@
...
+		},
+		Data: map[string][]byte{
+			dataKey: value,
+		},
+	}
+
+	var err error
+
+	secretsApi := s.getSecretsAPI()
+	if s.Exists(key) {
```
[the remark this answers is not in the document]
aaomidi: Unfortunately this doesn't describe it well either https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#create-secret-v1-core but just from experience of working with k8s I know a lot of the other commands doesn't like `create` commands if the resource already exists.

deploy-tool/main.go
```diff
@@ -142,10 +144,25 @@ func generateAndDeployKeyPair(namespace, keyName string) (*ecdsa.PrivateKey, err
 	return ecdsaKey, nil
 }
 
+// getConfigFilePath gets the configuration file path from the DEPLOY_CONFIG_PATH environment variable.
+// this function defaults to ./config.toml if that env variable is empty.
+func getConfigFilePath() string {
+	val := os.Getenv("DEPLOY_CONFIG_PATH")
+	if len(val) == 0 {
```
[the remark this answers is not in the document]
aaomidi: Oh I didn't know about this

deploy-tool/cert/k8_storage.go
aaomidi: Yep an update changes the object entirely I think

deploy-tool/cert/k8_storage.go
aaomidi: It's recursive search. Since technically we don't have a concept of directories, all of our searches are recursive.

aaomidi: It's not a __big__ deal since we're just returning more information than we would usually.

deploy-tool/cert/k8_storage.go
aaomidi: That's part of the `key`

deploy-tool/cert/cert.go
aaomidi: Actually this is probably something that I should make configurable. Just in case we need to run this locally.

deploy-tool/sample.config.toml
aaomidi: No, it's generally up to us if we want the indentation for sub keys. I've generally seen it without indentation so I'll remove them.

deploy-tool/main.go
aaomidi: The config is going to be something the person deploying needs to configure. I'll add a few more defaults but at the very least a few things need to be manually entered:

- Assuming DNS provider defaults to GCP, we'd need the `[dns.gcp]` section (when that gets built out).
- The `acme.email, acme.subscriber_agreement`.
- Defaulting `storage.driver` to `kubernetes` and `storage.kubernetes.namespace` to `prio-server`

deploy-tool/cert/k8_storage.go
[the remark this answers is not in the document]
tgeoghegan: Oh, I didn't know there was a length restriction here. I still think `k8s` is redundant, all the moreso if you're trying to make the best use of a limited length. Can we at least put a label or annotation (whichever is more appropriate) with a more descriptive name?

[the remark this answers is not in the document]
tgeoghegan: What I'm reading in k8s docs is that secret names must be "DNS subdomains", which can be "no more than 253" characters, not 63, so it seems like we have enough breathing room to be explicit here.

https://kubernetes.io/docs/concepts/configuration/secret/#overview-of-secrets
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

deploy-tool/config/config.go
```diff
@@ -0,0 +1,52 @@
...
+	Filesystem *FilesystemConfig
+	Kubernetes *KubernetesConfig
+}
+
+type FilesystemConfig struct {
+	Path string `default:"./deploy_tool_output"`
+}
+
+type KubernetesConfig struct {
+	Namespace string `default:"prio-server"`
```
tgeoghegan, line 39: Putting the ACME credential in a Kubernetes namespace seems like a fine idea, but if the idea is that this will eventually be used by a key rotation cronjob running in that same namespace, maybe "key-rotator" would be more appropriate.

deploy-tool/cert/k8_storage.go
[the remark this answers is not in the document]
aaomidi: Unfortunately these keys aren't really DNS subdomains, just random strings that we're then going to have to map to DNS subdomains if we need to

So we have the limit defined here https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

I agree though I'll get rid of the `k8s`. These are going to be invisible for us as the users though as it's really only ever going to be modified by the key rotation program

deploy-tool/cert/kubernetes_storage.go
```diff
@@ -0,0 +1,168 @@
...
+func (s *KubernetesSecretStorage) Load(key string) ([]byte, error) {
+	secretsApi := s.getSecretsAPI()
+
+	secret, err := secretsApi.Get(context.Background(), cleanKey(key), metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	data, ok := secret.Data[dataKey]
+	if ok {
```
tgeoghegan, line 87: Should this be `if !ok`?

deploy-tool/cert/cert.go
```diff
@@ -0,0 +1,90 @@
...
+	"fmt"
+	"github.com/caddyserver/certmagic"
+	"strings"
+)
+
+// IssueCertificate asks the Let's Encrypt ACMEApiEndpoint to validate and sign a certificate for a site using the DNS01 strategy
+func IssueCertificate(deployConfig config.DeployConfig, site string, privKey *ecdsa.PrivateKey) (string, error) {
+	if deployConfig.ACME.SubscriberAgreement == false {
+		return "", fmt.Errorf("let's encrypt ACMEApiEndpoint terms were not agreed to")
+	}
```
tgeoghegan, line 20: Why not just set `certmagic.DefaultACME.Agreed = deployConfig.ACME.SubscriberAgreement` below, and then let Boulder or CertMagic throw errors if they don't like the value? Once again, it would be less code and fewer test cases for us to worry about.

deploy-tool/main.go
tgeoghegan: I created `enpa-prio-ops@letsencrypt.org` yesterday for another purpose. Can we set that as the default email?

tgeoghegan (changes requested):

tgeoghegan (approved):

PR #57, opened 2020-10-15
Include the sum of the total individual clients inside the sum-part avro message

### LINKED ISSUE #43: Sum part message should include the number of shares that are included in the sum part
In order to de-bias the final data, we need to include the total number of shares that are included in the sum part. We will have to extend the `PrioSumPart` message to include a field for this and have the reduce step(s) record the value.

aaomidi:
Fixes #43


facilitator/src/idl.rs
```diff
@@ -649,6 +649,7 @@ pub struct SumPart {
     pub aggregation_start_time: i64,
     pub aggregation_end_time: i64,
     pub packet_file_digest: Vec<u8>,
+    pub total_individual_clients: i64,
```
tgeoghegan, line 652: I believe you'll also need to update the `read` and `write` methods on `SumPart` to serialize and deserialize this field. The tests at the bottom of this file should also be updated to read/write this field.

facilitator/src/aggregation.rs
```diff
@@ -125,6 +127,7 @@ impl<'a> BatchAggregator<'a> {
                 aggregation_start_time: self.aggregation_start.timestamp_millis(),
                 aggregation_end_time: self.aggregation_end.timestamp_millis(),
                 packet_file_digest: invalid_packets_digest.as_ref().to_vec(),
+                total_individual_clients,
```
tgeoghegan, line 130: Could you please update the test in `tests/integration_tests.rs` to check that this value is handled correctly?

tgeoghegan (changes requested):
It looks like this doesn't build at the moment.

tgeoghegan (approved):

PR #61, opened 2020-10-19
Move transport backends to their own modules

yuriks:
`transport.rs` was getting quite cluttered and hard to navigate, so move the implementation of each backend to its own module.

Along the way I noticed clippy in CI wasn't running against tests so fix that too.

I recommend reviewing individual commits since the first one is mostly just a bulk move + trimming imports.


aaomidi (approved):
LGTM

tgeoghegan (approved):

PR #63, opened 2020-10-19
K8s compatibility layer for the deploy tool

aaomidi:
This allows direct integration with the k8s secret storage system if we decide we want this to run inside k8s.


aaomidi:
Will make this part of the larger component review.

PR #65, opened 2020-10-20
terraform: create sum part writer service account

tgeoghegan:
As of just recently, MITRE/the portal server has assumed responsibility
for ownership of the buckets to which data share processors shall write
sum part batches. To enable them to grant write permissions to our
servers, we create a GCP service account, and advertise it from our
global manifest. We then allow all the service accounts we create for
individual data share processor instances to impersonate that account.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -89,10 +89,11 @@ provider "kubernetes" {
 }
 
 module "manifest" {
-  source      = "./modules/manifest"
-  environment = var.environment
-  gcp_region  = var.gcp_region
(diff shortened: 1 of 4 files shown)


tgeoghegan:
Example of the global manifest generated with these changes:
```
{
    "format": 0,
    "server-identity": {
        "aws-account-id": 338276578713,
        "gcp-service-account-email": "prio-demo-gcp-sum-writer@prio-bringup-290620.iam.gserviceaccount.com"
    }
}
```

yuriks (approved):

PR #67, opened 2020-10-20
Reintroduce signature message and adopt it

tgeoghegan:
When evaluating a batch, a data share processor needs the identifier of the key it was signed with so it can look up the right public key in the peer's manifest. These commits reintroduce that Avro structure and teaches facilitator to use it when reading and writing batches.

Addresses #25

### DIFF
--- a/avro-schema/batch-signature.avsc
+++ b/avro-schema/batch-signature.avsc
@@ -1,0 +1,17 @@
+{
+    "namespace": "org.abetterinternet.prio.v1",
+    "type": "record",
+    "name": "PrioBatchSignature",
+    "fields": [
+        {
+            "name": "batch_header_signature",
+            "type": "bytes",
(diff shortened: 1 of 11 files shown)


avro-schema/batch-signature.avsc
```diff
@@ -0,0 +1,17 @@
+{
+    "namespace": "org.abetterinternet.prio.v1",
+    "type": "record",
+    "name": "PrioBatchSignature",
+    "fields": [
+        {
+            "name": "batch_header_signature",
+            "type": "bytes",
+            "doc": "signature of the Avro encoded header object"
```
yuriks, line 9: Document the format of this like you mentioned.

yuriks (approved):

aaomidi (approved):

PR #79, opened 2020-10-22
CI build for deploy-tool

aaomidi:
Continuous integration for the deploy tool


deploy-tool/cert/kubernetes_storage_test.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 func TestCleanKey(t *testing.T) {
-	prefix := "cm.k8s."
+	prefix := "acme."
```
tgeoghegan, line 11: I assume this test was failing without this change? Always good to prove the test works!

deploy-tool/.gitignore
```diff
@@ -1,2 +1,4 @@
 deploy-tool
 config.toml
+vendor/
+coverage.coverprofile
```
tgeoghegan, line 4: No newline at end of file

.github/workflows/ci-build.yml
```diff
@@ -56,3 +56,35 @@ jobs:
...
+    - name: Test
+      run: go test -race --coverprofile=coverage.coverprofile --covermode=atomic ./...
+      working_directory: deploy-tool
+    
+    - name: Upload test coverage
+      if: success()
+      uses: codecov/codecov-action@v1
+      with:
+        fail_ci_if_error: false
+        token:
```
tgeoghegan, line 90: No newline, and also it seems like we must need to provide a token here? Possibly `${{ secrets.GITHUB_TOKEN }}` is what we want?

.github/workflows/ci-build.yml
```diff
@@ -56,3 +56,35 @@ jobs:
...
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.15
+    
+    - name: Get dependencies
+      run: go get -v -t -d ./...
+      working_directory: deploy-tool
+    
+    - name: Get linter
+      run: go install golang.org/x/lint/golint
```
tgeoghegan, line 75: Do we run this linter after installing it? Is that what `go vet` does?

tgeoghegan (changes requested):
It looks like there are syntax problems with the Github Action that must be resolved.

deploy-tool/cert/kubernetes_storage_test.go
aaomidi: Yep!

.github/workflows/ci-build.yml
aaomidi: I was looking at a few other places and they...just leave the token blank. Unfortunately I can't test it until its merged it seems :(

.github/workflows/ci-build.yml
aaomidi: ...I don't know how I forgot this.

.github/workflows/ci-build.yml
aaomidi: Nevermind I think I can test it.

tgeoghegan (approved):

PR #80, opened 2020-10-22
Support Google DNS in Deploy Tool

aaomidi:


deploy-tool/dns/dns.go
```diff
@@ -2,23 +2,33 @@ package dns
...
 			APIToken: deployConfig.DNS.CloudflareConfig.APIKey,
 		}
 
+		return provider, nil
+
+	case "gcp":
+		if deployConfig.DNS.GCPConfig == nil {
+			return nil, fmt.Errorf("gcp configuration of the configuration file was nil")
+		}
+		provider, _ := gcloud.NewGoogleDNSProvider(deployConfig.DNS.GCPConfig.Project, deployConfig.DNS.GCPConfig.ZoneMapping)
```
tgeoghegan, line 30: Are we dropping an error here? The enclosing function returns an error so we should be able to propagate the one returned from `NewGoogleDNSProvider` normally.

deploy-tool/dns/gcloud/gcloud.go
```diff
@@ -0,0 +1,169 @@
...
+		project:      project,
+		zoneMappings: zoneMappings,
+	}, nil
+}
+
+func convertRecordType(record libdns.Record) dns.ResourceRecordSet {
+	return dns.ResourceRecordSet{
+		Name:    fmt.Sprintf("%s.", record.Name),
+		Rrdatas: []string{record.Value},
+		Ttl:     record.TTL.Nanoseconds() / 1e9, // Expected in seconds
```
tgeoghegan, line 48: nit: seems like you could use `time.Duration.Seconds()` and cast to the appropriate integer type.

deploy-tool/dns/gcloud/gcloud.go
```diff
@@ -0,0 +1,169 @@
...
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return recs, nil
+}
+
+// Poll is a blocking polling for a given function with a given name. The poller must return true for the polling code to stop. The poller may also return an optional error to be reported
```
tgeoghegan, line 145: Suggest wrapping this comment at 80 or 100 chars. Also does this function need to be exported from the module? If not, rename it `poll`.

deploy-tool/dns/gcloud/gcloud.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+
+		stop, err := poller()
+
+		if stop {
+			return nil
+		}
+
+		if err != nil {
+			lastErr = err
```
tgeoghegan, line 164: If the poller returns an error, should we just bail out right away?

deploy-tool/dns/gcloud/gcloud.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return recs, nil
+}
+
+// DeleteRecords deletes DNS records from a given zone
+func (p *GoogleDNSProvider) DeleteRecords(ctx context.Context, zone string, recs []libdns.Record) ([]libdns.Record, error) {
```
tgeoghegan, line 105: Out of curiosity: does certmagic use this to clean up DNS records after satisfying an ACME DNS challenge?

deploy-tool/dns/gcloud/gcloud.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if result.Status == changeStatusDone {
+		return recs, nil
+	}
+
+	changeID := result.Id
+	err = Poll("delete records", 30*time.Second, 3*time.Second, func() (bool, error) {
```
tgeoghegan, line 126: I don't love the polling but I guess there's no way to block until the new DNS records are in place?

tgeoghegan (changes requested):

deploy-tool/dns/dns.go
aaomidi: Ah yikes on me there

deploy-tool/dns/gcloud/gcloud.go
aaomidi: I didn't like how seconds was implemented on this: 
image

And it seemed like since everything is basing it off of nanoseconds then this one won't hurt either. I'm not sure if I'm comfortable letting the time be set by Go's time conversions since I don't know enough about their edge cases.

deploy-tool/dns/gcloud/gcloud.go
aaomidi: No the response may be an error type but the function wants it to keep retrying - if the function wants it to quit on error it could send a `true, err` as its response

deploy-tool/dns/gcloud/gcloud.go
aaomidi: Yes

deploy-tool/dns/gcloud/gcloud.go
aaomidi: Google does reserve the right to send `pending` https://cloud.google.com/dns/docs/reference/v1/changes#resource, I haven't seen them actually send it but there is __technically__ a chance

tgeoghegan (approved):

PR #84, opened 2020-10-23
terraform: Google Cloud DNS for manifest domain

tgeoghegan:
Up until now, we had been using jrenken's donated domain portcull.is to
serve manifests. Amir bought us isrg-prio.org the other day, so we can
use the <env>.manifests.isrg-prio.org namespace now. Additionally, we
can now use a GCP managed DNS zone for manifests.isrg-prio.org. and
use Terraform to create A records in there for specific envs.

### DIFF
--- a/terraform/variables/demo-gcp.tfvars
+++ b/terraform/variables/demo-gcp.tfvars
@@ -4,10 +4,13 @@ gcp_project                = "prio-bringup-290620"
 machine_type               = "e2-small"
 peer_share_processor_names = ["test-pha-1", "test-pha-2"]
 aws_region                 = "us-west-1"
-# Graciously donated by jrenken
-manifest_domain = "portcull.is"
-ingestors = {
-  ingestor-1 = "portcull.is/ingestor-1"
-  ingestor-2 = "portcull.is/ingestor-2"
+manifest_domain            = "isrg-prio.org"
+managed_dns_zone = {
+  name        = "manifests"
+  gcp_project = "prio-bringup-290620"
 }
-peer_share_processor_manifest_domain = "portcull.is/pha-servers"
+ingestors = {
+  ingestor-1 = "storage.googleapis.com/prio-demo-gcp-manifests/ingestor-1"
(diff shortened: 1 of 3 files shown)


jsha (approved):
This looks good to me, but I am not proficient enough in terraform yet that you should take my word for it.

terraform/variables/demo-gcp.tfvars
```diff
@@ -4,10 +4,13 @@ gcp_project                = "prio-bringup-290620"
 machine_type               = "e2-small"
 peer_share_processor_names = ["test-pha-1", "test-pha-2"]
 aws_region                 = "us-west-1"
-# Graciously donated by jrenken
-manifest_domain = "portcull.is"
+manifest_domain            = "isrg-prio.org"
+managed_dns_zone = {
+  name        = "manifests"
+  gcp_project = "prio-bringup-290620"
```
aaomidi, line 10: Would it make sense for this (specifically gcp_project) to be a top level setting? Are we going to need to use gcp_project elsewhere?

aaomidi (approved):

terraform/variables/demo-gcp.tfvars
tgeoghegan: There is a top level variable for `gcp_project`. The rationale here, which I called out in comment in `manifest.tf`, is that the Google managed DNS zone object might not be in the same GCP project as the ENV you are deploying. Specifically, you created https://console.cloud.google.com/net-services/dns/zones/certificates/details?project=prio-bringup-290620 in the `prio-bringup-290620` project. As I understand it, we can only have one managed zone globally for `manifests.isrg-prio.org.`, so if @jsha wants to issue certs from his deployment in the project he started, or if we spin up a new project for the eventual `prod-us` deployment, we need those other envs to be able to create DNS records in a managed zone in a different project, which is why I provide that here, separately from the top-level `gcp_project` value. I haven't yet tested creating DNS records across GCP projects in the manner I describe, though.

PR #85, opened 2020-10-23
Add prio-workflow Graphviz diagram

yuriks:
I'm committing a pre-built version also so it can be conveniently linked
to if necessary.


tgeoghegan (approved):
Nice, thank you! Is there any way to embed the SVG in the top level README?

yuriks:
I think just including it as an image should work. I'll try that after merging.

PR #88, opened 2020-10-25
facilitator: Add GCSTransport

tgeoghegan:
Adds an implementation of the Transport trait which implements
authenticated reads and writes from Google Cloud Storage buckets. It
supports two auth scenarios:
 - Using the default Oauth token retrieved from the GKE metadata service
 - Obtaining an Oauth token to impersonate another service account
   specified in GCSTransport::new.

The former case will allow ISRG's GCP workload deployment to read and
write from its own GCS buckets. The latter will allow us to write to the
MITRE Corporation's sum part buckets.

This also refactors some argument handling at the facilitator.rs level to
make it possible to specify on a per-transport basis how to do S3
authentication (for "s3://" style transports) or whether to impersonate
a GCP service account (for "gs://" style transports). This would allow
us to e.g., do OIDC auth to read from an S3 ingestion bucket, then use
the default service account to read and write from a validation bucket,
and finally impersonate the sum part bucket writer SA to emit sum parts.

### DIFF
--- a/facilitator/src/transport/gcs.rs
+++ b/facilitator/src/transport/gcs.rs
@@ -361,6 +361,12 @@ impl StreamingTransferWriter {
             return Ok(());
         }
 
+        if !last_chunk && self.buffer.len() < self.minimum_upload_chunk_size {
+            return Err(anyhow!(
+                "insufficient content accumulated in buffer to upload chunk"
+            ));
+        }
+
         // We may not yet be at the last chunk, and hence may not know the total
         // length of eventual complete object, but this variable contains the
         // total length of all the bytes fed into write(), even if we haven't
@@ -371,15 +377,20 @@ impl StreamingTransferWriter {
         // should include the total object size, but otherwise should have * to
         // indicate to GCS that there is an unknown further amount to come.
         // https://cloud.google.com/storage/docs/streaming#streaming_uploads
-        let content_range_header_total_length_field = if last_chunk {
-            format!("{}", object_length_thus_far)
-        } else {
-            "*".to_owned()
-        };
+        let (body, content_range_header_total_length_field) =
+            if last_chunk && self.buffer.len() < self.minimum_upload_chunk_size {
+                (self.buffer.as_ref(), format!("{}", object_length_thus_far))
+            } else {
+                (
+                    &self.buffer[..self.minimum_upload_chunk_size],
+                    "*".to_owned(),
+                )
+            };
+
         let content_range = format!(
             "bytes {}-{}/{}",
             self.object_upload_position,
-            object_length_thus_far - 1,
+            self.object_upload_position + body.len() - 1,
             content_range_header_total_length_field
         );
 
@@ -388,7 +399,7 @@ impl StreamingTransferWriter {
             // By default, ureq will wait forever to connect or read
             .timeout_connect(10_000) // ten seconds
             .timeout_read(10_000) // ten seconds
-            .send_bytes(&self.buffer);
+            .send_bytes(body);
 
         // On success we expect HTTP 308 Resume Incomplete and a Range: header,
         // unless this is the last part and the server accepts the entire
@@ -419,6 +430,13 @@ impl StreamingTransferWriter {
                     ))?
                     .parse::<usize>()
                     .context("End in range header {} not a valid usize")?;
+                // end is usize and so parse would fail if the value in the
+                // header was negative, but we still defend ourselves against
+                // it being bigger than is possible given our position in the
+                // object.
+                if end >= object_length_thus_far {
+                    return Err(anyhow!("End in range header {} is invalid", range_header));
+                }
 
                 // If we have a little content left over, we can't just make
                 // another request, because if there's too little of it, Google
@@ -568,27 +586,27 @@ mod tests {
         mocked_post.assert();
 
         let first_mocked_put = mock("PUT", "/fake-session-uri")
-            .match_header("Content-Length", "10")
-            .match_header("Content-Range", "bytes 0-9/*")
-            .match_body("0123456789")
+            .match_header("Content-Length", "4")
+            .match_header("Content-Range", "bytes 0-3/*")
+            .match_body("0123")
             .with_status(308)
             .with_header("Range", "bytes=0-3")
             .expect_at_most(1)
             .create();
 
         let second_mocked_put = mock("PUT", "/fake-session-uri")
-            .match_header("Content-Length", "6")
-            .match_header("Content-Range", "bytes 4-9/*")
-            .match_body("456789")
+            .match_header("Content-Length", "4")
+            .match_header("Content-Range", "bytes 4-7/*")
+            .match_body("4567")
             .with_status(308)
-            .with_header("Range", "bytes=0-7")
+            .with_header("Range", "bytes=0-6")
             .expect_at_most(1)
             .create();
 
         let final_mocked_put = mock("PUT", "/fake-session-uri")
-            .match_header("Content-Length", "2")
-            .match_header("Content-Range", "bytes 8-9/10")
-            .match_body("89")
+            .match_header("Content-Length", "3")
+            .match_header("Content-Range", "bytes 7-9/10")
+            .match_body("789")
             .with_status(200)
             .expect_at_most(1)
             .create();
--- a/facilitator/src/transport/gcs.rs
+++ b/facilitator/src/transport/gcs.rs
@@ -367,19 +367,16 @@ impl StreamingTransferWriter {
             ));
         }
 
-        // We may not yet be at the last chunk, and hence may not know the total
-        // length of eventual complete object, but this variable contains the
-        // total length of all the bytes fed into write(), even if we haven't
-        // uploaded them all yet.
-        let object_length_thus_far = self.object_upload_position + self.buffer.len();
-
         // When this is the last piece being uploaded, the Content-Range header
         // should include the total object size, but otherwise should have * to
         // indicate to GCS that there is an unknown further amount to come.
         // https://cloud.google.com/storage/docs/streaming#streaming_uploads
         let (body, content_range_header_total_length_field) =
             if last_chunk && self.buffer.len() < self.minimum_upload_chunk_size {
-                (self.buffer.as_ref(), format!("{}", object_length_thus_far))
+                (
+                    self.buffer.as_ref(),
+                    format!("{}", self.object_upload_position + self.buffer.len()),
+                )
             } else {
                 (
                     &self.buffer[..self.minimum_upload_chunk_size],
@@ -432,9 +429,12 @@ impl StreamingTransferWriter {
                     .context("End in range header {} not a valid usize")?;
                 // end is usize and so parse would fail if the value in the
                 // header was negative, but we still defend ourselves against
-                // it being bigger than is possible given our position in the
-                // object.
-                if end >= object_length_thus_far {
+                // it being less than it was before this chunk was uploaded, or
+                // being bigger than is possible given our position in the
+                // overall object.
+                if end < self.object_upload_position
+                    || end > self.object_upload_position + body.len() - 1
+                {
                     return Err(anyhow!("End in range header {} is invalid", range_header));
                 }
 
@@ -443,8 +443,7 @@ impl StreamingTransferWriter {
                 // will reject it. Instead, leave the portion of the chunk that
                 // we didn't manage to upload back in self.buffer so it can be
                 // handled by a subsequent call to upload_chunk.
-                let new_buf = self.buffer.split_off(end + 1 - self.object_upload_position);
-                self.buffer = new_buf;
+                self.buffer = self.buffer.split_off(end + 1 - self.object_upload_position);
                 self.object_upload_position = end + 1;
                 Ok(())
             }
--- a/facilitator/src/transport/gcs.rs
+++ b/facilitator/src/transport/gcs.rs
@@ -7,15 +7,15 @@ use anyhow::{anyhow, Context, Result};
 use chrono::{prelude::Utc, DateTime, Duration};
 use serde::Deserialize;
 use std::{
-    io::{Cursor, Read, Write},
-    mem,
+    io,
+    io::{Read, Write},
 };
 
 const STORAGE_API_BASE_URL: &str = "https://storage.googleapis.com";
 const DEFAULT_OAUTH_TOKEN_URL: &str =
     "http://metadata.google.internal:80/computeMetadata/v1/instance/service-accounts/default/token";
 
-// A wrapper around an Oauth token and its expiration date.
+/// A wrapper around an Oauth token and its expiration date.
 #[derive(Debug)]
 struct OauthToken {
     token: String,
@@ -23,15 +23,15 @@ struct OauthToken {
 }
 
 impl OauthToken {
-    // Returns true if the token is not yet expired.
+    /// Returns true if the token is not yet expired.
     fn expired(&self) -> bool {
         Utc::now() < self.expiration
     }
 }
 
-// Represents the response from a GET request to the GKE metadata service's
-// service account token endpoint. Structure is derived from empirical
-// observation of the JSON scraped from inside a GKE job.
+/// Represents the response from a GET request to the GKE metadata service's
+/// service account token endpoint. Structure is derived from empirical
+/// observation of the JSON scraped from inside a GKE job.
 #[derive(Debug, Deserialize, PartialEq)]
 struct MetadataServiceTokenResponse {
     access_token: String,
@@ -39,9 +39,9 @@ struct MetadataServiceTokenResponse {
     token_type: String,
 }
 
-// Represents the response from a POST request to the GCP IAM service's
-// generateAccessToken endpoint.
-// https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken
+/// Represents the response from a POST request to the GCP IAM service's
+/// generateAccessToken endpoint.
+/// https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken
 #[derive(Debug, Deserialize, PartialEq)]
 #[serde(rename_all = "camelCase")]
 struct GenerateAccessTokenResponse {
@@ -49,18 +49,27 @@ struct GenerateAccessTokenResponse {
     expire_time: DateTime<Utc>,
 }
 
-// OauthTokenProvider manages a default service account Oauth token (i.e. the
-// one for a GCP service account mapped to a Kubernetes service account) and an
-// Oauth token used to impersonate another service account.
+/// OauthTokenProvider manages a default service account Oauth token (i.e. the
+/// one for a GCP service account mapped to a Kubernetes service account) and an
+/// Oauth token used to impersonate another service account.
 struct OauthTokenProvider {
+    /// Holds the service account email to impersonate, if one was provided to
+    /// OauthTokenProvider::new.
     service_account_to_impersonate: Option<String>,
+    /// This field is None after instantiation and is Some after the first
+    /// successful request for a token for the default service account, though
+    /// the contained token may be expired.
     default_service_account_oauth_token: Option<OauthToken>,
+    /// This field is None after instantiation and is Some after the first
+    /// successful request for a token for the impersonated service account,
+    /// though the contained token may be expired. This will always be None if
+    /// service_account_to_impersonate is None.
     impersonated_service_account_oauth_token: Option<OauthToken>,
 }
 
 impl OauthTokenProvider {
-    // Creates a token provider which can impersonate the specified service
-    // account.
+    /// Creates a token provider which can impersonate the specified service
+    /// account.
     fn new(service_account_to_impersonate: Option<String>) -> OauthTokenProvider {
         OauthTokenProvider {
             service_account_to_impersonate,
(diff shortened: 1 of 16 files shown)


facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+// generateAccessToken endpoint.
+// https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+struct GenerateAccessTokenResponse {
+    access_token: String,
+    expire_time: DateTime<Utc>,
+}
+
+// OauthTokenProvider manages a default service account Oauth token (i.e. the
```
jsha, line 52: This should document the allowable values for these Options. Can they all be None? Are there any relationships such that when one of these Options is Some we expect another of them to be Some?

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+
+    fn ensure_storage_access_oauth_token(&mut self) -> Result<String> {
+        match self.service_account_to_impersonate {
+            Some(_) => self.ensure_impersonated_service_account_oauth_token(),
+            None => self.ensure_default_service_account_oauth_token(),
+        }
+    }
+
+    // Obtains an Oauth token for the default service account, if one is not
+    // already held or if it has expired. Otherwise returns a copy of the token.
```
jsha, line 80: The wording here is pretty confusing - what token is it referring to in the "otherwise" clause?

How about:

Returns the current OAuth token for the default service account, if it is valid. Otherwise obtains and returns a new one.

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+    // The returned value is an owned reference because the token owned by this
+    // struct could change while the caller is still holding the returned token.
+    fn ensure_default_service_account_oauth_token(&mut self) -> Result<String> {
+        match &self.default_service_account_oauth_token {
+            Some(token) if !token.expired() => Ok(token.token.clone()),
```
jsha, line 85: I would write this as `=> return Ok` and terminate the `match` clause there. That lets you outdent the whole clause and makes things easier to read IMO.

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+                });
+
+                Ok(response.access_token)
+            }
+        }
+    }
+
+    // Obtains an Oauth token to impersonate the service account this provider
+    // was set up with, if one is not already held or it has expired. Otherwise
+    // reutrns a copy of the valid token.
```
jsha, line 121:
```suggestion
    // returns a copy of the valid token.
```

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+                    token: response.access_token.clone(),
+                    expiration: Utc::now() + Duration::seconds(response.expires_in),
+                });
+
+                Ok(response.access_token)
+            }
+        }
+    }
+
+    // Obtains an Oauth token to impersonate the service account this provider
```
jsha, line 119: Same comment as above; I think it's clearer to rewrite this in the "Returns X, and gets X if needed" form.

That said, it looks like this function doesn't actually have a short-circuit to return early if there's already an impersonation token.

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+// only authentication it supports is by means of impersonating a specified GCP
+// service account, which is assumed to have the necessary read or write
+// permissions on the bucket.
+pub struct GCSTransport {
+    path: GCSPath,
+    oauth_token_provider: OauthTokenProvider,
+}
+
+impl GCSTransport {
+    pub fn new(path: GCSPath, impersonate: Option<String>) -> GCSTransport {
```
jsha, line 182: This needs documentation for what the value of `impersonate` should be. Also, it looks like it's an error to pass None here; do I understand correctly? Maybe we should make this a `String` if so.

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+            STORAGE_API_BASE_URL, self.path.bucket, encoded_key
+        );
+
+        let response = ureq::get(&url)
+            // Ensures response body will be content and not JSON metadata.
+            // https://cloud.google.com/storage/docs/json_api/v1/objects/get#parameters
+            .query("alt", "media")
+            .set(
+                "Authorization",
+                &format!(
```
jsha, line 206: Style: when I find myself with cascading fields like this I try to break them out into variables, e.g.

```
let auth = format!(...);
...
.set("Authorization", auth);
```

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+struct StreamingTransferWriter {
+    upload_session_uri: String,
+    minimum_upload_chunk_size: usize,
+    object_upload_position: usize,
+    buffer: Vec<u8>,
+}
+
+impl StreamingTransferWriter {
+    fn new(
+        bucket: String,
```
jsha, line 279: This needs documentation of what these fields should be. For instance, is `object` is an object key or a full path to an object? Is it the destination or the source?

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+        object: String,
+        minimum_upload_chunk_size: usize,
```
jsha, line 281: It seems unlikely clients will want to configure this differently than the default 8MB you are currently passing in, and choosing this value relies on knowledge of the upload API that you are containing within this struct. I would recommend just making that a const.

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+    fn new_with_api_url(
+        bucket: String,
+        object: String,
+        minimum_upload_chunk_size: usize,
+        oauth_token: String,
+        storage_api_base_url: &str,
+    ) -> Result<StreamingTransferWriter> {
+        // Initiate the resumable, streaming upload. Unlike the GET side, the
+        // key is provided as a query parameter here, and thus doesn't need to
+        // be URL encoded.
```
jsha, line 302: I think that's incorrect. According to https://cloud.google.com/storage/docs/request-endpoints#encoding, they expect encoding for query parameters too (which is what I would expect; query parameters are always escaped in my experience). But I think the set of parameters to be encoded is different based on context. Probably this should use the `.query(..)` method on ureq.

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+        // https://cloud.google.com/storage/docs/performing-resumable-uploads#initiate-session
+        let upload_url = format!(
+            "{}/upload/storage/v1/b/{}/o/?uploadType=resumable&name={}",
+            storage_api_base_url, bucket, object
+        );
+        let http_response = ureq::post(&upload_url)
+            .set("Authorization", &format!("Bearer {}", oauth_token))
+            .set("Content-Length", "0")
```
jsha, line 310: Nit: You don't need to set `Content-Length: 0` if you use `.send_bytes(vec![])` instead of `.call()`.

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+                http_response
+            ));
+        }
+
+        // The upload session URI authenticates subsequent upload requests for
+        // this upload, so we no longer need the impersonated service account's
+        // Oauth token. Session URIs are valid for a week, which should be more
+        // than enough for any upload we perform.
+        // https://cloud.google.com/storage/docs/resumable-uploads#session-uris
+        let upload_session_uri = match http_response.header("Location") {
```
jsha, line 327: I think this could be:

```rust
let upload_session_uri = http_response.header("Location").unwrap_or_else(|| {
  Err(anyhow!("no Location header in response when initiating streaming transfer"))
})?;
```

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+            return Ok(());
+        }
+
+        let chunk = mem::replace(
+            &mut self.buffer,
+            Vec::with_capacity(self.minimum_upload_chunk_size * 2),
+        );
+        let chunk_len = chunk.len();
+        // We may not yet be at the last chunk, and hence may not know the total
+        // length of eventual complete object, but this variable the total
```
jsha, line 355:
```suggestion
        // length of eventual complete object, but this variable contains the total
```

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+        // https://cloud.google.com/storage/docs/streaming#streaming_uploads
+        let content_range_header_total_length_field = if last_chunk {
+            format!("{}", object_length_thus_far)
+        } else {
+            "*".to_owned()
+        };
+        let content_range = format!(
+            "bytes {}-{}/{}",
+            self.object_upload_position,
+            object_length_thus_far - 1,
```
jsha, line 372: I assume you've manually tested this to upload a >8MB file and confirmed it has the same hash after upload? Just wondering whether I should dig into double checking for off-by-one errors in the offsets here.

facilitator/src/transport/gcs.rs
tgeoghegan: I had to make it `if let Some(token) = &self.default_service_account_oauth_token { ... }` and then put the `!token.expired()` check inside the `if` block, because the compiler says we don't yet have https://github.com/rust-lang/rust/issues/53667. But you're right, it's much neater this way.

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+            content_range_header_total_length_field
+        );
+
+        // ureq::Request::send takes possession of the argument so we must clone
+        // here, in case the entire chunk doesn't get uploaded successfully.
+        let body = Cursor::new(chunk.clone());
+
+        let http_response = ureq::put(&self.upload_session_uri)
+            .set("Content-Length", &format!("{}", chunk_len))
```
jsha, line 381: If you use `.send_bytes(&body)` instead of `.send(body)`, it will set Content-Length for you (and not take ownership).

facilitator/src/transport/gcs.rs
tgeoghegan: It's not optional, because we also support authenticating as the default service account (the one whose credentials are available in the GKE metadata service). We support the impersonation case so that our per-data share processor GCP service accounts can impersonate the singleton sum part bucket writer service account and write sum parts to the MITRE owned buckets. We support the default SA case so that once our own storage moves to GCS, we can bless our per-data share processor SAs to use the buckets we create. I put in a doccomment.

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+            .timeout_read(10_000) // ten seconds
+            .send(body);
+
+        // On success we expect HTTP 308 Resume Incomplete and a Range: header,
+        // unless this is the last part and the server accepts the entire
+        // provided Content-Range, in which case it's HTTP 200, or 201 (?).
+        // https://cloud.google.com/storage/docs/performing-resumable-uploads#chunked-upload
+        match http_response.status() {
+            200 | 201 => {
+                if !last_chunk {
```
jsha, line 394: In a big `match` like this, having a fallthrough makes it hard to reason about. I'd make the `match` the final item in the function, and split this into two clauses:

```
200 | 201 if last_chunk => Ok(()),
200 | 201 if !last_chunk => Err(...chunks remaining...),
```

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+                    return Err(anyhow!(
+                        "received HTTP 200 or 201 response with chunks remaining"
+                    ));
+                }
+            }
+            308 => {
+                if !http_response.has("Range") {
```
jsha, line 401: Like the refactoring I suggested above, you can do:

```
308 if !http_response.has("Range") => Err(...no range...)
308 => {
  // do the work
}
```

facilitator/src/transport/gcs.rs
tgeoghegan: I will refactor this so that `new` uses a sensible default and `new_with_api_url` takes a size so that tests can work with smaller buffers.

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+                        http_response.into_string()
+                    ));
+                }
+                let range_header = http_response.header("Range").unwrap();
+                // The range header is like "bytes=111-222", and represents a
+                // range in the overall object, not the current chunk.
+                let range = range_header
+                    .strip_prefix("bytes=")
+                    .context(format!("unexpected Range header {}", range_header))?;
+                let mut values = range.splitn(2, '-');
```
jsha, line 413: I think it would simplify the following code to collect into a Vec:

```
let values: Vec<_> = range.splitn(2, '-').collect();
if values.len() != 2 {
 // return error
}
let start: usize = values[0].parse().context("unexpected ...
```

facilitator/src/transport/gcs.rs
tgeoghegan: You're right, the documentation is clear. It's funny, because empirically, the GCS API is fine with query params on this PUT that contain unencoded `-`, but the GET API rejected them. I'll change this to encode the object name regardless.

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+                if start != self.object_upload_position {
+                    return Err(anyhow!("unexpected Range header {}", range_header));
+                }
+
+                self.object_upload_position = end + 1;
+
+                // If we have a little content left over, we can't just make
+                // another request, because if there's too little of it, Google
+                // will reject it. Instead, put the portion of the chunk that we
+                // didn't manage to upload back into self.buffer so it can be
```
jsha, line 434: I think this use case is a great match for `VecDeque`. You can `.send_bytes()` on an appropriately-sized slice of the VecDeque, then `pop_front` however many bytes actually got written.

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+        Ok(())
+    }
+}
+
+impl Write for StreamingTransferWriter {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        // Write into memory buffer, and upload to GCS if we have accumulated
+        // enough content
+        self.buffer.extend_from_slice(buf);
+        if self.buffer.len() >= self.minimum_upload_chunk_size {
```
jsha, line 458: Should be `while`, I think.

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+            self.upload_chunk(false).map_err(|e| {
+                std::io::Error::new(std::io::ErrorKind::Other, Error::AnyhowError(e))
```
jsha, line 460: Style nit: `use std::io;` at the top of the program, and call these `io::Error::new(..)`

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+            })?;
+        }
+
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
```
jsha, line 468: This should call `self.upload_chunk()` with whatever's in the buffer (or add a TODO, or documentation that this is unflushable).

facilitator/src/transport/gcs.rs
```diff
@@ -0,0 +1,601 @@
...
+
+    fn cancel_upload(&mut self) -> Result<()> {
+        // https://cloud.google.com/storage/docs/performing-resumable-uploads#cancel-upload
+        let http_response = ureq::delete(&self.upload_session_uri)
+            .set("Content-Length", "0")
+            // By default, ureq will wait forever to connect or read
+            .timeout_connect(10_000) // ten seconds
+            .timeout_read(10_000) // ten seconds
+            .call();
+        if http_response.status() != 499 {
```
jsha, line 488: I am surprised GCS signals success with a 499! 🤷🏻 

Slight style nit: I would probably make this a final match statement:

```
match http_response.status() {
  499 => Err(..),
  _ => Ok(()),
}
```

facilitator/src/transport/gcs.rs
tgeoghegan: You're right, but `anyhow::Context` makes this even nicer.

facilitator/src/transport/gcs.rs
tgeoghegan: Yes, and the test, below, exercises an upload split into two chunks, though I also shall retest manually after pushing fixes to your various observations.

facilitator/src/transport/gcs.rs
tgeoghegan: Nope! Consider: upload size is 8 MiB, and you have 8 MiB + 100 kB in the buffer. You try to upload the whole buffer, but Google tells you they only accepted 8 MiB of it. If you then try to upload the remaining 100 kB, Google will reject it, because they don't allow chunks less than 256 kB unless it's the last chunk. So we intentionally don't drain the buffer here and only do that in `complete_upload`, where we are able to construct a `Content-Range` header that specifies the total size of the object, in which case Google lets us upload a chunk of less than 256 kilobytes.

jsha: I don't understand. Are you saying the buffer grows indefinitely because it's never shrunk until `complete_upload`? Because this reads to me as if it is replacing buffer with a new zero-length Vec: https://github.com/abetterinternet/prio-server/pull/88/files#diff-465ffcef811e76d63ff9b6f105387456ffff0155b61560ead82aa6d32ef9d7e2R349-R351.

> Google tells you they only accepted 8 MiB of it. If you then try to upload the remaining 100 kB, Google will reject it,

In this case, you would drop out of the `while` loop because buffer.len() (i.e. 100kB) is not greater than minimum_upload_chunk_size.

Imagine the case where you're getting `.write()` calls with 100MB each, but Google is only accepting 50MB per `PUT`. You'd fall behind.

tgeoghegan: Oh, of course, you're right. I changed it to while.

facilitator/src/transport/gcs.rs
tgeoghegan: We now rely on that implicit behaviour of `send_bytes` elsewhere in this file, so for consistency I'll change it as you suggest.

facilitator/src/transport/gcs.rs
```diff
@@ -373,98 +383,81 @@ impl StreamingTransferWriter {
...
+                    ))?
                     .parse::<usize>()
-                    .context(format!("unexpected Range header {}", range_header))?;
-                let end = values
-                    .next()
-                    .context(format!("unexpected Range header {}", range_header))?
-                    .parse::<usize>()
-                    .context(format!("unexpected Range header {}", range_header))?;
-
-                if start != self.object_upload_position {
```
jsha: It looks like the expected semantics of the `start` part of the range changed in the latest commit, from expecting it to be object_upload_position to always being 0. It seems like one version or the other should have resulted in throwing errors?

facilitator/src/transport/gcs.rs
```diff
@@ -373,98 +383,81 @@ impl StreamingTransferWriter {
...
-                    .extend_from_slice(&chunk[self.object_upload_position - start..]);
-            }
-            _ => {
-                return Err(anyhow!(
-                    "failed to upload part to GCS: {} synthetic: {}\n{:?}",
-                    http_response.status(),
-                    http_response.synthetic(),
-                    http_response.into_string()
-                ))
+                let new_buf = self.buffer.split_off(end + 1 - self.object_upload_position);
```
jsha, line 428: Here, and below when we update object_upload_position, we are giving a lot of trust to the `end` value read from the network. A malicious value could crash us. Before we trust it, we should check that it's reasonable - greater than zero and less than or equal to `object_length_thus_far`.

jsha (changes requested):
By the way, one thing I noticed that's a little unusual: The size of chunks that get written depends on the size of `write()` calls and also on the size that Google accepts at any given time. For instance, if I call `write()` with a 50MB buffer, and Google accepts 10MB per PUT, the GCSTransport would try to upload 50MB, then try to upload 40MB, then try to upload 30MB, etc. That's going to be pretty inefficient because we won't get prompt notice if Google has killed a PUT early - and in fact might wind up writing a lot of bytes that don't get used.

My understanding of the linked docs is that you are supposed to pick a chunk size (e.g. 8MB) and always send chunks of that size (plus a final chunk). That also seems like a more predictable way to do things.

facilitator/src/transport/gcs.rs
```diff
@@ -419,6 +430,13 @@ impl StreamingTransferWriter {
                     ))?
                     .parse::<usize>()
                     .context("End in range header {} not a valid usize")?;
+                // end is usize and so parse would fail if the value in the
+                // header was negative, but we still defend ourselves against
+                // it being bigger than is possible given our position in the
+                // object.
+                if end >= object_length_thus_far {
```
jsha, line 437: Now that we've switched up uploading 8MB at a time, I think this check should be against the amount that we actually uploaded - I.e. `self.object_upload_position + body.len() - 1` from above (which can be extracted into a variable). I think `object_length_thus_far` becomes unneeded.

Also, looking at this again, I think we can check for `end > object_upload_position`, i.e. the end of the range should never get smaller. Arguably that could be a `>=`, but if it's `>` that means we made no progress with this chunk, which I think would be some sort of error rather than a 308.

jsha (changes requested):
Looking great! Just a couple more small tweaks.

jsha (approved):
Beautiful. 👨🏻‍🍳 (GitHub doesn't yet have the "chef kiss" emoji so we will have to make do with these inferior tools).

PR #89, opened 2020-10-25
facilitator: improve key parameter handling

tgeoghegan:
This pull request improves facilitator's handling of crypto key arguments, anticipating their usage by the workflow manager that will create Kubernetes jobs.

### DIFF
--- a/facilitator/src/aggregation.rs
+++ b/facilitator/src/aggregation.rs
@@ -9,7 +9,10 @@ use crate::{
 };
 use anyhow::{anyhow, Context, Result};
 use chrono::NaiveDateTime;
-use prio::{encrypt::PrivateKey, server::VerificationMessage};
+use prio::{
+    encrypt::PrivateKey,
+    server::{Server, VerificationMessage},
+};
 use ring::signature::{KeyPair, UnparsedPublicKey, ECDSA_P256_SHA256_ASN1};
 use std::convert::TryFrom;
 use uuid::Uuid;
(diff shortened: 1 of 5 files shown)


facilitator/src/aggregation.rs
```diff
@@ -265,15 +289,34 @@ impl<'a> BatchAggregator<'a> {
...
+                        if !valid {
+                            invalid_uuids.push(peer_validation_packet.uuid);
+                        }
+                    }
+                    Err(e) => {
+                        last_err = Some(Err(e));
+                        continue;
+                    }
+                }
+                did_aggregate_shares = true;
```
jsha, line 310: I think this would read more clearly if you moved `did_aggregate_shares = true` and `break` into the `Ok(valid)` branch since they will only execute in that case.

facilitator/src/batch.rs
```diff
@@ -123,7 +123,7 @@ impl<'a, H: Header, P: Packet> BatchReader<'a, H, P> {
 
     /// Return the parsed header from this batch, but only if its signature is
     /// valid.
-    pub fn header(&self, key: &UnparsedPublicKey<Vec<u8>>) -> Result<H> {
+    pub fn header(&mut self, key: &UnparsedPublicKey<Vec<u8>>) -> Result<H> {
```
jsha, line 126: Why does header need a `mut self` now?

jsha (approved):

facilitator/src/batch.rs
tgeoghegan: The implementation of `Transport::get` in `GCSTransport` has to be `&mut self` because it mutates its `oauth_token_provider` member when it caches an auth token. `Transport::get` should always have taken `&mut self` because it's easy to imagine a reader where doing a read advances a cursor or consumes a stream, so I didn't feel bad about doing that, but Rust will then make you propagate mutness all over the program, kind of like that "is your function red or blue" blog you showed me the other day.

jsha: Interesting. I don't think you generally have to propagate mutness all over the program in the same way that async leads to. But it looks like this is inherited from https://github.com/abetterinternet/prio-server/pull/88, so I'll finish reviewing that and see if I grok it.

PR #92, opened 2020-10-26
Initial Go workflow-manager.

jsha:
This adds a workflow manager that can look for batches in GCS, and can
talk to the Kubernetes API (currently just a demonstration of listing
pods).

This adds a Dockerfile to build the workflow-manager container, and
changes to kubernetes.tf to set up the appropriate role and role
bindings for workflow-manager.

The demonstration cron job used to run `facilitator`; in this PR I
change it to run `workflow-manager`. I tweaked the successful job count
because I felt that made it more explicit how many jobs I should see
sticking around.

This does not yet have any GitHub Actions rigged up to push the
workflow-manager image.


terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -205,15 +201,48 @@ resource "kubernetes_cron_job" "workflow_manager" {
...
 }
 
+resource "kubernetes_role" "workflow-manager-role" {
+  metadata {
+    name      = "${var.environment}-${var.data_share_processor_name}-wfm-role"
+    namespace = var.kubernetes_namespace
+  }
+
+  rule {
+    api_groups = [""]
```
tgeoghegan, line 221: I find it surprising that an empty string here is implicitly the "core" API group, so that's probably worth a comment, we well as one explaining what privileges we're granting this role and why.

terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -205,15 +201,48 @@ resource "kubernetes_cron_job" "workflow_manager" {
...
+    resources  = ["namespaces", "pods"]
+    verbs      = ["get", "list", "watch"]
+  }
+}
+
+resource "kubernetes_role_binding" "workflow-manager-rolebinding" {
```
tgeoghegan, line 227: nit: I like to use snake_case labels (i.e., `workflow_manager_rolebinding` so that then when you're referencing this object elsewhere, you get a consistent dotted identifier like `kubernetes_role_binding.workflow_manager_rolebinding`. However `terraform fmt` has no opinions about this and I'm sure I have some kebab case somewhere in here.

workflow-manager/main.go
```diff
@@ -0,0 +1,126 @@
...
+	"log"
+	"os"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/client-go/rest"
+
+	"cloud.google.com/go/storage"
```
tgeoghegan, line 21: Just to make sure you know: for Narnia the ingestion buckets will be in S3, so I think you'll need to use [that SDK](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/) and a client setup with [stscreds](https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/stscreds/) so that it can assume the appropriate AWS role (hence the `AWS_ROLE_ARN` and `AWS_ACCOUNT_ID` environment variables wired into the cron job).

terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -149,26 +149,22 @@ resource "kubernetes_cron_job" "workflow_manager" {
...
               args = [
-                "generate-ingestion-sample",
-                "--s3-use-credentials-from-gke-metadata",
-                "--aggregation-id", "fake-1",
-                "--batch-id", "eb03ef04-5f05-4a64-95b2-ca1b841b6885",
-                "--date", "2020/09/11/21/11",
-                "--facilitator-output", "s3://${var.ingestion_bucket}",
-                "--pha-output", "/tmp/sample-pha"
+                "--k8s-namespace", var.kubernetes_namespace,
+                "--input-bucket", "lotophagi-ingestion-bucket",
```
tgeoghegan, line 167: I'm guessing `lotophagi-ingestion-bucket` was created manually? Between this and the `prio-workflow-manager` name above (which doesn't exist in [DockerHub](https://hub.docker.com/orgs/letsencrypt/repositories), this module won't work in any env but yours. If it helps keep you moving we should merge this, though.

tgeoghegan (changes requested):

workflow-manager/main.go
jsha: Yep, I'm aware I need to add S3 support; I just figured since we know we'll need GS anyhow, and I'm learning the whole GCS stack, I'd start with that branch of things.

terraform/modules/kubernetes/kubernetes.tf
jsha: I've published my Docker image to `prio-workflow-manager`, and changed this to refer to a variable that exists in the .tf (though I'm not sure it's the *right* variable).

tgeoghegan: It won't be, because it'll be an S3 bucket and your code only supports GCS buckets for now. But that's OK: it'll deploy just fine, and then the jobs will fail, which just generates noise in the GCP console.

tgeoghegan (approved):

PR #94, opened 2020-10-26
Introduce cert-renewals

aaomidi:
- Update configuration file to include "Jobs"
- Update cert.go to be a little more defined about what it expects
- Update main.go to handle jobs alongside stdin

TODO:
- [ ] Actually update the manifests

### DIFF
--- a/deploy-tool/cert/cert.go
+++ b/deploy-tool/cert/cert.go
@@ -14,12 +14,17 @@ import (
 	"github.com/caddyserver/certmagic"
 )
 
-// IssueCertificate asks the Let's Encrypt ACMEApiEndpoint to validate and sign a certificate for a site using the DNS01 strategy
-func IssueCertificate(deployConfig config.DeployConfig, site string, privKey *ecdsa.PrivateKey) (string, error) {
-	dnsProvider, err := dns.GetACMEDNSProvider(deployConfig)
+// CertificateManager is a structure that defines the Certificate Issuance and renewal implementation
+type CertificateManager struct {
+	Config config.DeployConfig
+}
+
+func (certManager *CertificateManager) getCertMagic() (*certmagic.Config, error) {
+	deployConfig := certManager.Config
+	dnsProvider, err := dns.GetACMEDNSProvider(certManager.Config)
 
 	if err != nil {
-		return "", fmt.Errorf("error when getting the dns provider: %v", err)
+		return nil, fmt.Errorf("error when getting the dns provider: %v", err)
 	}
 
 	certmagic.DefaultACME.DNS01Solver = &certmagic.DNS01Solver{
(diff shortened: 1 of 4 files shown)


deploy-tool/cert/cert.go
```diff
@@ -14,12 +14,17 @@ import (
...
 
-// IssueCertificate asks the Let's Encrypt ACMEApiEndpoint to validate and sign a certificate for a site using the DNS01 strategy
-func IssueCertificate(deployConfig config.DeployConfig, site string, privKey *ecdsa.PrivateKey) (string, error) {
-	dnsProvider, err := dns.GetACMEDNSProvider(deployConfig)
+// CertificateManager is a structure that defines the Certificate Issuance and renewal implementation
+type CertificateManager struct {
+	Config config.DeployConfig
+}
+
+func (certManager *CertificateManager) getCertMagic() (*certmagic.Config, error) {
```
tgeoghegan, line 22: Go convention would be to name the receiver `cm` or even just `c`, which then makes it more obvious in the method body that you're referring to "this/self"

deploy-tool/cert/cert.go
```diff
@@ -31,24 +36,15 @@ func IssueCertificate(deployConfig config.DeployConfig, site string, privKey *ec
...
+		return nil, err
 	}
 
-	return string(issued.Certificate), nil
+	return acme, nil
 }
 
-func setStorageDriver(deployConfig config.DeployConfig, acme *certmagic.Config) error {
+func (certManager *CertificateManager) setStorageDriver(acme *certmagic.Config) error {
+	deployConfig := certManager.Config
```
tgeoghegan, line 47: I think assigning a struct member to `deployConfig` and using that throughout the method obscures the fact that the value being accessed here is a member of this object, and not an argument, which I want to be obvious to the reader, especially since a different config object gets passed into the method.

deploy-tool/cert/cert.go
```diff
@@ -71,6 +67,26 @@ func setStorageDriver(deployConfig config.DeployConfig, acme *certmagic.Config)
 	return nil
 }
 
+// IssueCertificate asks the Let's Encrypt ACMEApiEndpoint to validate and sign a certificate for a site using the DNS01 strategy
+func (certManager *CertificateManager) IssueCertificate(site string, privKey *ecdsa.PrivateKey) (string, error) {
```
tgeoghegan, line 71: It's a bit of a shame that this returns `string` because it's hard to guess what that string represents. I think in this case it's the PEM certificate chain. We could define a type alias (not certain if that's very idiomatic in Go) but at least let's document what the returned value is. Also nit: wrap long comment lines.

deploy-tool/main.go
```diff
@@ -162,6 +162,13 @@ func main() {
 		log.Fatalf("failed to parse config.toml: %v", err)
 	}
 
+	if deployConfig.Job.Enable {
+		handleJob(deployConfig)
+		// TODO eventually maybe use the output from this to generalize pushing the changes with gsutil
```
tgeoghegan, line 167: Well now that this tool has a proper maintainer and it's a real Go module with deps and everything, it should properly pull in the GCP SDK and manage bucket contents that way. Same goes for those ghastly `kubectl` invocations; they should be replaced with proper API usage--and then we wouldn't have to write out keys to temp files.

deploy-tool/cert/cert.go
```diff
@@ -14,12 +14,17 @@ import (
 	"github.com/caddyserver/certmagic"
 )
 
-// IssueCertificate asks the Let's Encrypt ACMEApiEndpoint to validate and sign a certificate for a site using the DNS01 strategy
-func IssueCertificate(deployConfig config.DeployConfig, site string, privKey *ecdsa.PrivateKey) (string, error) {
-	dnsProvider, err := dns.GetACMEDNSProvider(deployConfig)
+// CertificateManager is a structure that defines the Certificate Issuance and renewal implementation
+type CertificateManager struct {
+	Config config.DeployConfig
+}
```
tgeoghegan, line 20: Per jsha's advice in Slack, I think there should be a `New` function to instantiate a `CertificateManager` that takes a `config.DeployConfig`. Also I don't think anything outside of the `CertificateManager` implementation ever needs to see its inner config object, so it could be private (i.e., rename to `config`).

tgeoghegan (commented):
I really like the idea of the `CertificateManager` object. We should talk about how we end up making this into a Kubernetes cronjob, though.

deploy-tool/cert/cert.go
mholt: If you mean something like `type CertificatePEM string`, that's very clean Go practice, from what I've seen.

PR #95, opened 2020-10-27
Add a .dockerignore.

jsha:
This decreases the amount of context that has to be sent to the Docker
daemon on each build. Particularly valuable for the large files in the
facilitator/target directory.


.dockerignore
```diff
@@ -0,0 +1,2 @@
+./facilitator/target/
+./terraform
```
tgeoghegan, line 2: nit: no newline at EOF.

tgeoghegan (approved):

PR #96, opened 2020-10-27
Workflow manager updates for Monday

jsha:
The workflow manager can now create jobs.

Clean up the dummy code that counts pods in the cluster.

Embed some build information in workflow-manager to print at startup.

### DIFF
--- a/terraform/modules/kubernetes/kubernetes.tf
+++ b/terraform/modules/kubernetes/kubernetes.tf
@@ -223,7 +223,7 @@ resource "kubernetes_role" "workflow_manager_role" {
 
   rule {
     // API group "" means the core API group.
-    api_groups = [""]
+    api_groups = ["batch", ""]
     // Workflow manager can list pods and create jobs.
     // Note: Some of these permissions will probably wind up not being needed.
     // Starting with a moderately generous demonstration set.
--- a/workflow-manager/main.go
+++ b/workflow-manager/main.go
@@ -13,6 +13,8 @@ import (
 	"os"
 	"strings"
 
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
@@ -21,6 +23,9 @@ import (
 	"cloud.google.com/go/storage"
 	"google.golang.org/api/iterator"
 )
+
+var BuildID string
+var BuildTime string
 
 // batchMap is used to store information about a batch, indexed by the basename
 // of that batch.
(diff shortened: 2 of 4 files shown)


terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -223,7 +223,7 @@ resource "kubernetes_role" "workflow_manager_role" {
 
   rule {
     // API group "" means the core API group.
-    api_groups = [""]
+    api_groups = ["batch", ""]
     // Workflow manager can list pods and create jobs.
     // Note: Some of these permissions will probably wind up not being needed.
     // Starting with a moderately generous demonstration set.
```
tgeoghegan, line 229: I see your comment here about narrowing the permissions later, but I wonder if we can have multiple `rule` blocks so we can use separate sets of nouns and verbs with the `batch` and `""` API groups, or if we have to define distinct `kubernetes_roles` to do that.

workflow-manager/main.go
```diff
@@ -111,16 +110,78 @@ func work(inputBucket, k8sNS string) error {
...
+								"batch-intake",
+								"--s3-use-credentials-from-gke-metadata",
+								"--aggregation-id", "fake-1",
+								"--batch-id", batchID,
+								"--date", batchDate,
+								"--ingestion-bucket", "gs://tbd",
+								"--validation-bucket", "gs://tbd",
+								"--ingestor-public-key", "what_is_my_pubkey",
+								"--ecies-private-key", "what_is_my_privkey",
+								"--share-processor-private-key", "what_is_my_other_privkey",
```
tgeoghegan, line 161: These parameter names are out of date as of #89. In any case, since facilitator can accept these keys as arguments or environment variables, I see two ways to provide batch signing keys to the facilitator:
1. Plumb them into the workflow-manager process via environment variables or args, and then have workflow-manager pass them into the facilitator
2. Have the workflow-manager construct the Go API equivalent of this Terraform block:
```
env {
    name = "BATCH_SIGNING_PRIVATE_KEY"
    value_from {
        secret_key_ref {
            name = $(batch_signing_key_secret_name)
            key  = "signing_key"
        }
    }
}
```
(2) is a little nicer than (1) because it avoids ever exposing the secret to the workflow-manager's address space.

Providing the packet decryption key(s) is trickier, because facilitator needs all available private keys provided as a comma separated list (it needs all available keys provided because we can't tell up front what keys individual packets in a batch will have been encrypted to). We could rig up the Terraform side to provide all the decryption keys to workflow-manager, then have workflow-manager construct the comma-separated list. Also, for Narnia, we can take a shortcut and assume there will only be a single available packet decryption key and thus plug that secret in as an environment variable.

tgeoghegan (approved):
Nothing objectionable here, go ahead and merge if it helps you out.

workflow-manager/main.go
```diff
@@ -111,16 +110,78 @@ func work(inputBucket, k8sNS string) error {
...
+						{
+							Args: []string{
+								"batch-intake",
+								"--s3-use-credentials-from-gke-metadata",
+								"--aggregation-id", "fake-1",
+								"--batch-id", batchID,
+								"--date", batchDate,
+								"--ingestion-bucket", "gs://tbd",
+								"--validation-bucket", "gs://tbd",
+								"--ingestor-public-key", "what_is_my_pubkey",
```
tgeoghegan, line 159: The value for `ingestion-bucket` is the `input-bucket` value this program got.

`validation-bucket` must be retrieved from the peer data share processor's relevant specific manifest. To find it, facilitator will need to be provided their manifest hosting base URL (i.e. wire `var.peer_share_processor_manifest_domain` from the top-level Terraform module into facilitator) and the data share processor name (i.e., wire `var.data_share_processor_name` from data_share_processor.tf into facilitator).

`ingestor-public-key` must be retrieved from the ingestor's global manifest. To find it, facilitator will need to be provided the relevant ingestor's manifest hosting base URL (so wire the appropriate value from the `var.ingestors` map in the top level TF module into the facilitator).

jsha:
Thanks for the review! Going to merge and work on those feedback points as a followup PR.

PR #97, opened 2020-10-27
Speed up Docker build for facilitator

jsha:
Previously, the Docker build was not caching anything, meaning we would download and build all dependencies each time we build the container. This uses some tricks from https://shaneutt.com/blog/rust-fast-small-docker-image-builds/ to break the build process across cacheable Docker layers so rebuilds are much faster. I also switched to stripped debug builds, which are faster than release builds and suitable for our current use cases. Later on we should make debug / release an env variable, so the automatic publish in GH Actions can take its time and do a release build. And I moved from shipping Alpine to a bare binary in the container.

Size went from
580MB (full OS) to
18MB (release binary only) to
26 MB (debug binary only, stripped)

Time went from

4m52 (release, full build) to
2m22 (debug, full build) to
1m8s (release, incremental) to
33s (debug, incremental)

Tested on a 32 vCPU instance in the cloud. Weirdly, the last stat (debug, incremental) was faster on my laptop - 19s.

### DIFF
--- a/facilitator/Dockerfile
+++ b/facilitator/Dockerfile
@@ -1,6 +1,11 @@
 FROM rust:1.46-alpine as builder
 
 RUN apk add libc-dev && apk update
+# Attempt to install a nonexistent package. This triggers
+# updating the crates.io index separately from building the
+# dependencies, so if dependencies change we don't have to
+# re-download the whole index.
+RUN cargo install _update_crates_io_failure_is_expected_ ; true
 
 WORKDIR /usr/src/prio-server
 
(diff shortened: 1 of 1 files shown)


tgeoghegan:
This is clever, and I'm glad you were able to do it all in the Dockerfile and didn't need to introduce any new dummy targets in Cargo.toml. Do you have a sense of how fragile this is? I guess I can't see anything here that's obviously risky.

When you merge, can you please file an issue to introduce and use the release variant of the image?

facilitator/Dockerfile
```diff
@@ -1,16 +1,39 @@
 FROM rust:1.46-alpine as builder
 
 RUN apk add libc-dev && apk update
+# Attempt to install a nonexistent package. This triggers
+# updating the crates.io index separately from building the
+# dependencies, so if dependencies change we don't have to
+# re-download the whole index.
+RUN cargo install _update_crates_io_failure_is_expected_ ; true
```
tgeoghegan, line 8: I'm surprised that `cargo` gets as far as installing all the dependencies before failing because the target doesn't exist. Is that documented behavior of `cargo`?

tgeoghegan (approved):

aaomidi (approved):

facilitator/Dockerfile
jsha: Crates has an "index" - a big git repository containing a JSON file with metadata for every crate in the ecosystem. It downloads the whole thing on first run, and then updates it periodically. This command triggers the first fetch of that index (but not any dependency installs). Otherwise the first fetch would run with the first `cargo build`. It's perhaps a bit of an over-optimization -- it makes full builds when you rebuild the dependencies faster -- but I think it's fine.

The behavior that `cargo install` fetches the index is not documented but is a necessary implication of the design. Cargo can't tell that a package is nonexistent until it has a copy of the index.

jsha:
> Do you have a sense of how fragile this is?

In general Rust's build cache feels fairly fragile to me. I find it rebuilds at the drop of a hat. In part that's been because I was switching between command line builds and vscode builds, which evidently use different build flags and so can't reuse cached objects from each others' builds.

As a concrete example, the two `cargo build` commands have to be exactly the same (but I documented that). Even the `cargo install` command that was originally here wouldn't reuse the build cache.

The `sed` to remove build.rs might be slightly fragile.

If we later add a dependency that dynamically links a .so, we'll have to remember to copy that .so into the image. If we fail to do so, we'll get a somewhat mystifying "no such file or directory" when we try to run the container (which comes from trying to load the .so and not finding it).

There's a little dance where we create a dummy main.rs and then delete it. If we don't delete it, the build fails because having a `main.rs` conflicts with having a `bin/` in Cargo's (or Rust's?) file layout.

That said, I think all the above is fairly well handled in the Dockerfile and unlikely to spontaneously change. Just trying to think of all the possibilities.

PR #101, opened 2020-10-28
Add S3 input bucket support to workflow-manager

jsha:
Add the necessary terraform to grant permission on ingestion buckets.

Add a flag to control which service is in use (GCS vs S3).

Pull a token from the GKE metadata endpoint and use it to authenticate to the S3 API.

### DIFF
--- a/terraform/modules/kubernetes/kubernetes.tf
+++ b/terraform/modules/kubernetes/kubernetes.tf
@@ -12,10 +12,8 @@ variable "container_registry" {
 }
 
 variable "workflow_manager_image" {
-  type = string
-  # This should be "prio-data-share-processor" but we have not yet renamed the
-  # container image
-  default = "prio-workflow-manager"
+  type    = string
+  default = "letsencrypt"
 }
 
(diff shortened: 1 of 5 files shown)


terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -12,10 +12,8 @@ variable "container_registry" {
 }
 
 variable "workflow_manager_image" {
-  type = string
-  # This should be "prio-data-share-processor" but we have not yet renamed the
-  # container image
-  default = "prio-workflow-manager"
+  type    = string
+  default = "letsencrypt"
```
tgeoghegan, line 16: I get why you deleted the obsolete comment, but `letsencrypt` is wrong, and [`prio-workflow-manager`](https://hub.docker.com/repository/docker/letsencrypt/prio-workflow-manager) was correct.

tgeoghegan (approved):
LGTM except for the default value of `workflow_manager_image` in `kubernetes.tf`. Go ahead and merge after changing that.

PR #102, opened 2020-10-28
facilitator: discover parameters in peer manifests

tgeoghegan:
These commits teach the facilitator to handle multiple peer batch signing keys when evaluating trust over batches, and also to fetch JSON manifests from peers if adequate command line arguments are provided.

### DIFF
--- a/facilitator/src/bin/facilitator.rs
+++ b/facilitator/src/bin/facilitator.rs
@@ -6,13 +6,14 @@ use ring::signature::{
     EcdsaKeyPair, KeyPair, UnparsedPublicKey, ECDSA_P256_SHA256_ASN1,
     ECDSA_P256_SHA256_ASN1_SIGNING,
 };
-use std::str::FromStr;
+use std::{collections::HashMap, str::FromStr};
 use uuid::Uuid;
 
 use facilitator::{
     aggregation::BatchAggregator,
     config::StoragePath,
     intake::BatchIntaker,
+    manifest::{IngestionServerGlobalManifest, PortalServerGlobalManifest, SpecificManifest},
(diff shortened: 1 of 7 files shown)


facilitator/src/bin/facilitator.rs
```diff
@@ -717,22 +902,76 @@ fn batch_signing_key_from_arg(matches: &ArgMatches) -> Result<BatchSigningKey> {
     })
 }
 
-fn transport_for_output_path(
-    path: &str,
+fn ingestion_transport_and_keys_from_args(
+    matches: &ArgMatches,
+) -> Result<(
+    Box<dyn Transport>,
+    HashMap<String, UnparsedPublicKey<Vec<u8>>>,
```
jsha, line 909: I think we need a couple levels of typedefs to make this and other spots more readable:

```
type Pubkeys = HashMap<String, UnparsedPublicKey<Vec<u8>>>;

struct VerifiableTransport(Box<dyn Transport>, Pubkeys);
```

I think we can then even do:

```
impl VerifiableTransport {
  from_args(matches: &ArgMatches) -> Self {
     ...
  }
}
```

Which isn't necessarily any less typing but conveys the idea a little more clearly of "here's how you make an X from a Y."

jsha (approved):
Basically looks good; I recommended factoring out some types. Feel free to merge after that refactoring or merge as-is and consider refactoring later.

facilitator/src/bin/facilitator.rs
tgeoghegan: I thought about this, but worried it'd make the diff even bigger than it is. I'll give it a try though, this kind of tidying can be soothing.

PR #105, opened 2020-10-29
Store copy of validation batches in internal bucket

tgeoghegan:
This PR amends the Terraform to create an additional GCS bucket into which we shall write a copy of validation batches, to ensure we have our own validations available at aggregation time. It also extends the intake map step implementation in facilitator so that it can take two `SignableTransports`, which required teaching `BatchWriter` and `SidecarWriter` to deal with many potential writers under the covers.

### DIFF
--- a/facilitator/src/bin/facilitator.rs
+++ b/facilitator/src/bin/facilitator.rs
@@ -452,6 +452,7 @@ fn main() -> Result<(), anyhow::Error> {
                 .add_storage_arguments(Entity::Ingestor, InOut::Input)
                 .add_manifest_base_url_argument(Entity::Peer)
                 .add_storage_arguments(Entity::Peer, InOut::Output)
+                .add_storage_arguments(Entity::Own, InOut::Output),
         )
         .subcommand(
             SubCommand::with_name("aggregate")
@@ -521,12 +522,12 @@ fn main() -> Result<(), anyhow::Error> {
                 .add_storage_arguments(Entity::Ingestor, InOut::Input)
                 .add_batch_public_key_arguments(Entity::Ingestor)
                 .add_manifest_base_url_argument(Entity::Own)
-                .add_storage_arguments(Entity::Own,  InOut::Output)
+                .add_storage_arguments(Entity::Own, InOut::Input)
                 .add_manifest_base_url_argument(Entity::Peer)
-                .add_storage_arguments(Entity::Peer,  InOut::Input)
+                .add_storage_arguments(Entity::Peer, InOut::Input)
                 .add_batch_public_key_arguments(Entity::Peer)
                 .add_manifest_base_url_argument(Entity::Portal)
-                .add_storage_arguments(Entity::Portal,  InOut::Output)
+                .add_storage_arguments(Entity::Portal, InOut::Output)
                 .add_packet_decryption_key_argument()
                 .add_batch_signing_key_arguments()
                 .arg(Arg::with_name("is-first").long("is-first").help(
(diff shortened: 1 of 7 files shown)


terraform/modules/data_share_processor/data_share_processor.tf
```diff
@@ -228,6 +232,29 @@ POLICY
...
+  # be re-created by a subsequent deploy so no reason to keep them around.
+  force_destroy               = true
+  uniform_bucket_level_access = true
+}
+
+# Permit the workflow manager and facilitator service account to manage the
+# bucket
+resource "google_storage_bucket_iam_binding" "own_validation_bucket_admin" {
+  bucket = google_storage_bucket.own_validation_bucket.name
+  role   = "roles/storage.objectAdmin"
```
aaomidi, line 252: Out of curiosity does it need admin level permissions - or would objectCreator be enough?

aaomidi (approved):

terraform/modules/data_share_processor/data_share_processor.tf
tgeoghegan: Excellent question, I was a little surprised at this, too. Per [GCP docs](https://cloud.google.com/storage/docs/access-control/iam-roles), `storage.ObjectAdmin` gets you `storage.objects.*`. I thought that meant `storage.objects.create`, `storage.objects.get`, `storage.objects.delete` and `storage.objects.list`, all of which are needed to read, write and delete objects. Now that I look at it again, it _also_ must get you `storage.objects.getIamPolicy` and `storage.objects.setIamPolicy`, which we _don't_ want to grant to this service account. Annoyingly, I don't see a predefined role that includes `storage.objects.delete`, so I think I need to create a custom role here that include that permission.

tgeoghegan: turns out `storage.legacyBucketWriter` fits the bill, and is fine to use in spite of the ominous `legacy`.

facilitator/src/bin/facilitator.rs
```diff
@@ -605,24 +606,36 @@ fn main() -> Result<(), anyhow::Error> {
...
-                    base_url,
-                    sub_matches.value_of("instance-name").unwrap(),
-                )?
-                .validation_bucket()
-            } else {
-                Err(anyhow!("peer-output or peer-manifest-base-url required."))
-            }?;
+            // peer data share processor, which can either be fetched from the
+            // peer manifest or provided directly via command line argument. We
+            // check the manifest argument first because peer-output will always
```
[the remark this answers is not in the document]
tgeoghegan: I'm a little torn here. I think as a convenience to users of the tool, it's nice to be able to specify a manifest to get keys out of it, but override any buckets it advertises with an explicit `-output` argument. On the other hand, this kind of user convenience thing is tricky to get right, and we don't _need_ it, so we should probably punt on making `facilitator`'s UX particularly good, and focus on enabling it to work in our actual deployment, which would argue for removing the default value.

facilitator/src/lib.rs
```diff
@@ -72,30 +72,34 @@ impl Write for DigestWriter {
     }
 }
 
-/// SidecarWriter wraps an std::io::Write, but also writes any buffers passed to
-/// it into a second std::io::Write.
+/// SidecarWriter wraps a vector of std::io::Writes of one type and writes all
```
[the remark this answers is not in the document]
tgeoghegan: I wish, but doing that makes other methods harder to implement, because then you have to recurse through the stacked `SidecarWriters` to eventually get to the `DigestWriter` at the bottom so you can retrieve the digest over the content.

jsha (approved):

Issue #21, opened 2020-09-23
Code coverage in CI


tgeoghegan:
We can use [this GitHub action](https://github.com/actions-rs/grcov) to get code coverage reports from test runs in CI.

aaomidi:
Unfortunately https://github.com/actions-rs/grcov doesn't support non-top-level rust code :( https://github.com/actions-rs/grcov/issues/50

Not a big deal, we can just run the commands manually.

PR #108, opened 2020-10-29
Facilitator github workflow

aaomidi:
Fixes https://github.com/abetterinternet/prio-server/issues/21

I had to do the build/test coverage creation manually by hand since unfortunately the actions didn't support changing the working directory :(


.github/workflows/facilitator-test.yml
```diff
@@ -0,0 +1,46 @@
+name: facilitator-test
```
tgeoghegan, line 1: We already have build, test and clippy in the `build-rust` job in `.github/worksflows/ci-build.yml`. I think the thing to do is add the new coverage target there.

tgeoghegan (changes requested):
Please resolve this with the existing Rust checks in `.github/workflows/ci-build.yml`.

aaomidi:
Do you want me to make another job there or just integrate with `build-rust`? Because I need to use the nightly version of rust for this to work - are we fine changing the build to just use rust nightly?

tgeoghegan:
We should not use the nightly Rust toolchain for the regular CI build, because we want the build and test runs to reflect how we will build production binaries. Also it looks like using the nightly toolchain broke one of the steps in the existing job. I think making a separate job is the right thing, and please include a comment explaining why we need a separate toolchain.

tgeoghegan:
Looking good! Where do we actually get the coverage results though?

aaomidi:
The coverage results will be posted as a comment on the PR and available through the codecov website.https://codecov.io/gh/abetterinternet/prio-server/pull/108

jsha (approved):

tgeoghegan (approved):

PR #112, opened 2020-10-29
Tweak facilitator args help.

jsha:
Previously we had a lot of args that all had the same (long) help
explaining about GCS vs S3 storage or auth. It made it hard to read the
help because it wasn't clear that each stanza was identical. It also
obscured the most relevant information for each flag (the entity type,
i.e. peer, ingestor, self, or portal).

This moves that detailed help text out into a single staza at the top,
and shortens the help text for each flag.

I also renamed all the "ingestion-bucket" args to "ingestor-bucket",
because the bucket is owned by an ingestor. We should describe what we
do to that bucket as "intake," not "ingestion" to avoid confusion with
the "ingestor" term.

Renamed batch-intake to intake-batch so it's a verb like the "aggregate"
subcommand.


facilitator/src/bin/facilitator.rs
```diff
@@ -78,6 +78,23 @@ trait AppArgumentAdder {
     fn add_packet_decryption_key_argument(self: Self) -> Self;
 }
 
+const SHARED_HELP: &str =
+    "Service accounts: Any flag ending in -gcp-sa-email specifies a GCP service account \
+(identified by an email address). Requests to Google Storage (gs://) will be \
+authenticated as this account, by means of impersonation using GKE's Workload \
+Identity feature: https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity.\
```
tgeoghegan, line 85: This isn't quite accurate: when using `gs://` buckets, we use GKE Workload Identity no matter what, because that's how we map our Kubernetes service accounts to the per-data share processor GCP SA. The `*-gcp-sa-email` argument specifies a _further_ GCP SA account that we will impersonate once we have obtained the default account's token. Concretely, we use this so that our per-data share processor GCP SA can impersonate the singleton GCP SA that has been blessed by MITRE to write to the sum part buckets.

facilitator/src/bin/facilitator.rs
```diff
@@ -78,6 +78,23 @@ trait AppArgumentAdder {
...
+Such flags are ignored when reading or writing an S3 bucket (s3://). \
+\
+Flags ending in -s3-arn specify an AWS service account (identified by an `arn:` URI). \
+Requests to S3 will be authenticated by assuming this role. The credential to assume \
+the role is an OIDC auth token obtained by the GKE metadata service. If an -s3-arn flag \
```
tgeoghegan, line 90: nit: "obtained *from* the GKE metadata service"

tgeoghegan (approved):
Approved, with a nit or two

Issue #26, opened 2020-09-25
Break current aggregation reduce step into validation map step and final sum reduce


tgeoghegan:
The implementation of aggregation in #4 implements peer validation share validation and per-batch summing in the final reduce/aggregation step, but those steps could be implemented as a map step in parallel across the batches. We should revisit the implementation in `lib/aggregation.rs` and break the `aggregate_share` method into a separate step.

PR #113, opened 2020-10-30
Collapse AWS and GCP identity flags

jsha:
Previously, for each bucket-related argument, we had two flags: One
was a boolean controlling whether we consult $AWS_ROLE_ARN to
impersonate an IAM role, and the other was the name of a GCP service
account.

These flags both contained and identity, and were mutually exclusive, so
I've collapsed them into a single flag per bucket: `--foo-identity`.

I've also abstracted out the notion of an "Entity": Ingestor, Peer, Self
(or "Own"), and Portal. This makes the calls to the argument adding
methods simpler and more expressive. There's also a new enum for the
role of a bucket in the operation: input or output.

This also replaces reading AWS_ROLE_ARN from the environment
with getting it from the identity flag.


facilitator/src/bin/facilitator.rs
```diff
@@ -97,6 +87,57 @@ Keys: All keys are P-256. Public keys are base64-encoded DER SPKI. Private keys
...
+}
+
+/// One of the organizations participating in the Prio system.
+enum Entity {
+    Ingestor,
+    Peer,
+    Own,
+    // Hack: the aggregation bucket is part of the "Own" entity (i.e. whoever is running this).
+    // For the purpose of command line flags it acts as a separate entity.
+    Aggregation,
```
tgeoghegan, line 112: Given that we haven't yet broken the second map step out of the aggregation reduce step (#26), does it make sense to separate `Aggregation` from `Portal` yet?

facilitator/src/lib.rs
```diff
@@ -17,6 +17,10 @@ pub use workflow::{workflow_main, WorkflowArgs};
 
 pub const DATE_FORMAT: &str = "%Y/%m/%d/%H/%M";
 
+/// Identity represents a cloud identity: Either an AWS IAM ARN (i.e. "arn:...")
+/// or a GCP ServiceAccount (i.e. "foo@bar.com").
+pub type Identity = String;
```
tgeoghegan, line 22: The type alias is adequate for now, but it'd be neat if this was an enum akin to `enum StoragePath` in `config.rs`. Come to think of it that would be a more natural place for this defintion.

facilitator/src/transport/gcs.rs
```diff
@@ -200,14 +200,21 @@ pub struct GCSTransport {
 
 impl GCSTransport {
     /// Instantiate a new GCSTransport to read or write objects from or to the
-    /// provided path. If impersonate is None, GCSTransport authenticates to GCS
-    /// as the default service account. If impersonate contains a service
+    /// provided path. If identity is "", GCSTransport authenticates to GCS
+    /// as the default service account. If identity contains a service
```
tgeoghegan, line 204: `Option<String>` feels more rustacular than checking for the empty string

facilitator/src/transport/s3.rs
```diff
@@ -668,28 +667,33 @@ mod tests {
...
         });
+        let mut transport = S3Transport::new_with_client(s3_path.clone(), None, provider);
 
         let ret = transport.get(TEST_KEY);
         assert!(ret.is_err(), "unexpected return value {:?}", ret.err());
 
-        let mut transport = S3Transport::new_with_client(s3_path.clone(), false, |region, _| {
-            Ok(S3Client::new_with(
-                // Successful GetObject request
-                MockRequestDispatcher::with_status(200)
```
tgeoghegan, line 686: This reminds me: now that we know about [mockito](https://docs.rs/mockito/0.27.0/mockito/index.html), which has proper support for verifying that mocks were consumed than what I rigged up here with Rusoto's mocking framework, we should rewrite these tests. #114

tgeoghegan (approved):
LGTM!

PR #115, opened 2020-10-30
Add passthrough of args to workflow manager.

jsha:
Common args are stored in a ConfigMap, which is used by workflow-manager.

Remove the secret keys from the workflow container, and have the workflow container configure the facilitator container to receive them. This introduces two new flags to the workflow container to pass the names of those secrets.

### DIFF
--- a/terraform/modules/kubernetes/kubernetes.tf
+++ b/terraform/modules/kubernetes/kubernetes.tf
@@ -2,6 +2,18 @@ variable "data_share_processor_name" {
   type = string
 }
 
+variable "peer_share_processor_aws_account_id" {
+  type = string
+}
+
+variable "peer_share_processor_manifest_domain" {
+  type = string
+}
+
+variable "peer_validation_bucket_name" {
+  type = string
+}
+
 variable "environment" {
   type = string
 }
--- a/terraform/modules/kubernetes/kubernetes.tf
+++ b/terraform/modules/kubernetes/kubernetes.tf
@@ -167,6 +167,22 @@ resource "kubernetes_cron_job" "workflow_manager" {
               args = [
                 "--k8s-namespace", var.kubernetes_namespace,
                 "--input-bucket", var.ingestion_bucket,
+                "--bsk-secret-name", kubernetes_secret.batch_signing_key.metadata[0].name,
+                "--pdks-secret-name", var.packet_decryption_key_kubernetes_secret,
+                "passthrough",
+                # These args are passed to facilitator in addition to the per-batch
+                # args that workflow-manager builds to pass in.
+                "intake-batch",
+                "--aggregation-id", "fake-1",
+                "--instance-name", "foo",
+
+                "--ingestor-input", var.ingestion_bucket,
+                "--ingestor-identity", var.ingestion_bucket_role,
+                "--ingestor-manifest-base-url", "I dunno",
+
+                "--peer-output", var.peer_validation_bucket_name,
+                "--peer-identity", var.peer_share_processor_aws_account_id,
+                "--peer-manifest-base-url", "https://${var.peer_share_processor_manifest_domain}/",
               ]
               env {
                 name  = "AWS_ROLE_ARN"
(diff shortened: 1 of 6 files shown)


terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -2,6 +2,18 @@ variable "data_share_processor_name" {
...
+  type = string
+}
+
+variable "peer_share_processor_manifest_domain" {
+  type = string
+}
+
+variable "peer_validation_bucket_name" {
+  type = string
+}
```
tgeoghegan, line 15: Boy I wish I hadn't made these separate modules. Plumbing all these input variables is just a chore. Nothing to be done about it in this change, but I filed #117 to clean this up.

terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -167,6 +179,22 @@ resource "kubernetes_cron_job" "workflow_manager" {
               args = [
                 "--k8s-namespace", var.kubernetes_namespace,
                 "--input-bucket", var.ingestion_bucket,
+                "--bsk-secret-name", kubernetes_secret.batch_signing_key.metadata[0].name,
+                "--pdks-secret-name", var.packet_decryption_key_kubernetes_secret,
```
tgeoghegan, line 183: I think this will be fine for Narnia, but at some point, we have to work out how to provide _all_ the packet decryption keys to facilitator so it can try them all.

terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -167,6 +179,22 @@ resource "kubernetes_cron_job" "workflow_manager" {
               args = [
                 "--k8s-namespace", var.kubernetes_namespace,
                 "--input-bucket", var.ingestion_bucket,
```
tgeoghegan, line 181: I'm guessing `input-bucket` is used by `workflow-manager` but `ingestor-input` is what goes to `facilitator`? Seems fine, especially since this may shortly get replaced by a configmap.

terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -167,6 +179,22 @@ resource "kubernetes_cron_job" "workflow_manager" {
...
+                "passthrough",
+                # These args are passed to facilitator in addition to the per-batch
+                # args that workflow-manager builds to pass in.
+                "intake-batch",
+                "--aggregation-id", "fake-1",
+                "--instance-name", "foo",
+
+                "--ingestor-input", var.ingestion_bucket,
+                "--ingestor-identity", var.ingestion_bucket_role,
+                "--ingestor-manifest-base-url", "I dunno",
```
tgeoghegan, line 193: This has to be plumbed in from main.tf. I am just tinkering with a branch where I wire all this into a config map.

workflow-manager/main.go
```diff
@@ -234,20 +241,32 @@ func launch(ctx context.Context, readyBatches []string) error {
...
 
 	pathComponents := strings.Split(batchName, "/")
 	batchID := pathComponents[len(pathComponents)-1]
 	batchDate := strings.Join(pathComponents[0:len(pathComponents)-1], "/")
+
+	var args []string
+	args = append(args, passthroughArgs...)
+	args = append(args,
+		"--batch-id", batchID,
+		"--date", batchDate,
```
tgeoghegan, line 267: I think you need to tease an `aggregation-id` value out of the `batchName`.

terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -167,6 +179,22 @@ resource "kubernetes_cron_job" "workflow_manager" {
...
+
+                "--peer-output", var.peer_validation_bucket_name,
+                "--peer-identity", var.peer_share_processor_aws_account_id,
+                "--peer-manifest-base-url", "https://${var.peer_share_processor_manifest_domain}/",
```
tgeoghegan, line 197: Now that #105 has landed, you also need `--own-output`, `--own-identity`

tgeoghegan (dismissed):
If we're going to revisit all this with a ConfigMap later today, then the only thing I'm concerned about in this PR is the handling of the `aggregation-id` parameter. But I don't think you'll be able to usefully invoke the facilitator without providing `own-output` and the ingestor manifest base URL.

terraform/modules/kubernetes/kubernetes.tf
jsha: I figure we can just cram them all into the one secret, comma-separated like the flag wants.

jsha (commented):
Approved your pushed changes @tgeoghegan. I can't actually hit "approve" because I'm the author.

yuriks (approved):
As per slack request:
> Could one of you hit approve on https://github.com/abetterinternet/prio-server/pull/115? Tim pushed changes on top of mine; I approve of his changes, and implicitly he approves of mine, but I can't "Approve" the PR because I'm the author. Also additional review comments very welcome if you'd like to give them!

I skimmed the changes but didn't look very closely.

aaomidi (approved):

PR #119, opened 2020-10-30
Manifest updater

aaomidi:
Plan with this PR:

I'm going to need these manifest structures in another piece of code (operator/job) soon, so I'm removing them out of the deploy-tool, and making them their own module. 

I have another branch ready that builds on top of this that removes the manifest declarations on the deploy-tool and uses this gomodule for them instead: https://github.com/abetterinternet/prio-server/compare/amir/manifest-updater...amir/manifest_updater/deploy-tool


tgeoghegan:
I like using the real GCP SDK for GCS uploads, but couldn't this be a module built into the existing deploy-tool? I'm not sure why it's a separate binary. Also it seems like this duplicates a number of struct definitions already present in `deploy-tool`. Finally, I don't think you're going to need the global manifest definitions: we construct and post our global manifest from Terraform.

aaomidi:
Ah I should've probably explained this better, but I'm not planning on using this as its own standalone binary - rather just reference it with the go modules style references from deploy-tool.

Since I'm working on an operator based deploy tool, I was going to need a manifest updater in another codebase as well, and thought it would be better if the deploy tool and the new operator deploy tool share the same manifest updater code.

aaomidi:
@tgeoghegan you might be interested in seeing what this could end up looking like: https://github.com/abetterinternet/prio-server/compare/amir/manifest-updater...amir/manifest_updater/deploy-tool#diff-82dda3d59181ac4501f0e3f170e2c0b12ff02ee4e76e65020c30acd453278d20

PR #120, opened 2020-10-30
Add workflow-manager publish to GitHub Actions.

jsha:
Note: I haven't tested this yet, but adapted from what was already here. Need to figure out how to test these; advice welcome!


tgeoghegan:
I also don't know how to test this short of a test, so let's just merge it and cut a release. Worst case it doesn't work and we do another PR.

tgeoghegan (approved):

PR #128, opened 2020-10-31
Bundle ca-certificates in facilitator.

jsha:
The change to ship facilitator as a bare binary broke interactions with
S3 (which uses rusoto and rustls-native-certs). This unbreaks.


.github/workflows/push-docker-image.yml
```diff
@@ -25,10 +25,14 @@ jobs:
         file: ./facilitator/Dockerfile
```
tgeoghegan, line 25: This `build` step should be removed since you added the one that invokes `build.sh`

facilitator/build.sh
```diff
@@ -0,0 +1,14 @@
...
+cd -- "$(dirname -- "$0")"/..
+
+# The container needs a copy of trusted roots. Try to copy them from the OS.
+# See
+# https://medium.com/@kelseyhightower/optimizing-docker-images-for-static-binaries-b5696e26eb07
+# and
+# https://github.com/golang/go/blob/master/src/crypto/x509/root_linux.go#L7
+cp /etc/ssl/certs/ca-certificates.crt . \
+ || cp /etc/pki/tls/certs/ca-bundle.crt ca-certificates.crt \
+ || cp /etc/pki/tls/certs/ca-bundle.crt ca-certificates.crt
```
tgeoghegan, line 12: Why have the same `cp` command twice?

tgeoghegan (changes requested):

jsha:
I wound up re-doing this and reverting the GH Actions changes, because I realized I can just copy the cert bundle from the builder image. Much simpler, hardly any changes.

tgeoghegan (approved):

PR #129, opened 2020-10-31
Various fixes to get intake-batch going

tgeoghegan:
deploy-tool:
libprio-rs expects to receive packet decryption keys as the X9.62
uncompressed public key, concatenated with the secret scalar, and not
the PKCS#8 documents that we were plugging into Kubernetes secrets.
deploy-tool now has logic to serialize the ECDSA keys it generates into
the desired format.

facilitator:
 - generate-ingestion-sample now consults environment variables for the
   ECIES encryption keys to use when generating data.
 - argument handling was incorrectly looking for "batch-intake" instead
   of "intake-batch"
 - manifest::fetch_manifest now makes sure to construct https:// URLs
   when fetching manifests.

terraform:
 - Use the same key name within all Kubernetes secrets for simplicity
 - Fix parameters to sample_maker cronjob

workflow-manager:
 - parse aggregation ID from batch filename
 - fix arguments to `facilitator intake-batch`
 - fix `SecretKeyRefs` set on intake-batch jobs to correctly reference
   key inside secrets

### DIFF
--- a/deploy-tool/main.go
+++ b/deploy-tool/main.go
@@ -31,7 +31,7 @@ import (
 // BatchSigningPublicKey represents a public key used for batch signing.
 type BatchSigningPublicKey struct {
 	// PublicKey is the PEM armored base64 encoding of the ASN.1 encoding of the
-	// PKIX SubjectPublicKeyInfo structure. It must be an ECDSA P256 key.
+	// PKIX SubjectPublicKeyInfo structure. It must be a P-256 key.
 	PublicKey string `json:"public-key"`
 	// Expiration is the ISO 8601 encoded UTC date at which this key expires.
 	Expiration string `json:"expiration"`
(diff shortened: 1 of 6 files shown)


facilitator/src/manifest.rs
```diff
@@ -61,10 +61,12 @@ pub struct SpecificManifest {
 
 /// Obtains a manifest file from the provided URL
 fn fetch_manifest(manifest_url: &str) -> Result<Response> {
-    if !manifest_url.starts_with("https://") {
-        return Err(anyhow!("Manifest must be fetched over HTTPS"));
-    }
-    let response = ureq::get(manifest_url)
+    let url_with_scheme = if !manifest_url.starts_with("https://") {
```
jsha, line 64: Style-wise I tend to prefer to accept just one type of input - i.e. require the `https://` prefix - and fix whatever is providing the wrong sort of input. Is there a particular issue this is solving?

deploy-tool/main.go
```diff
@@ -85,17 +85,34 @@ type TerraformOutput struct {
...
+
+// marshalX962UncompresedPrivateKey encodes an ECDSA P256 private key into the
+// format expected by libprio-rs encrypt::PrivateKey, which is the X9.62
+// uncompressed public key concatenated with the secret scalar.
+func marshalX962UncompressedPrivateKey(ecdsaKey *ecdsa.PrivateKey) ([]byte, error) {
+	marshaledPublicKey := elliptic.Marshal(elliptic.P256(), ecdsaKey.PublicKey.X, ecdsaKey.PublicKey.Y)
+	return append(marshaledPublicKey, ecdsaKey.D.Bytes()...), nil
+}
+
+// marshakPKCS8PrivateKey encodes an ECDSA P256 private key into a PKCS#8
```
jsha, line 98:
```suggestion
// marshalPKCS8PrivateKey encodes an ECDSA P256 private key into a PKCS#8
```

deploy-tool/main.go
```diff
@@ -85,17 +85,34 @@ type TerraformOutput struct {
...
+// document. This function adapts x509.MarshalPKCS8PrivateKey to the
```
jsha, line 99: I don't think an adapter is needed; you can just pass `x509.MarshalPKCS8PrivateKey` directly.

deploy-tool/main.go
```diff
@@ -85,17 +85,34 @@ type TerraformOutput struct {
 	} `json:"specific_manifests"`
 }
 
+type privateKeyMarshaler func(*ecdsa.PrivateKey) ([]byte, error)
+
+// marshalX962UncompresedPrivateKey encodes an ECDSA P256 private key into the
```
jsha, line 90:
```suggestion
// marshalX962UncompresedPrivateKey encodes an P-256 private key into the
```

I think it's more correct to say just P-256, since we're not using these for ECDSA. And I believe P-256 is the official name, as opposed to P256.

jsha (changes requested):
Looks nice! Some minor changes.

facilitator/src/manifest.rs
tgeoghegan: No, ultimately these values come from the `ingestors`, `peer_share_processor_manifest_base_url` and `portal_server_manifest_base_url` variables in `<env>.tfvars`, and we can make sure that they're `https://` there. What I want is to make sure that we're fetching these manifests over HTTPS.

deploy-tool/main.go
tgeoghegan: I cannot:

jsha: Ohhh I see, it takes a blank interface! That makes sense. I'm fine with this as it stands.

You could also change the `privateKeyMarshaler` and `marshalX962UncompresedPrivateKey` types to take a blank interface, which would save you having to make a wrapper. But I don't think it's necessary.

deploy-tool/main.go
jsha: Edited comment to catch another typo.

deploy-tool/main.go
tgeoghegan: ...and if I change `privateKeyMarshaler` to take `interface{}` to match `x509.MarshalPKCS8PrivateKey`, then I would have to do a type check and cast in `marshalX962UncompressedPrivateKey` which I find uglier than providing the adapter for `x509.MarshalPKCS8PrivateKey`.

deploy-tool/main.go
tgeoghegan: Good point about us not using these keys for DSA. I updated a number of comments and variables in this file accordingly.

jsha (approved):

PR #136, opened 2020-11-01
Start aggregation jobs in workflow-manager

tgeoghegan:
With these changes, workflow-manager actually spawns Kubernetes jobs to run `facilitator aggregate` with suitable parameters. It doesn't yet implement the aggregation window, and I'm running into an issue with signature verification on peer validation batch signatures.

### DIFF
--- a/terraform/modules/kubernetes/kubernetes.tf
+++ b/terraform/modules/kubernetes/kubernetes.tf
@@ -268,13 +268,14 @@ resource "kubernetes_cron_job" "workflow_manager" {
               name  = "workflow-manager"
               image = "${var.container_registry}/${var.workflow_manager_image}:${var.workflow_manager_version}"
               args = [
-                "--is-first", var.is_first ? "true" : "false",
+                "--is-first=${var.is_first ? "true" : "false"}",
                 "--k8s-namespace", var.kubernetes_namespace,
                 "--k8s-service-account", kubernetes_service_account.workflow_manager.metadata[0].name,
                 "--ingestor-input", "s3://${var.ingestion_bucket}",
                 "--ingestor-identity", var.ingestion_bucket_role,
-                "--own-validation-bucket", "gs://${var.own_validation_bucket}",
-                "--peer-validation-bucket", "s3://${var.peer_validation_bucket}",
+                "--own-validation-input", "gs://${var.own_validation_bucket}",
+                "--peer-validation-input", "s3://${var.peer_validation_bucket}",
+                "--peer-validation-identity", var.ingestion_bucket_role,
                 "--bsk-secret-name", kubernetes_secret.batch_signing_key.metadata[0].name,
                 "--pdks-secret-name", var.packet_decryption_key_kubernetes_secret,
                 "--intake-batch-config-map", kubernetes_config_map.intake_batch_job_config_map.metadata[0].name,
--- a/terraform/modules/kubernetes/kubernetes.tf
+++ b/terraform/modules/kubernetes/kubernetes.tf
@@ -268,10 +268,13 @@ resource "kubernetes_cron_job" "workflow_manager" {
               name  = "workflow-manager"
               image = "${var.container_registry}/${var.workflow_manager_image}:${var.workflow_manager_version}"
               args = [
+                "--is-first", var.is_pha ? "true" : "false",
                 "--k8s-namespace", var.kubernetes_namespace,
                 "--k8s-service-account", kubernetes_service_account.workflow_manager.metadata[0].name,
                 "--ingestor-input", "s3://${var.ingestion_bucket}",
                 "--ingestor-identity", var.ingestion_bucket_role,
+                "--own-validation-bucket", "gs://${var.own_validation_bucket}",
+                "--peer-validation-bucket", "s3://${var.peer_validation_bucket}",
                 "--bsk-secret-name", kubernetes_secret.batch_signing_key.metadata[0].name,
                 "--pdks-secret-name", var.packet_decryption_key_kubernetes_secret,
                 "--intake-batch-config-map", kubernetes_config_map.intake_batch_job_config_map.metadata[0].name,
--- a/terraform/modules/kubernetes/kubernetes.tf
+++ b/terraform/modules/kubernetes/kubernetes.tf
@@ -268,7 +268,7 @@ resource "kubernetes_cron_job" "workflow_manager" {
               name  = "workflow-manager"
               image = "${var.container_registry}/${var.workflow_manager_image}:${var.workflow_manager_version}"
               args = [
-                "--is-first", var.is_pha ? "true" : "false",
+                "--is-first", var.is_first ? "true" : "false",
                 "--k8s-namespace", var.kubernetes_namespace,
                 "--k8s-service-account", kubernetes_service_account.workflow_manager.metadata[0].name,
                 "--ingestor-input", "s3://${var.ingestion_bucket}",
--- a/workflow-manager/main.go
+++ b/workflow-manager/main.go
@@ -495,11 +495,16 @@ func getReadyValidationBatchesS3(ctx context.Context, validationsBucket string, 
 }
 
 func launchAggregationJob(ctx context.Context, readyBatches []string) error {
+	if len(readyBatches) == 0 {
+		log.Printf("no batches to aggregate")
+		return nil
+	}
+
 	config, err := rest.InClusterConfig()
 	if err != nil {
 		return fmt.Errorf("cluster config: %w", err)
 	}
-	_, err = kubernetes.NewForConfig(config)
+	clientset, err := kubernetes.NewForConfig(config)
 	if err != nil {
 		return fmt.Errorf("clientset: %w", err)
 	}
@@ -515,7 +520,7 @@ func launchAggregationJob(ctx context.Context, readyBatches []string) error {
 			return err
 		}
 		batchIDs = append(batchIDs, batchPath.batchID)
-		batchTimes = append(batchIDs, batchPath.batchDateString())
+		batchTimes = append(batchTimes, batchPath.batchDateString())
 
 		// All batches should have the same aggregation ID?
 		if aggregationID != "" && aggregationID != batchPath.aggregationID {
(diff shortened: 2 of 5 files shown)


workflow-manager/main.go
```diff
@@ -53,16 +53,106 @@ type batch struct {
...
+	}
+	if !strings.HasPrefix(val, "s3://") && !strings.HasPrefix(val, "gs://") {
+		log.Fatalf("invalid %s: %q", arg, val)
+	}
+
+	return val[0:2], val[4:]
+}
+
 var k8sNS = flag.String("k8s-namespace", "", "Kubernetes namespace")
+var isFirst = flag.String("is-first", "false", "Whether this set of servers is \"first\", aka PHA servers")
```
jsha, line 155: I'd recommend a bool flag instead. Then you can get rid of parseIsFirst. https://golang.org/pkg/flag/#Bool

workflow-manager/main.go
```diff
@@ -53,16 +53,106 @@ type batch struct {
...
+			return true
+		}
+	}
+
+	return false
+}
+
+func checkStoragePath(arg string, val string) (string, string) {
+	if val == "" {
+		log.Fatalf("%s is requored", arg)
```
jsha, line 145:
```suggestion
		log.Fatalf("%s is required", arg)
```

workflow-manager/main.go
```diff
@@ -53,16 +53,106 @@ type batch struct {
...
+	for _, v := range s {
+		if v == str {
+			return true
+		}
+	}
+
+	return false
+}
+
+func checkStoragePath(arg string, val string) (string, string) {
```
jsha, line 143: Please add a doc comment.

workflow-manager/main.go
```diff
@@ -71,18 +161,16 @@ var intakeConfigMap = flag.String("intake-batch-config-map", "", "Name of config
 var aggregateConfigMap = flag.String("aggregate-config-map", "", "Name of config map for aggregate jobs")
 var ingestorInput = flag.String("ingestor-input", "", "Bucket for input from ingestor (s3:// or gs://) (Required)")
 var ingestorIdentity = flag.String("ingestor-identity", "", "Identity to use with ingestor bucket (Required for S3)")
+var ownValidationInput = flag.String("own-validation-input", "", "Bucket for input from self (s3:// or gs://) (required)")
```
jsha, line 164:
```suggestion
var ownValidationInput = flag.String("own-validation-input", "", "Bucket for input of validation batches from self (s3:// or gs://) (required)")
```

workflow-manager/main.go
```diff
@@ -93,25 +181,66 @@ func main() {
 		log.Fatal(err)
 	}
 
-	var readyBatches []string
-	switch service {
+	var readyIngestionBatches []string
```
jsha, line 184: Let's call these "intake" batches, to avoid the confusion I mentioned on Friday about how we don't actually do any ingestion; we do intake of batches produced by the ingestors.

workflow-manager/main.go
```diff
@@ -93,25 +181,66 @@ func main() {
...
+	switch ingestorService {
+	case "s3":
+		readyIngestionBatches, err = getReadyIngestionBatchesS3(context.Background(), ingestorInputBucket, *ingestorIdentity)
```
jsha, line 187: These names are starting to get pretty long. I think we can drop `ready` without loss of meaning, since the fact that we don't care about non-ready batches is an implementation detail of `getReadyIngestionBatchesS3` (or, as I propose renaming: `getIntakeBatchesS3`).

workflow-manager/main.go
```diff
@@ -93,25 +181,66 @@ func main() {
...
+
+	if err := launchIntake(context.Background(), readyIngestionBatches, ageLimit); err != nil {
+		log.Fatal(err)
+	}
+
+	var readyOwnValidationBatches []string
+	switch ownValidationService {
 	case "s3":
-		readyBatches, err = getReadyBatchesS3(context.Background(), inputBucket, *ingestorIdentity)
+		readyOwnValidationBatches, err = getReadyValidationBatchesS3(context.Background(),
```
jsha, line 207: This clause is missing an `err` check.

workflow-manager/main.go
```diff
@@ -93,25 +181,66 @@ func main() {
...
+	}
+
+	log.Printf("found ready own validations: %v", readyOwnValidationBatches)
+
+	var readyPeerValidationBatches []string
+	switch peerValidationService {
+	case "s3":
+		// These are the peer's validation, and if we were first, then they were
+		// not, so negate parseIsFirst's result
+		readyPeerValidationBatches, err = getReadyValidationBatchesS3(context.Background(),
```
jsha, line 229: This clause is missing an `err` check.

workflow-manager/main.go
```diff
@@ -93,25 +181,66 @@ func main() {
...
+		readyOwnValidationBatches, err = getReadyValidationBatchesGS(context.Background(),
+			ownValidationInputBucket, parseIsFirst(*isFirst), readyIngestionBatches)
 		if err != nil {
 			log.Fatal(err)
 		}
+	default:
+		log.Fatalf("unknown service %s", ownValidationService)
+	}
+
+	log.Printf("found ready own validations: %v", readyOwnValidationBatches)
```
jsha, line 222: This is going to be quite noisy and hard to read, with everything on one line. Let's just print the count of batches that we're going to work on. Each batch will print something as it's kicked off.

workflow-manager/main.go
```diff
@@ -93,25 +181,66 @@ func main() {
...
 		if err != nil {
 			log.Fatal(err)
 		}
 	default:
-		log.Fatalf("unknown service %s", service)
+		log.Fatalf("unknown service %s", ownValidationService)
 	}
 
-	if err := launch(context.Background(), readyBatches, ageLimit); err != nil {
+	log.Printf("found ready peer validations: %v", readyPeerValidationBatches)
```
jsha, line 244: This is going to be noisy and hard to read; as above, let's print the number.

workflow-manager/main.go
```diff
@@ -268,50 +666,41 @@ func launch(ctx context.Context, readyBatches []string, ageLimit time.Duration)
...
+		return err
 	}
-	var dateComponents []int
-	for _, c := range batchDate {
-		parsed, err := strconv.ParseInt(c, 10, 64)
-		if err != nil {
-			return fmt.Errorf("parsing date component %q in %q: %w", c, batchName, err)
-		}
-		dateComponents = append(dateComponents, int(parsed))
+	batchDate, err := batchPath.batchDateTime()
```
jsha, line 686: Go style would be to omit the redundant `batch` in the naming of this function. And `DateTime` is a little redundant since we're just returning a `time`. So I'd call this method `.time()`.

workflow-manager/main.go
```diff
@@ -268,50 +666,41 @@ func launch(ctx context.Context, readyBatches []string, ageLimit time.Duration)
...
+	if err != nil {
+		return err
 	}
 
-	batchTime := time.Date(dateComponents[0], time.Month(dateComponents[1]), dateComponents[2], dateComponents[3], dateComponents[4], 0, 0, time.UTC)
-	age := time.Now().Sub(batchTime)
+	age := time.Now().Sub(batchDate)
 	if age > ageLimit {
 		log.Printf("skipping batch %q because it is too old (%s)", batchName, age)
+		return nil
```
jsha, line 694: Nice catch, thanks. :-)

terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -268,10 +268,13 @@ resource "kubernetes_cron_job" "workflow_manager" {
               name  = "workflow-manager"
               image = "${var.container_registry}/${var.workflow_manager_image}:${var.workflow_manager_version}"
               args = [
+                "--is-first", var.is_first ? "true" : "false",
                 "--k8s-namespace", var.kubernetes_namespace,
                 "--k8s-service-account", kubernetes_service_account.workflow_manager.metadata[0].name,
                 "--ingestor-input", "s3://${var.ingestion_bucket}",
                 "--ingestor-identity", var.ingestion_bucket_role,
+                "--own-validation-bucket", "gs://${var.own_validation_bucket}",
+                "--peer-validation-bucket", "s3://${var.peer_validation_bucket}",
```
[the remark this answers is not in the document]
tgeoghegan: Not for the `gs://` bucket, as it's in the same cloud as us and we set it up so the workflow manager's GCP service account can use it, so nothing special is required. We could add a `peer-validation-identity` though it'd have the same value as `ingestor-identity`.

workflow-manager/main.go
```diff
@@ -254,7 +389,270 @@ func getReadyBatchesGS(ctx context.Context, inputBucket string) ([]string, error
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	createdJob, err := clientset.BatchV1().Jobs(*k8sNS).Create(ctx, job, metav1.CreateOptions{})
+	if err != nil {
+		// TODO: Checking an error for a substring is pretty clumsy. Figure out if k8s client
```
jsha, line 641: I fixed this in my recent PR; you can use `if errors.IsAlreadyExists(err)`.

workflow-manager/main.go
```diff
@@ -254,7 +389,270 @@ func getReadyBatchesGS(ctx context.Context, inputBucket string) ([]string, error
...
+										SecretKeyRef: &corev1.SecretKeySelector{
+											LocalObjectReference: corev1.LocalObjectReference{
+												Name: *bskSecretName,
+											},
+											Key: "secret_key",
+										},
+									},
+								},
+								{
+									Name: "PACKET_DECRYPTION_KEYS",
```
jsha, line 622: Does the aggregation step use the packet decryption keys? I see that it's an arg on the `aggregate` subcommand, but I don't see it actually used.

workflow-manager/main.go
```diff
@@ -254,7 +389,270 @@ func getReadyBatchesGS(ctx context.Context, inputBucket string) ([]string, error
...
+		}, Spec: batchv1.JobSpec{
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					ServiceAccountName: *k8sServiceAccount,
+					RestartPolicy:      "Never",
+					Containers: []corev1.Container{
+						{
+							Args:            args,
+							Name:            "facile-container",
+							Image:           "us.gcr.io/prio-bringup-290620/prio-facilitator:latest",
```
jsha, line 598: Just noting that this is hardcoded to a different image name than the one for intake jobs. I'm fine landing as-is; we need to land a PR to parameterize these, and that doesn't need to block this change.

workflow-manager/main.go
```diff
@@ -254,7 +389,270 @@ func getReadyBatchesGS(ctx context.Context, inputBucket string) ([]string, error
...
+
+	jobName := strings.ReplaceAll(fmt.Sprintf("a-%s", earliestBatch.batchDateString()), "/", "-")
+
+	log.Printf("starting aggregation job with args %s", args)
+
+	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      jobName,
+			Namespace: *k8sNS,
+		}, Spec: batchv1.JobSpec{
```
jsha, line 589: Should be a newline after this comma.

workflow-manager/main.go
```diff
@@ -254,7 +389,270 @@ func getReadyBatchesGS(ctx context.Context, inputBucket string) ([]string, error
...
+
+	for _, batchName := range readyBatches {
+		batchPath, err := newBatchPath(batchName)
+		if err != nil {
+			return err
+		}
+		batchIDs = append(batchIDs, batchPath.batchID)
+		batchTimes = append(batchTimes, batchPath.batchDateString())
+
+		// All batches should have the same aggregation ID?
```
jsha, line 531: This is an error that could be triggered by a mistake our peer makes, and it would bring processing to a halt until they resolved it. We should be robust to the error (e.g. log an error and ignore the mismatch), or enforce this property internally by doing a filter on the readyBatches before passing them to this function.

workflow-manager/main.go
```diff
@@ -254,7 +389,270 @@ func getReadyBatchesGS(ctx context.Context, inputBucket string) ([]string, error
...
+		if aggregationID != "" && aggregationID != batchPath.aggregationID {
+			return fmt.Errorf("found batch with aggregation ID %s, wanted %s", batchPath.aggregationID, aggregationID)
+		}
+		aggregationID = batchPath.aggregationID
+
+		// Find earliest and latest batch in the aggregation
+		currentBatchTime, err := batchPath.batchDateTime()
```
jsha, line 538: Let's calculate and store the dateTime when a new batchPath is created. That way we can make creation of the batchPath fallible (since it only happens once), and make checking the time infallible (that happens a bunch of times, so it simplifies the code).

workflow-manager/main.go
```diff
@@ -254,7 +389,270 @@ func getReadyBatchesGS(ctx context.Context, inputBucket string) ([]string, error
...
+		if err != nil {
+			return err
+		}
+		if earliestBatch == nil {
```
jsha, line 542: We can instantiate earliestBatch and latestBatch to `readyBatches[0]` before the `for` loop. We've already checked at the top of the function that `len(readyBatches[0])` is nonzero.

terraform/modules/kubernetes/kubernetes.tf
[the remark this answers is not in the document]
jsha: I'd like to have a `--peer-validation-identity` even though it's the same as `ingestor-identity`. It documents to someone reading the terraform what identity is used to access that bucket, since we've established the pattern that `--foo-{input,output}` flags come in pairs with `--foo-identity` flags.

workflow-manager/main.go
tgeoghegan: I renamed a bunch of members and functions to remove redundant `batch` occurences.

workflow-manager/main.go
tgeoghegan: Yes! We need to decrypt the shares in the original ingestion batch.

workflow-manager/main.go
tgeoghegan: Good catch. That's the image in my container registry I've been using. Yes, let's fix that when we parameterize the container image.

workflow-manager/main.go
tgeoghegan: I'm honestly not sure if the aggregation ID ever changes, but wanted to get any aggregation working before circling back to that question, or implementing logic to enforce it.

jsha (approved):

jsha (approved):

Issue #140, opened 2020-11-02
Refine and narrow the role policy used for bucket writing


tgeoghegan:
`data_share_processor.tf` creates a role that is used to access S3 buckets. At the moment, we use this policy:

```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": [
        "s3:*"
      ],
      "Effect": "Allow",
      "Resource": "*"
    }
  ]
}
```

Note that this is not the same thing as the role's role assumption policy, nor is it the same as the bucket's policy. This is the policy on the role. Anyway, granting arbitrary s3 actions on arbitrary resources got our workflow working, and as I understand it, the restrictive _bucket_ policy applies regardless, but we should still see if we can define this more narrowly, as I worry that this would grant this role access to any bucket that didn't define a policy that denies access.

tgeoghegan:
Some relevant AWS documentation: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_evaluation-logic.html

Sadly this one does not just go away once we do #68, because we still will need to write content into NIH owned S3 buckets.

PR #141, opened 2020-11-02
terraform: Add IAM policy for bucket role

tgeoghegan:
For each data share processor, we create an AWS IAM role which is used
by that DSP to access various S3 buckets. We also create a bucket to
which peer DSPs will write their validation batches. We aim to grant the
entire peer AWS account access to that bucket, which we achieve with
Principal { AWS { aws-account-id } } in the _bucket_ policy. It turns
out we also need a _role_ policy permitting use of the S3 API.

Issue #140 tracks restricting this policy further.

### DIFF
--- a/terraform/modules/data_share_processor/data_share_processor.tf
+++ b/terraform/modules/data_share_processor/data_share_processor.tf
@@ -152,6 +152,33 @@ ROLE
   }
 }
 
+# The bucket_role defined above is created in our AWS account and used to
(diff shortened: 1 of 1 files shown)


aaomidi (approved):

jsha (approved):
We should definitely pursue #140 quickly. This definitely strikes me as the sort of "get it working" change that winds up making it into production and then winding up with some wide-open permissions unintentionally.

PR #142, opened 2020-11-02
Apply Cache-Control: no-cache header on manifests

jsha:
It looks like the load balancer observes the cache-control header on the
origin, and Google Storage sets cache-control on a per-object basis,
with a default of 1 hour for public objects.

This prevents getting confusing results during debugging if you've edited
a manifest.

### DIFF
--- a/deploy-tool/main.go
+++ b/deploy-tool/main.go
@@ -278,8 +278,10 @@ func main() {
 		destination := fmt.Sprintf("gs://%s/%s-manifest.json",
 			terraformOutput.ManifestBucket.Value, dataShareProcessorName)
 		log.Printf("uploading specific manifest %s", destination)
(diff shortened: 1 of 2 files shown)


tgeoghegan (changes requested):
This achieves what it says on the tin for global manifests, but we should do the same thing for specific manifests, which get uploaded from deploy-tool. It seems it's easy enough to do that with an extra `-h` argument to `gsutil`, alongside where we currently set `Content-Type:application/json`.

https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata

jsha:
Good call, thanks! Updated to include the deploy-tool change.

tgeoghegan (approved):

PR #144, opened 2020-11-02
README: add guide to service account and roles

tgeoghegan:
Extends the main README with some tips on how to use `gcloud` and `aws`
to simulate authentication flows that we use in the facilitator and
workflow-manager, as a debugging aid.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,3 +17,55 @@ We use a GitHub Action to build Docker images and push them to [DockerHub](https
 - Bump the version number in `facilitator/Cargo.toml` and merge that change to `main`.
 - Tag that commit on main, either in `git` or in [GitHub's releases UI](https://github.com/abetterinternet/prio-server/releases/new).
 - Publish a release in [GitHub's releases UI](https://github.com/abetterinternet/prio-server/releases/new).
+
+## Impersonating cloud service accounts
+
+Because of this project's multi-cloud nature, we wind up using some unusual authentication flows to share data across different accounts, across different cloud providers. Debugging these authentication flows can be tricky so here is a short guide to simulating those flows from your workstation.
+
+### Service account impersonation vs. role assumption
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,3 +17,47 @@ We use a GitHub Action to build Docker images and push them to [DockerHub](https
...
+## Impersonating cloud service accounts
+
+Because of this project's multi-cloud nature, we wind up using some unusual authentication flows to share data across different accounts, across different cloud providers. Debugging these authentication flows can be tricky so here is a short guide to simulating those flows from your workstation.
+
+### Assuming an AWS IAM role from a Google Cloud Platform service account
+
+In this flow, a GCP service account assumes an AWS IAM role using OIDC identity federation. You will need:
+
+- The email of the GCP service account
+- The OIDC audience to scope the OIDC token to (its value should match the role assumption policy on the AWS IAM role you are assuming)
```
jsha, line 30: How does one find the role assumption policy? Also, I've been curious - we construct these as `sts.amazonaws.com/<accountid>`. Is that format mandated by some third party, or do we just put what we want in there? Could it be `foo.bar.example/<accountid>`?

README.md
```diff
@@ -17,3 +17,47 @@ We use a GitHub Action to build Docker images and push them to [DockerHub](https
...
+- The IAM ARN of the AWS IAM role you are assuming
+
+First, you need an _identity token_ from the GCP IAM service scoped to the correct _audience_.
```
jsha, line 33:
```suggestion
First, you need an _identity token_ from the GCP IAM service scoped to the correct AWS _audience_.
```

jsha (changes requested):
Looks great, thanks for writing this up.

README.md
yuriks: Yeah, we can put whatever we want in there. I suggested that URL because it seemed like a reasonable way to scope the token. (It maps to AWS roles, which are namespaced by the account id, so it seems reasonably specific and misuse resistant, but I'm not an expert in OAuth2 or OIDC.)

README.md
tgeoghegan: I think OIDC audience is more correct than AWS audience.

jsha (approved):

PR #148, opened 2020-11-03
More robust paired test env

tgeoghegan:
This PR makes it easier to set up paired test environments via Terraform, with fake ingestion servers and also buckets into which sum parts can be written. It also removes ALL default values from facilitator (except that generate-ingestion-sample will still generate batch UUIDs and use the current date when constructing object keys), to make misconfigurations more obvious.

### DIFF
--- a/facilitator/src/bin/facilitator.rs
+++ b/facilitator/src/bin/facilitator.rs
@@ -511,6 +511,7 @@ fn main() -> Result<(), anyhow::Error> {
                         .long("aggregation-start")
                         .value_name("DATE")
                         .help("Beginning of the timespan covered by the aggregation.")
+                        .required(true)
                         .validator(date_validator),
                 )
                 .arg(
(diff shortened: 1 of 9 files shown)


facilitator/src/bin/facilitator.rs
```diff
@@ -540,21 +511,13 @@ fn main() -> Result<(), anyhow::Error> {
                         .long("aggregation-start")
                         .value_name("DATE")
                         .help("Beginning of the timespan covered by the aggregation.")
-                        .long_help(
-                            "Beginning of the timespan covered by the \
-                            aggregation. If omitted, the current time is used.",
-                        )
                         .validator(date_validator),
```
jsha, line 514: You didn't mark this one required. Was that intentional?

jsha (approved):

facilitator/src/bin/facilitator.rs
tgeoghegan: No, just an oversight, thank you for catching that. I tagged this and `aggregation-end` as required.

PR #152, opened 2020-11-04
workflow-manager: separate bucket reading from batch filtering

jsha:
This defines a new type `bucket`, which abstracts over the S3 vs
GS backend, and offers a `listFiles` method. It also provides new
methods to filter batches by readiness, introduces a notion of `infix`
to cover the `.batch.avro` / `.validity_0.avro` / `.validity_1.avro`
distinctions, and moves the bool fields used to calculate readiness
into `batchPath`. Also, batches are passed around as `batchPath`
objects now rather than strings.

### DIFF
--- a/workflow-manager/main.go
+++ b/workflow-manager/main.go
@@ -58,9 +58,12 @@ type batchPath struct {
 	dateComponents []string
 	ID             string
 	time           time.Time
+	metadata       bool
+	avro           bool
+	sig            bool
 }
 
-func newBatchPath(batchName string) (batchPath, error) {
+func newBatchPath(batchName string) (*batchPath, error) {
 	// batchName is like "kittens-seen/2020/10/31/20/29/b8a5579a-f984-460a-a42d-2813cbf57771"
 	pathComponents := strings.Split(batchName, "/")
 	batchID := pathComponents[len(pathComponents)-1]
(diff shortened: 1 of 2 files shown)


workflow-manager/main.go
```diff
@@ -58,38 +58,45 @@ type batchPath struct {
 	dateComponents []string
 	ID             string
 	time           time.Time
+	metadata       bool
+	avro           bool
+	sig            bool
 }
 
-func newBatchPath(batchName string) (batchPath, error) {
+func newBatchPath(batchName string) (*batchPath, error) {
```
tgeoghegan, line 66: I thought it was Go convention to return a value from a `new...` function instead of a pointer. I guess it doesn't matter in a GC language, and I think your Go is much sharper than mine.

workflow-manager/main.go
```diff
@@ -174,9 +170,18 @@ func main() {
 	log.Printf("starting %s version %s - %s. Args: %s", os.Args[0], BuildID, BuildTime, os.Args[1:])
 	flag.Parse()
 
-	ingestorService, ingestorInputBucket := checkStoragePath("--ingestor-input", *ingestorInput)
-	ownValidationService, ownValidationInputBucket := checkStoragePath("--own-validation-input", *ownValidationInput)
-	peerValidationService, peerValidationInputBucket := checkStoragePath("--peer-validation-input", *peerValidationInput)
```
tgeoghegan, line 179: I believe these were the only usages of the `checkStoragePath` function, which can now be deleted.

tgeoghegan (approved):

workflow-manager/main.go
jsha: This is actually a weird thing about Go that bites everybody new to the language. I was very confused about it for years.

Basically when you write a type `Foo`, you get to choose whether to pass it around as a pointer or a value. You express this choice mostly indirectly - by defining your methods with pointer receivers vs value receivers. If your methods have pointer receivers, your `newFoo` should return a pointer; if they have value receivers, your `newFoo` should return a value.

This is somewhat masked by the fact that Go does automatic dereferences: https://golang.org/ref/spec#Selectors.

So how do you choose whether Foo should be manipulated as a pointer vs a value? One simple rule of thumb: If the type is smaller than two pointers, you can save some heap allocations and generally be faster by making it a value. If it's bigger, go ahead and use a pointer.

The other upside of using a pointer, which I used here, is that in all the error cases, you can return `nil` instead of `foo{}`, which is tidier and easier to type.

PR #154, opened 2020-11-05
Rename peer_share_processor_names -> localities.

jsha:
Also, change job names for workflow-manager and sample-maker. We now put
the category of job (workflow-manager) first, so similar things group
together. Also, this eliminates the redundant copy of the locality name
in the job names - locality name is already in the namespace. New name
format:

workflow-manager-[INGESTOR]-[ENV]

Also, replace the example locality names in the demo tfvars with some
fictional places, so the role of the locality is clearer (since PHA is somewhat
overloaded). And use three localities instead of two to make it clear that
the N anticipated localities are different from the two anticipated ingestors
and two anticipated peers.


aaomidi (approved):
I like this, easier to understand

tgeoghegan (approved):

PR #161, opened 2020-11-05
Support multiple aggregations.

jsha:


jsha:
This is ready to review; I've manually tested it and it works. I haven't tested it with multiple input aggregation IDs yet, FYI, but I feel reasonably confident it will work.

tgeoghegan (approved):
One comment about aggregation job name uniqueness but we can deal with that in a later PR.

PR #163, opened 2020-11-05
workflow-manager: add cursoring to S3 API.

jsha:
Also:

- sort batches before working on them so output is more consistent
  and easier to read.
- use withinInterval to filter out too-old batches, rather than
  logging a line for each one.
- shorten the String() implementation for batchPath.
- remove some dead types.

### DIFF
--- a/terraform/modules/kubernetes/kubernetes.tf
+++ b/terraform/modules/kubernetes/kubernetes.tf
@@ -12,7 +12,7 @@ variable "environment" {
 
 variable "container_registry" {
   type    = string
-  default = "letsencrypt"
+  default = "us.gcr.io/jsha-prio-bringup/letsencrypt"
 }
 
 variable "workflow_manager_image" {
@@ -366,7 +366,7 @@ resource "kubernetes_cron_job" "sample_maker" {
               ]
               env {
                 name  = "RUST_LOG"
-                value = "1"
+                value = "info"
               }
               env {
                 name  = "RUST_BACKTRACE"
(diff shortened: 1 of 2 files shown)


terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -12,7 +12,7 @@ variable "environment" {
 
 variable "container_registry" {
   type    = string
-  default = "letsencrypt"
+  default = "us.gcr.io/jsha-prio-bringup/letsencrypt"
```
tgeoghegan, line 15: I don't think we want this on `main`

workflow-manager/main.go
```diff
@@ -340,13 +350,30 @@ func (b *bucket) listFilesS3(ctx context.Context) ([]string, error) {
...
-		return nil, fmt.Errorf("Unable to list items in bucket %q, %w", b.bucketName, err)
-	}
 	var output []string
-	for _, item := range resp.Contents {
-		output = append(output, *item.Key)
+	var nextContinuationToken string = ""
+	for {
+		// We choose a lower number than the default max of 1000 to ensure we exercise the
+		// cursoring case regularly.
+		var hundred int64 = 100
```
tgeoghegan, line 358: nit: You could use [`aws.Int64`](https://godoc.org/github.com/aws/aws-sdk-go-v2/aws#Int64) and avoid this

tgeoghegan (changes requested):

tgeoghegan (approved):

PR #168, opened 2020-11-06
Create deploy-operator and manifest-updater

aaomidi:
This is the code for:

- deploy-operator
- manifest-updater
- terraform changes for the deploy-operator

Most code under deploy-operator is generated using [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder). Interesting pieces of code under the deploy operator are:

- `deploy-operator/api/v1/locality_types.go`
- `deploy-operator/config/samples/prio_v1_locality_narnia.yaml`
- `deploy-operator/controllers/locality_controller.go`

I've created a `locality_kubernetes` module, which is enumerated based upon each locality, instead of the existing terraform module that was enumerated based on the data share processor for each locality. 

Reason: The manifest writer service accounts are going to be locality enumerated, rather than locality-data_share_processor enumerated.


deploy-operator/.gitignore
```diff
@@ -0,0 +1,24 @@
...
+
+# Kubernetes Generated files - skip generated files, except for vendored files
+
+!vendor/**/zz_generated.*
+
+# editor and IDE paraphernalia
+.idea
+*.swp
+*.swo
+*~
```
jsha, line 24: A lot of this is redundant with the repo-level .gitignore. Let's just put it all there.

deploy-operator/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# Build the manager binary
+FROM golang:1.13 as builder
```
jsha, line 2: Let's use go1.15.

deploy-operator/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# Build the manager binary
```
jsha, line 1: Let's name the binary `deploy-operator`. If we call it `manager`, that's confusing with `workflow-manager`.

deploy-operator/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+WORKDIR /workspace
+# Copy the Go Modules manifests
+COPY go.mod go.mod
+COPY go.sum go.sum
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+RUN go mod download
+
+# Copy the go source
+COPY main.go main.go
```
jsha, line 13: This seems like it could lead to headaches if we add additional files in the root directory, or additional subdirectories. I'd recommend just `COPY . .`

deploy-operator/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+COPY api/ api/
+COPY controllers/ controllers/
+
+# Build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
```
jsha, line 18: I think `GO111MODULE=on` is unnecessary since there's a go.mod in this directory. `-a` forces rebuilding of packages that are already up-to-date and isn't necessary here either.

deploy-operator/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:nonroot
```
jsha, line 22: We can do `FROM scratch` here to have an even more minimal install; I don't think we need distroless. See workflow-manager for an example.

deploy-operator/Makefile
```diff
@@ -0,0 +1,80 @@
+
```
jsha, line 1: I would prefer to avoid doing this in make if we can avoid it. Make is good for managing dependencies, but not for building ad-hoc commandline interfaces. In this case, since there are no dependency trees being managed by Make, I think this would be better as a shell script.

deploy-operator/api/v1/locality_types.go
```diff
@@ -0,0 +1,67 @@
...
+limitations under the License.
+*/
+
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
jsha, line 23: Was this from a template you forgot to edit out?

deploy-operator/api/v1/locality_types.go
```diff
@@ -0,0 +1,67 @@
+/*
+
+
```
jsha, line 3: Unnecessary blank lines.

jsha (commented):
Thanks for writing this! Partial review below, will send more tomorrow.

deploy-operator/Dockerfile
aaomidi: I'm not sure I want to do this. 
This project comes with a ton of random files here and there and it feels wrong to just bring them into the dockerfile if they're not necessary. What do you think?

I'm not 100% sure on what I'm saying either though, so feel free to override this!

deploy-operator/Makefile
aaomidi: Yeah this is one of the things that just came with kubebuilder, we don't really need `make`, it's just a couple of commands.

deploy-operator/api/v1/locality_types.go
aaomidi: Yep!

deploy-operator/Dockerfile
jsha: This Dockerfile uses the builder pattern, so the final image that gets shipped contains only the binary and some minimal dependencies. Copying the random files in here won't affect the final output. Given that most people's mental model of the docker build is "it builds based on the files in this directory," rather than "it builds based on the subset of files listed in Dockerfile," I think copying everything is worthwhile.

aaomidi: Sounds good to me then! Thanks for that explanation :)

manifest-updater/secrets/secret.go
```diff
@@ -0,0 +1,174 @@
...
+		err := sApi.Delete(context.Background(), secret.Name, v1.DeleteOptions{GracePeriodSeconds: nil})
+		if err != nil {
+			return fmt.Errorf("problem with deleting secret %s: %w", secret.Name, err)
+		}
+	}
+
+	return nil
+}
+
+func (k *Kube) getSortedSecretsWithLabel(labelSelector string) ([]corev1.Secret, error) {
```
tgeoghegan, line 148: This deserves a doccomment explaining that it sorts secrets by creation date

terraform/modules/data_share_processor/data_share_processor.tf
```diff
@@ -376,6 +380,22 @@ resource "google_storage_bucket_iam_binding" "own_validation_bucket_admin" {
   ]
 }
 
+resource "google_storage_bucket_object" "global_manifest" {
```
tgeoghegan, line 383: This should be `"specific_manifest"`, I think.

terraform/modules/locality_kubernetes/kubernetes.tf
```diff
@@ -0,0 +1,110 @@
+variable "environment" {
```
tgeoghegan, line 1: Could you please move the other per-locality Kubernetes resources created in `main.tf` into this module? `resource "kubernetes_namespace" "namespaces"` and `resource "kubernetes_secret" "ingestion_packet_decryption_keys"` would fit in here well. The values that are then needed in `data_share_processors.tf` can be exposed from this module as output variables.

terraform/main.tf
```diff
@@ -216,6 +216,19 @@ locals {
...
+# each locality/namespace
+module "locality_kubernetes" {
+  for_each             = kubernetes_namespace.namespaces
+  source               = "./modules/locality_kubernetes"
+  environment          = var.environment
+  gcp_project          = var.gcp_project
+  manifest_bucket      = module.manifest.bucket
+  kubernetes_namespace = each.value.metadata[0].name
+
+  depends_on = [module.manifest]
```
tgeoghegan: I believe Terraform can infer this dependency because the `manifest_bucket` parameter references an output variable on `module.manifest`.

deploy-operator/Makefile
tgeoghegan: If I remember our conversation from a while ago correctly, we only need the makefile to generate boilerplate from some kind of template, right? And the other targets in here are going to be replaced by `provisioner` blocks that emit Kubernetes YAML which we will `kubectl apply` into place. If that's true, I'd like to see this `Makefile` trimmed down to just what's needed for codegen, and then we should have a README in here explaining when codegen needs to be re-run and how to do it.

manifest-updater/secrets/batchSigning.go
```diff
@@ -0,0 +1,101 @@
...
+		UTC().
+		Format(time.RFC3339)
+
+	secret := corev1.Secret{
+		ObjectMeta: v1.ObjectMeta{
+			GenerateName: name,
+			Namespace:    k.namespace,
+			Labels: map[string]string{
+				"type": "batch-signing-key",
+				"dsp":  dsp,
```
tgeoghegan, line 80: `dsp` seems unnecessarily terse. Could we put `instance-name` instead?

manifest-updater/secrets/packetEncryption.go
```diff
@@ -0,0 +1,89 @@
...
+	"encoding/base64"
+	"fmt"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"time"
+)
+
+const (
+	//packetDecryptionKeyMaxAge = 20 * 24 * time.Hour
+	packetDecryptionKeyMaxAge = 20 * time.Second
```
tgeoghegan, line 14: I definitely don't think we want to rotate every 20 seconds!

terraform/modules/locality_kubernetes/kubernetes.tf
```diff
@@ -0,0 +1,141 @@
...
+
+# Bind the GCP and K8s service accounts together
+resource "google_service_account_iam_binding" "manifest_updater_workload" {
+  provider           = google-beta
+  service_account_id = google_service_account.manifest_updater.name
+  role               = "roles/iam.workloadIdentityUser"
+  members = [
+    local.manifest_updater_sa
+  ]
+}
```
tgeoghegan, line 64: I don't love having this copy pasted from `kubernetes/kubernetes.tf`. I think maybe you could refactor this into a module that creates a GCP SA, a Kubernetes SA and then brings them together, then use that from both this module and `kubernetes/kubernetes.tf`.

tgeoghegan (changes requested):
So, if I understand right, once we merge this, the new operator and manifest updater bits will be inert, because nothing is yet putting those CRDs into the cluster. If I took this and `make apply`ed it against an existing cluster, I'd get some new service accounts created due to the `locality_kubernetes/kubernetes.tf` module, but that's it, right?

deploy-operator/Makefile
aaomidi: The makefile is much shorter than what it was initially, but I'll remove some of the other targets too.

deploy-operator/config/manager/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+resources:
+- manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: us.gcr.io/prio-bringup-290620/deploy-operator
```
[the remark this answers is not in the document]
aaomidi: This is the image names in Google Container Registry for our project. I'll convert them to the Dockerhub one. Wasn't sure if we wanted to immediately do that or do it later.

aaomidi:
@tgeoghegan That's right, and you'd get a `crds.yml` file.

terraform/modules/locality_kubernetes/kubernetes.tf
```diff
@@ -0,0 +1,141 @@
...
+    }
+  })
+}
+
+resource "null_resource" "crd" {
+  triggers = {
+    applied_crd  = local.crd
+  }
+  provisioner "local-exec" {
+    command = "echo '${local.crd}\n---\n' >> crds.yml"
```
aaomidi, line 139: I really don't like this way of making the file. If the `local.crd` has a `'` this is going to fail/create arbitrary yamls. Potentially a security issue if someone gets a carefully crafted string in one of those variables used to create the crd.

manifest-updater/secrets/packetEncryption.go
aaomidi: Yep! Fragments of testing.

manifest-updater/secrets/batchSigning.go
aaomidi: That should actually be ingestor. I think everywhere you see datashareprocessor or the like should be ingestor.

manifest-updater/secrets/batchSigning.go
```diff
@@ -0,0 +1,101 @@
...
+	"fmt"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"time"
+)
+
+const (
+	batchSigningKeyExpirationDuration = 90 * 24 * time.Hour
+	// TODO: This needs to change to something like 70 days in prod
+	batchSigningKeyMaxAge = 20 * time.Second
```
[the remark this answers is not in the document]
aaomidi: There's some problems with making this configurable because then we're going to start getting into conflict with the expiration duration. I want to rotate earlier than the expiration, but if I rotate too early then we might have already deleted a key that we considered "valid".

And on the point of configuration, how do you want the config to pass into the manifest updater? Do we introduce a new field in the CRD that defines the expiration date for these? What do we do if the CRD updates with a shorter expiration date, which expiration date do we recognize to be valid? Especially if we've advertised the other key before.

There's a lot of questions with making this configurable, we should probably write out the various scenarios before writing the code for it.

manifest-updater/manifest/updater.go
```diff
@@ -0,0 +1,117 @@
...
+		log:                    log.WithField("source", "manifest updater"),
+		environmentName:        environmentName,
+		locality:               locality,
+		manifestBucketLocation: manifestBucketLocation,
+		dataShareProcessors:    dataShareProcessors,
+	}, nil
+}
+
+func (u *Updater) UpdateDataShareSpecificManifest(keys map[string]BatchSigningPublicKeys, certificate PacketEncryptionKeyCSRs) error {
+	if keys == nil && certificate == nil {
```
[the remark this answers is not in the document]
aaomidi: Yes, they may have different expiries so they get updated independently of each other.

deploy-operator/go.mod
```diff
@@ -0,0 +1,13 @@
+module github.com/abetterinternet/prio-server/deploy-operator
+
+go 1.13
```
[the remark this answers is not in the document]
aaomidi: I don't think there is any reason. It was just the version that the code generator created. I'll play around with 1.15 and just do a quick verification to make sure everything keeps working.

deploy-operator/controllers/locality_controller.go
```diff
@@ -0,0 +1,222 @@
...
+func (r *LocalityReconciler) scheduleNewJob(locality *priov1.Locality) (ctrl.Result, error) {
+	cronJob := r.createCronJobTemplate(locality)
+
+	if err := r.Client.Create(r.ctx, &cronJob); err != nil {
+		r.Log.Error(err, "unable to create a new CronJob")
+		return ctrl.Result{}, err
+	}
+
+	job := r.createJobFromCronJob(&cronJob)
+	if err := r.Client.Create(r.ctx, &job); err != nil {
```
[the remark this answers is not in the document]
aaomidi: Yep the second a CRD is added this will pick it up and create the cronjob and the job

deploy-operator/controllers/locality_controller.go
```diff
@@ -0,0 +1,222 @@
...
+	env := []v12.EnvVar{
+		{Name: "KR_ENVIRONMENT_NAME", Value: locality.Spec.EnvironmentName},
+		{Name: "KR_MANIFEST_BUCKET_LOCATION", Value: locality.Spec.ManifestBucketLocation},
+		{Name: "KR_DATA_SHARE_PROCESSORS", Value: strings.Join(locality.Spec.DataShareProcessors, " ")},
+		{Name: "KR_LOCALITY", ValueFrom: &v12.EnvVarSource{
+			FieldRef: &v12.ObjectFieldSelector{
+				FieldPath: "metadata.namespace",
+			},
+		}},
+	}
```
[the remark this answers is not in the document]
aaomidi: I'm not a big fan of making a ConfigMap because that means the operator now needs to keep track of two things for state management, effectively doubling how much work it needs to do. It also means that the operator needs to get extra permissions to read/write ConfigMaps

deploy-operator/controllers/locality_controller_test.go
```diff
@@ -0,0 +1,143 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+	v1 "github.com/abetterinternet/prio-server/deploy-operator/api/v1"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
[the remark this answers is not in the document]
aaomidi: Allows me to use things like `Context`, `In` etc without specifying the module name. E.g.:
```
	Context("When pushing and updating a new locality", func() {
		It("Should validate lifecycle of CRDs", func() {
			By("Creating a Locality")
```

aaomidi:
What remains from the comments that I haven't addressed yet:

- [ ] Consolidating more terraform things - Will address this tomorrow
- [ ] Figuring out the various edge cases with configurable expiration/rotation times - I'll need @tgeoghegan's help with this. Maybe a quick call?
- [ ] Triple checking that `ingestor` is the valid word to use - @tgeoghegan can you verify this is okay?
- [ ] Discussion on ConfigMap vs Environment variables for the jobs - @tgeoghegan Let's discuss if the extra state management buys us anything here.

terraform/modules/locality_kubernetes/kubernetes.tf
aaomidi: Will address this in a future PR.

deploy-operator/controllers/locality_controller_test.go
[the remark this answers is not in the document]
tgeoghegan: I didn't know that about Go, neat!

tgeoghegan:
>Triple checking that `ingestor` is the valid word to use - @tgeoghegan can you verify this is okay?
I'm not sure that it is. You seem to be using `ingestor` to mean an entity that has a batch signing key. Now, ingestors do have such keys, because they use them to sign ingestion batches that they send to data share processors (i.e., us). But from reading the code here, it seems like you're using ingestor key to describe the keys that go into a specific manifest. All of our keys belong to data share processors, and further they aren't per-ingestor but per instance. That is, our `co-apple` instance has a different batch signing key than `co-g-enpa` or `ma-apple` or `az-g-enpa`.

>Discussion on ConfigMap vs Environment variables for the jobs - @tgeoghegan Let's discuss if the extra state management buys us anything here.
I'm fine with using env vars. 

>Figuring out the various edge cases with configurable expiration/rotation times - I'll need @tgeoghegan's help with this. Maybe a quick call?
What cases are you worried about?

deploy-operator/Makefile
tgeoghegan: We still need this makefile trimmed and some documentation of how the codegen works.

deploy-operator/controllers/locality_controller.go
```diff
@@ -0,0 +1,233 @@
...
+	Scheme *runtime.Scheme
+
+	ctx context.Context
+}
+
+const (
+	jobNameField    = ".metadata.name"
+	jobName         = "manifest-updater"
+	keyRotatorImage = "letsencrypt/prio-manifest-updater:latest"
+	// keyRotatorImage    = "us.gcr.io/prio-bringup-290620/manifest-updater:latest"
```
tgeoghegan, line 50: The commented out line should be deleted

terraform/modules/account_mapping/account_mapping.tf
```diff
@@ -0,0 +1,77 @@
+variable "google_account_name" {
```
tgeoghegan, line 1: Please restore the comments explaining why we're associating all these objects to each other as well as the links to GCP documentation.

tgeoghegan (changes requested):
Looking better! A few more comments, and there's some unaddressed comments from prior reviews, as well.

deploy-operator/README.md
```diff
@@ -0,0 +1,18 @@
+# deploy-operator
+
+This deploy-operator handlaes the key creation and rotation jobs for each locality we deploy to Kubernetes.
```
tgeoghegan, line 3: "handles". Also this README should discuss how the code generation works. What are the inputs -- does the YAML generate the Go, or is it the magic comments in the Go that generate other Go files? When would codegen need to be re-run? How does one run codegen? I'm less concerned with the `deploy` and `install` targets because in any case, those ought to get removed in a subsequent PR, no?

aaomidi: I don't plan on removing them in subsequent PRs. Eventually this will be integrated with Helm.

I also don't want to repeat the documentation of kubebuilder in this readme. Do you think a link to kubebuilder would suffice?

aaomidi: https://github.com/abetterinternet/prio-server/pull/168/commits/81f7e4546ca3dc3a946513cf994d0d7cbb5c245b Added a quick basic overview of the files

tgeoghegan (approved):

PR #169, opened 2020-11-06
add region code to batch headers

gurayAlsac:


jsha:
It looks like there are legitimately failing tests here, e.g.: https://github.com/abetterinternet/prio-server/pull/169/checks?check_run_id=1364959787

Do you have what you need to get the tests passing?

gurayAlsac:
> It looks like there are legitimately failing tests here, e.g.: https://github.com/abetterinternet/prio-server/pull/169/checks?check_run_id=1364959787
> 
> Do you have what you need to get the tests passing?

This PR was mostly to spur discussion around how ingestors might pass on some sense of region/locality through data processors so ultimately it's available for the visualization tools. It sounds like from the chat that we're preferring to have data processors handle this mapping (perhaps based on the bucket created for the PHA) and not have it be contained in the files themselves.

PR #170, opened 2020-11-06
facilitator: linter for manifests

tgeoghegan:
Adds a `lint-manifest` subcommand to facilitator that can parse and
validate all kinds of manifests from either local files or fetches over
HTTPS.

I was going to write a pretty-printer and validator for Avro messages, too, but then I learned that Avro already publishes a Java jar that can do that (https://www.michael-noll.com/blog/2013/03/17/reading-and-writing-avro-files-from-the-command-line/).

### DIFF
--- a/facilitator/src/bin/facilitator.rs
+++ b/facilitator/src/bin/facilitator.rs
@@ -6,14 +6,17 @@ use ring::signature::{
     EcdsaKeyPair, KeyPair, UnparsedPublicKey, ECDSA_P256_SHA256_ASN1,
     ECDSA_P256_SHA256_ASN1_SIGNING,
 };
-use std::{collections::HashMap, str::FromStr};
+use std::{collections::HashMap, fs::File, str::FromStr};
 use uuid::Uuid;
 
 use facilitator::{
     aggregation::BatchAggregator,
-    config::{Identity, StoragePath},
+    config::{Identity, ManifestKind, StoragePath},
     intake::BatchIntaker,
-    manifest::{IngestionServerGlobalManifest, PortalServerGlobalManifest, SpecificManifest},
+    manifest::{
+        DataShareProcessorGlobalManifest, IngestionServerGlobalManifest,
+        PortalServerGlobalManifest, SpecificManifest,
+    },
     sample::generate_ingestion_sample,
     transport::{
         GCSTransport, LocalFileTransport, S3Transport, SignableTransport, Transport,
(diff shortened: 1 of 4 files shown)


facilitator/src/bin/facilitator.rs
```diff
@@ -539,6 +542,50 @@ fn main() -> Result<(), anyhow::Error> {
...
+                    .possible_value(leak_string(ManifestKind::IngestorGlobal.to_string()))
+                    .possible_value(leak_string(ManifestKind::DataShareProcessorGlobal.to_string()))
+                    .possible_value(leak_string(ManifestKind::DataShareProcessorSpecific.to_string()))
+                    .possible_value(leak_string(ManifestKind::PortalServerGlobal.to_string()))
+                    .required(true)
+                )
+            .arg(
+                Arg::with_name("instance")
+                    .long("instance")
+                    .value_name("LOCALITY-TECH_GIANT")
```
jsha, line 580: For consistency let's say:

```suggestion
                    .value_name("LOCALITY-INGESTOR")
```

facilitator/src/bin/facilitator.rs
```diff
@@ -539,6 +542,50 @@ fn main() -> Result<(), anyhow::Error> {
...
+            )
+            .arg(
+                Arg::with_name("manifest-path")
+                    .long("manifest-path")
+                    .value_name("PATH")
+                    .help("path to local manifest file to lint")
+                    .required_unless("manifest-base-url")
+            )
+            .arg(
+                Arg::with_name("manifest-kind")
```
jsha, line 567: Since all these flag names are scoped to the `lint-manifest` subcommand, let's drop the `manifest-` prefix and save some typing.

facilitator/src/bin/facilitator.rs
```diff
@@ -793,6 +841,76 @@ fn aggregate(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
...
 }
 
+fn lint_manifest(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
+    let manifest_base_url = sub_matches.value_of("manifest-base-url");
+    let manifest_file_reader = match sub_matches.value_of("manifest-path") {
+        Some(path) => Some(File::open(path).context("failed to open manifest file")?),
+        None => None,
+    };
+
+    let manifest_kind = ManifestKind::from_str(
```
jsha, line 851: I don't think this needs changing but wanted to point out an alternate syntax that's sometimes nice:

```
let kind: ManifestKind = sub_matches.value_of("manifest-kind").context("manifest-kind is required")?.parse()?;
```

That gets you a couple of things:
 - The type is annotated on the variable
 - It's a linear chain of calls rather than nested parens.

facilitator/src/bin/facilitator.rs
```diff
@@ -793,6 +841,76 @@ fn aggregate(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
...
+            let manifest = if let Some(base_url) = manifest_base_url {
+                DataShareProcessorGlobalManifest::from_https(base_url)?
+            } else if let Some(reader) = manifest_file_reader {
+                DataShareProcessorGlobalManifest::from_reader(reader)?
+            } else {
+                return Err(anyhow!(
+                    "one of manifest-base-url or manifest-path is required"
+                ));
+            };
+            println!("{:#?}", manifest);
```
jsha, line 880: There's some variation among what gets printed for the different cases; is that intentional?

jsha (approved):

facilitator/src/bin/facilitator.rs
tgeoghegan: Yes: there's no validate method on this one kind of manifest.

facilitator/src/bin/facilitator.rs
tgeoghegan: This does read more nicely, good tip!

PR #172, opened 2020-11-06
facilitator: set Cargo version based on tag version.

jsha:


tgeoghegan (approved):
It kind of bugs me that the crate version won't be reflected in the Cargo.toml checked into git, but that would be a much more complex overhaul of the release action.

PR #174, opened 2020-11-09
facilitator: use string for GCP service account ID

tgeoghegan:
Google is a big data company, so the numeric unique IDs for GCP service
accounts don't fit into u64. This commit changes the type of the
google_service_account field of IngestionServerIdentity to Option<u128>
to make sure we can parse account IDs.


hostirosti:
LGTM but would prefer using string with regex check.

jsha (approved):
Will this require demo deployments to re-run deploy-tool?

tgeoghegan:
It will not, because this only affects ingestor global manifests, which are not created or manipulated by our deployments, except in paired test env deploys, but in that case, the `google-service-account` value is omitted from the synthetic ingestor global manifests.

PR #177, opened 2020-11-09
specify defaults on types unioned with null

gurayAlsac:


gurayAlsac:
As far as I can tell this seems to be a bug in the rust Avro library. Avro canonicalizes the json schema definition (so serialized blobs can be checked as comparable) but the rust implementation chokes on `null`:
```panicked at 'got invalid JSON value for canonical form of schema: null', ```

But null is a valid JSON value in the latest JSON spec (https://www.json.org/json-en.html):
"A value can be a string in double quotes, or a number, or true or false or null, or an object or an array."

And the underlying JSON library that avro-rs uses supports null values:
https://github.com/serde-rs/json/blob/bda64205e3abd2788d0e4006ceb94c7236da7c56/src/value/mod.rs#L119

But avro-rs doesn't include that in the enum it has when converting to the canonical form:
https://github.com/flavray/avro-rs/blob/2e24700c6fded6ac9c21e60bd72e3c8d1aab2fdc/src/schema.rs#L823

It would be nice to make all of these schema types unioned with `null` actually support defaults and actually be optional, but it might be easier in the interim to at least convert those unioned with `"string"` to just be primitive string types that default to empty string. It looks like that's a slightly deeper change than I was hoping to make in the repo since that also touches your internal type in idl.rs:
https://github.com/abetterinternet/prio-server/blob/3db3697c0538b15a6c0ebb232f8dbde72c974ded/facilitator/src/idl.rs#L371

I think I have just enough grasp of Rust to make that change. The alternative would be to keep the union type (and optional) but still default to empty string, which feels a bit kludgy.

What do you all think?

tgeoghegan:
I would strongly prefer not to make any changes to the schema if we don't have to, because it's a hassle coordinating them across all the implementations. Can you remind me why we need this in the first place? Is it because the Avro library you're using can't write the `null` values unless the `default` is explicitly provided?

gurayAlsac:
At the moment the schema only partially expresses the intent that some of these fields are optional. The idea was to fully express that in the proper way, and to get the change in sooner rather than later when it only becomes harder to coordinate the change.

The Java avro library throws unless we explicitly set the field as null, which seems sorta silly to require if the field is truly optional:
`org.apache.avro.AvroMissingFieldException: Field encryption_key_id type:UNION pos:2 not set and has no default value`

However, because of what seems to be an Avro rust library bug, its more of an awkward change than I was hoping.

Maybe we can keep all of these fields as types unioned with null, but at least in the case of strings we can default to empty string (since null has problems)? I think fields like hamming weight are always specified anyway.

tgeoghegan:
For Narnia (i.e., Friday), I'd prefer not to touch the schemas, and it sounds like you have a viable workaround on the Java side, which is to explicitly encode the null values. We can talk about schema changes post-Narnia, but to level with you, I'm not sanguine about the prospect of getting any bugfixes into avro-rs in time for ENPA scaled launch, and I think everyone's implementation having to deal with explicitly encoding nulls is a better workaround than reserving the empty string as a sentinel value.

gurayAlsac:
Fair enough. If I were more familiar/competent with Rust I might at least create a minimal repro and file bug against avro-rs, but alas that'll have to wait. I'll retract the PR.

PR #178, opened 2020-11-10
Use env credentials if no identity is provided.

jsha:


workflow-manager/main.go
```diff
@@ -316,6 +316,24 @@ func (b *bucket) listFiles(ctx context.Context) ([]string, error) {
 	}
 }
 
+func webIDP(sess *aws_session.Session, identity string) (*credentials.Credentials, error) {
+	arnComponents := strings.Split(identity, ":")
```
tgeoghegan, line 320: This code isn't new, but could we use [arn.Parse](https://docs.aws.amazon.com/sdk-for-go/api/aws/arn/#Parse) to make this a little more robust? Just think of all the test cases you won't have to write later on if you make Amazon do the parsing for you!

tgeoghegan (approved):
I think the idea here is for NCI to pull this patch into their fork of the repo, so on that premise we don't need this for `release/narnia`, right?

If you opt not to adopt `arn.Parse` from the AWS SDK, could you please file an issue with the link to the SDK documentation so someone else can do it later?

jsha:
Good call, updated.

jsha:
> I think the idea here is for NCI to pull this patch into their fork of the repo, so on that premise we don't need this for release/narnia, right?

Actually, if NCI confirms this works, I'd like to merge it into `release/narnia`, so they can use it more easily and it can be included in Narnia release tags.

tgeoghegan:
SGTM, merge to either `main` or `release/narnia` at your discretion.

PR #179, opened 2020-11-10
facilitator: improve logging

jsha:
For GETs and PUTs, log not just the bucket name, but also the key.

Remove the derive(Debug)  on OauthTokenProvider and provided a custom
Debug impl, so we don't log the tokens.

Rename the fields of OauthTokenProvider to be shorter.

### DIFF
--- a/facilitator/src/transport/gcs.rs
+++ b/facilitator/src/transport/gcs.rs
@@ -8,7 +8,7 @@ use chrono::{prelude::Utc, DateTime, Duration};
 use log::info;
 use serde::Deserialize;
 use std::{
-    io,
+    fmt, io,
     io::{Read, Write},
 };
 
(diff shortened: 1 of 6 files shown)


facilitator/src/transport/gcs.rs
```diff
@@ -215,8 +225,15 @@ impl GCSTransport {
...
 
 impl Transport for GCSTransport {
+    fn path(&self) -> String {
+        self.path.to_string()
+    }
+
     fn get(&mut self, key: &str) -> Result<Box<dyn Read>> {
-        info!("get {} as {:?}", self.path, self.oauth_token_provider);
+        info!(
+            "get {}{} as {:?}",
```
tgeoghegan, line 234: nit: `get {}/{} as {:?}`, unless `self.path` is already guaranteed to be /-terminated.

tgeoghegan (approved):
I think we could take this into `release/narnia`, as it will be particularly useful in debugging integration problems over the next week.

jsha:
Rebased against the Narnia branch and merged there.

PR #185, opened 2020-11-10
Advertise ingestion writer role ARN from manifest

tgeoghegan:
In order for ingestors in GCP to be able to write to ingestion buckets,
they must know what AWS IAM role to assume. We were already creating the
necessary role and configuring ingestion buckets to allow it to write,
but we hadn't exposed the ARN in the specific manifest, which this
commit fixes.

### DIFF
--- a/deploy-tool/main.go
+++ b/deploy-tool/main.go
@@ -61,6 +61,10 @@ type SpecificManifest struct {
 	// IngestionBucket is the region+name of the bucket that the data share
 	// processor which owns the manifest reads ingestion batches from.
 	IngestionBucket string `json:"ingestion-bucket"`
+	// IngestionIdentity is the ARN of the AWS IAM role that should be assumed
+	// by an ingestion server to write to this data share processor's ingestion
+	// bucket, if the ingestor does not have an AWS account of their own.
+	IngestionIdentity string `json:"ingestion-identity"`
 	// PeerValidationBucket is the region+name of the bucket that the data share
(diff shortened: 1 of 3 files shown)


jsha:
Are we going to need identity information for all the other possible buckets in manifests? Do we want to adopt a scheme similar to what we use in facilitator, where buckets and identities come in pairs, and the identity can be either a GCP one or IAM one depending?

tgeoghegan:
That's an interesting idea. In some cases that would essentially be advisory (i.e. when ISRG creates a GCS ingestion bucket, we consult the ingestion server's manifest to figure out what GCP SA email address to grant access to) but it might be a good debugging aid to be able to glance at a manifest and see what AWS role or GCP service account is supposed to be able to read or write a bucket. I want to minimize change for Narnia, but we could incorporate that idea into the `format = 1` specific manifest.

jsha:
Agreed on minimizing changes for narnia. But as a nod in that direction, how about naming this field `ingestion-identity`? That comports with our flag naming in facilitator.

tgeoghegan:
Good idea! And to expand on my earlier argument rather than just gesturing at the concept of risk: if we add a single key as this commit does, then only Google has to deal with the schema change, whereas if we introduced a more robust structure that wraps both a bucket and the identity, Apple has to change their ingestion server by Friday, and they're busy enough as it is.

jsha (approved):

PR #193, opened 2020-11-12
facilitator: support GCP SA key files

tgeoghegan:
To enable our disginguished colleagues at NCI to write to Google Cloud
Storage buckets from outside of GCP, we add support for using a GCP
service account key file to obtain OAuth tokens for the default service
account. Previously, we only supported fetching an OAuth token from the
GKE metadata service, but now we can be provided the path to a service
account key file (as one might obtain through the Google Cloud console),
and will use the service account email and private key therein to
construct a signed JWT and then send a request to an OAuth API endpoint
to get an OAuth token suitable for authenticating in turn to the Google
Cloud Storage service.

This commit also fixes a bug where the logic for determining if an OAuth
token has expired was exactly backwards.


facilitator/src/transport/gcs.rs
```diff
@@ -82,15 +122,26 @@ impl fmt::Debug for OauthTokenProvider {
...
-        OauthTokenProvider {
+    fn new(
+        account_to_impersonate: Option<String>,
+        key_file_reader: Option<Box<dyn Read>>,
+    ) -> Result<OauthTokenProvider> {
+        let key_file: Option<ServiceAccountKeyFile> = match key_file_reader {
+            Some(reader) => {
+                serde_json::from_reader(reader).context("failed to deserialize JSON key file")?
+            }
+            None => None,
```
jsha, line 137: Nit: this could be `key_file_reader.map( |reader| { ... } )`

jsha (approved):

facilitator/src/transport/gcs.rs
tgeoghegan: I'm not sure it can, because we would need to return a `Result<>` from the closure passed to `Option::map` (since `serde_json::from_reader` may fail), and I don't think you can do that from an `FnOnce`.

jsha: Hm, good point about returning a result. I think you can (an `Fn` or `FnOnce` can return whatever you like), but type inference gets pretty shaky when it comes to closures, so you'd probably wind up having to annotate things to the extent it would be less clear instead of more. 👍🏻

PR #194, opened 2020-11-12
Add Prometheus monitoring.

jsha:
Note: Right now Prometheus is mainly monitoring itself. We need to
figure out how to push metrics from our cron jobs, since Prometheus
can't scrape them (by virtue of being cron jobs).


terraform/modules/monitoring/monitoring.tf
```diff
@@ -0,0 +1,135 @@
...
+
+  data = {
+    "prometheus.yml" = file("${path.module}/prometheus.yml")
+  }
+}
+
+resource "kubernetes_cluster_role" "prometheus_cluster_role" {
+  metadata {
+    name = "prometheus-role"
+    #namespace = "monitor"
```
tgeoghegan, line 70: Is this role not namespaced? There's a distinction between a `kubernetes_cluster_role` and a [`kubernetes_role`](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/role).

terraform/modules/monitoring/monitoring.tf
```diff
@@ -0,0 +1,135 @@
...
+    persistent_volume_reclaim_policy = "Retain"
+    persistent_volume_source {
+      gce_persistent_disk {
+        pd_name = google_compute_disk.prom_store.name
+      }
+    }
+  }
+}
+
+resource "google_compute_disk" "prom_store" {
```
tgeoghegan, line 129: nit: `prometheus_store`?

terraform/modules/monitoring/monitoring.tf
```diff
@@ -0,0 +1,135 @@
...
+  name = "prom-store"
+  type = "pd-ssd"
+  # Hack: We have a region in the vars, but to create a disk we want a zone. Just append "-a"
+  zone    = "${var.gcp_region}-a"
```
tgeoghegan, line 133: If the backing store for Prometheus data is in `$REGION-a`, what does that mean about the availability of Prometheus? Can it only run in zone a?

terraform/modules/monitoring/monitoring.tf
```diff
@@ -0,0 +1,135 @@
...
+  }
+}
+
+resource "kubernetes_config_map" "prometheus_config" {
+  metadata {
+    name = "prometheus-config"
+  }
+
+  data = {
+    "prometheus.yml" = file("${path.module}/prometheus.yml")
```
tgeoghegan, line 63: nit: it's a shame having some of this farmed out to a separate YAML file, especially since some of the values in there need to agree with values here in Terraform (e.g., the port numbers). I think you could use [yamlencode](https://www.terraform.io/docs/configuration/functions/yamlencode.html) to instead define all those parameters inline in Terraform.

tgeoghegan (dismissed):

terraform/modules/monitoring/monitoring.tf
jsha: Correct, this is meant to be a cluster role (not a role) because it has to be non-namespaced in order to reach into all the namespaces.

terraform/modules/monitoring/monitoring.tf
jsha: FWIW, I am more used to maintaining Prometheus config as its own YAML, and prefer it that way. For instance if we need to copy a config block from some example, we don't need to transmute it into HCL first.

The one parameter that needs agreement is the port, and 9090 is a standard port for Prometheus, so I'm not worried about writing it twice.

jsha:
Updates with some fixes to make this actually use a persistent store. I'll wait for a re-review before merging.

terraform/modules/monitoring/monitoring.tf
yuriks: GCP supports regional disks (they're just potentially slightly slower, but not by that much), so I think we should just use that for better availability. Unless those don't work here for some reason?

yuriks: Ah, seems like they're a just a different type of resource in the API (and therefore TF), so you'd miss that if just looking at `google_compute_disk`: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_region_disk

terraform/modules/monitoring/monitoring.tf
```diff
@@ -0,0 +1,181 @@
...
+    persistent_volume_reclaim_policy = "Retain"
+    persistent_volume_source {
+      gce_persistent_disk {
+        pd_name = google_compute_disk.prometheus_disk.name
+      }
+    }
+  }
+}
+
+resource "google_compute_disk" "prometheus_disk" {
```
yuriks, line 175: You should probably add `provider = google-beta` or this will use incorrect defaults (because we don't provide configuration for the default non-beta `google` one) and that tends to cause confusing error messages. (I plan on PRing a fix for this in general soon though.)

terraform/modules/monitoring/monitoring.tf
```diff
@@ -0,0 +1,181 @@
...
+    name      = "prometheus"
+    namespace = "monitor"
+  }
+  spec {
+    selector = {
+      app = kubernetes_pod.prometheus.metadata.0.labels.app
+    }
+    port {
+      port        = 9090
+      target_port = 9090
```
tgeoghegan, line 30: Could you add a comment explaining what is exposed on these ports? Is this the Prometheus API?

terraform/modules/monitoring/monitoring.tf
```diff
@@ -0,0 +1,181 @@
...
+    namespace = "monitor"
+    labels = {
+      app = "Prometheus"
+    }
+  }
+
+  spec {
+    service_account_name = kubernetes_service_account.prometheus_account.metadata[0].name
+    container {
+      image = "prom/prometheus:latest"
```
tgeoghegan, line 49: Is pulling `latest` safe? Does Prometheus have release channels that provide any kind of stability guarantees?

terraform/modules/monitoring/monitoring.tf
```diff
@@ -0,0 +1,181 @@
...
+        mount_path = "/config"
+      }
+      volume_mount {
+        name       = "storage-volume"
+        mount_path = "/data"
+      }
+    }
+    security_context {
+      fs_group        = 2000
+      run_as_user     = 1000
```
tgeoghegan, line 66: Could you include a comment explaining what these values are and why they're needed? I followed your investigation in the `gcp` channel on the ENPA slack and my takeaway was "ooh, that's a bit confusing".

terraform/modules/monitoring/monitoring.tf
```diff
@@ -0,0 +1,181 @@
...
+resource "kubernetes_persistent_volume_claim" "prometheus_disk_claim" {
+  metadata {
+    name      = "prometheus-disk-claim"
+    namespace = "monitor"
+  }
+  spec {
+    access_modes = ["ReadWriteOnce"]
+    resources {
+      requests = {
+        storage = "5Gi"
```
tgeoghegan, line 148: This is the volume that Prometheus is going to store time series data in, right? Any thoughts on whether 5 GB or 20 GB will be sufficient? I know that's a tricky capacity planning question given that we don't even yet know what kind of data we want to store here, but maybe we should make these parameters configurable so it's easy to grow these volumes in the future?

terraform/modules/monitoring/monitoring.tf
tgeoghegan: Sounds like we should use a regional disk, then.

tgeoghegan (changes requested):
I'm a dunce, and like having things explained to me in simple terms, so I'd love to see a discussion in the top level README explaining what we use Prometheus for and how it works, and how that impacts namespace and network topology. For instance, let's say we want `workflow-manager` to emit a data point to a metric every time it schedules an `intake-batch` job. How does that work? Are we going to have to run a Prometheus exporter in each pod that relays metrics to the central Prometheus? Or do I have it backwards and each pod will have to add a service that listens on 9090 so that the central Prometheus can scrape metrics from them? I suppose a lot of that will become obvious once we wire up the first metrics from a workflow-manager.

terraform/modules/monitoring/monitoring.tf
jsha: For DockerHub images, `latest` generally means "latest release." And Prometheus releases are pretty stable over time - no big API breaks. I think we're better off staying up-to-date with `latest` than pinning a version. Also, this is one of the recommended install methods: https://prometheus.io/docs/prometheus/latest/installation/

terraform/modules/monitoring/monitoring.tf
jsha: These are total shots in the dark. My plan is to deploy and measure how much we're using. As far as configurability - I think it's pretty easy to edit the hardcoded value here in the future, or turn it into a var.

One potential stumbling block: The "claim" is actually at 5Gi rather than the full 20Gi of its persistent disk. During testing I set it to be much lower, worried that the claim wasn't becoming bound because for some reason it was asking for too much. Later I tried increasing the claim, and ran into some errors that it couldn't be resized. Seems like we may need to do some extra stuff to allow resizes.

terraform/modules/monitoring/monitoring.tf
tgeoghegan: Thanks for double checking

jsha:
Closing in favor of #230.

PR #199, opened 2020-11-13
facilitator: Add AWS ChainProvider for WebIDP.

jsha:
Rusoto offers a "ChainProvider", which will try each of the various auth
methods available on AWS in order, and return as soon as one works. The
ChainProvider offered by Rusoto omits WebIdentityProvider, which we
need.

This change copies the implementation of ChainProvider, and adds
WebIdentityProvider to the list of providers tried.

### DIFF
--- a/facilitator/src/transport/s3.rs
+++ b/facilitator/src/transport/s3.rs
@@ -843,6 +843,7 @@ pub struct ChainProvider {
     instance_metadata_provider: InstanceMetadataProvider,
     container_provider: ContainerProvider,
     profile_provider: Option<ProfileProvider>,
+    webidp_provider: WebIdentityProvider,
 }
 
 impl ChainProvider {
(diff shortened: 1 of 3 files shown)


facilitator/src/transport/s3.rs
```diff
@@ -764,3 +766,153 @@ mod tests {
...
+            return Ok(creds);
+        }
+    }
+    if let Ok(creds) = provider.container_provider.credentials().await {
+        return Ok(creds);
+    }
+    if let Ok(creds) = provider.instance_metadata_provider.credentials().await {
+        return Ok(creds);
+    }
+    if let Ok(creds) = WebIdentityProvider::from_k8s_env().credentials().await {
```
tgeoghegan, line 874: I think we want to hold on to an instance of `WebIdentityProvider` in `ChainProvider`, so that the `WebIdentityProvider` can cache the OIDC token. That said all that amounts to is more calls to `os.Env` so maybe stateless is better!

tgeoghegan (approved):

facilitator/src/transport/s3.rs
jsha: Good catch, I was wondering why what I wrote didn't match the existing pattern. Fixed.

PR #201, opened 2020-11-16
Narnia: add ta-ta locality

tgeoghegan:
This PR contains config changes to support the `ta-ta` locality we are using for integration testing. None of this ever needs to be merged to main.


terraform/modules/data_share_processor/data_share_processor.tf
```diff
@@ -113,8 +113,8 @@ locals {
     ) : (
     var.ingestor_aws_role_arn
   )
-  ingestion_bucket_name       = "${local.resource_prefix}-ingestion"
-  peer_validation_bucket_name = "${local.resource_prefix}-peer-validation"
+  ingestion_bucket_name       = "${local.resource_prefix}-ingestion-a"
```
jsha, line 116: Why is this change here?

terraform/main.tf
```diff
@@ -16,7 +16,7 @@ variable "aws_region" {
 
 variable "aws_profile" {
   type    = string
-  default = "leuswest2"
+  default = "leuseast1"
```
jsha, line 19: This change seems like it's redundant with the setting in the tfvars?

terraform/modules/data_share_processor/data_share_processor.tf
tgeoghegan: We had to move the S3 buckets to `us-east-1` to work around a bug in Google's S3 client where they couldn't send content to two different regions (I think they since fixed it but I'm not going to move the buckets back). Then we couldn't create buckets with the same names in `us-east-1` as they already existed in `us-west-1`, so I added this suffix to work around that. As I said in the PR message, we shouldn't take this change in main since it only supports Narnia.

jsha (approved):

PR #207, opened 2020-11-18
facilitator: refactor manifest fetching.

jsha:
This adds a new get_url function that wraps ureq gets, turning ureq's
synthetic_errors into Results, and using anyhow to add context about
which URL failed.

This also changes the manifest reader functions to take a slice of bytes
instead of a reader.

### DIFF
--- a/facilitator/src/http.rs
+++ b/facilitator/src/http.rs
@@ -1,0 +1,22 @@
+use anyhow::{anyhow, Result};
+
+pub(crate) fn get_url(url: &str) -> Result<String> {
+    let resp = ureq::get(url)
+        // By default, ureq will wait forever to connect or
+        // read.
+        .timeout_connect(10_000) // ten seconds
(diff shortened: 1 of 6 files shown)


facilitator/src/http.rs
```diff
@@ -0,0 +1,22 @@
...
+        Err(anyhow!(
+            "fetching {}: {}",
+            url,
+            resp.into_synthetic_error().unwrap()
+        ))
+    } else if !resp.ok() {
+        Err(anyhow!("fetching {}: status {}", url, resp.status()))
+    } else {
+        resp.into_string()
+            .map_err(|e| anyhow!("reading body of {}: {}", url, e))
```
tgeoghegan, line 20: nit: Could you do `.context(format!("reading body of {}", url))` instead of `map_err`?

tgeoghegan (approved):

PR #209, opened 2020-11-18
workflow-manager: Use ToLower on job names.

jsha:


tgeoghegan:
I think this one can be closed in favor of #211 right?

PR #212, opened 2020-11-18
Merge release/narnia into main

jsha:


terraform/modules/data_share_processor/data_share_processor.tf
```diff
@@ -113,8 +113,8 @@ locals {
     ) : (
     var.ingestor_aws_role_arn
   )
-  ingestion_bucket_name       = "${local.resource_prefix}-ingestion"
-  peer_validation_bucket_name = "${local.resource_prefix}-peer-validation"
+  ingestion_bucket_name       = "${local.resource_prefix}-ingestion-a"
+  peer_validation_bucket_name = "${local.resource_prefix}-peer-validation-a"
```
tgeoghegan, line 117: We don't want this hunk on `main`. This is a hack I did just so we could move some buckets from one AWS region to another.

tgeoghegan (changes requested):

tgeoghegan (approved):

Issue #160, opened 2020-11-05
Opt into GCP APIs from Terraform


tgeoghegan:
When we bring up a new env in a new GCP project, we get a bunch of failures because various APIs (Compute, GKE, KMS, etc.) are not yet enabled. We should use [google_project_service](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_project_service) resources in Terraform to automatically enable APIs in a project.

tgeoghegan:
Resolved by the related PRs.

PR #219, opened 2020-11-19
terraform: Use private GKE clusters

### LINKED ISSUE #52: Use private network for GKE cluster
Our GKE cluster is currently configured to use public networks managed by Google/GKE. This means each worker node gets a public IP, and (I think) communication between worker nodes goes over the public internet. While this makes it trivial for our jobs to perform the egress they need (e.g. to AWS or GCP APIs or to fetch peer manifests), this is wasteful (ISRG is committed to environmentally responsible practices, which means reducing, reusing, recycling IPv4 addresses) and could be more secure. We should configure a private network for the GKE cluster and then narrowly control what egress and ingress we permit to worker nodes.

yuriks:
Fixes #52 
Also happens to very partially address #160 (only for the services I personally ran into while setting it up, and with caveats.)

I considered adding more restrictive egress firewall rules (the defaults are to allow all egress and no ingress) but they would probably just allow [80,443] to 0.0.0.0/0, which is so general as to seem somewhat pointless. Regardless we can still add something later on, especially once we don't have the data plane jobs fetching manifests anymore, since then we could allow only DockerHub, AWS STS and S3 access on the data plane, I think, instead of whitelisting whatever random IPs the other parties will use for manifests.

(Commit text reproduced below.)

---

Private clusters don't assign public IP addresses to individual cluster nodes. This greatly conserves valuable external IP addresses, but requires extra work to hook everything together since we can't use some of the defaults that GKE normally comes up with for you. It also has security benefits, because nodes are not addressable from the Internet.

To still allow external connectivity, which we need in order to load container images from DockerHub, fetch manifests from other parties, as well as using the AWS APIs, a Cloud NAT gateway provides outgoing-only Internet access using a smaller IP pool.

The cluster API servers still have public endpoints, so kubectl will continue to work as normal, including "exec", "port-forward", etc. However, nodes in the pool will not be directly reachable via SSH anymore. This should be fine since nodes are maintained by GKE itself. If the need arises in the future, a bastion host or Cloud VPN can be setup to provide this kind of access.

### DIFF
--- a/terraform/network.tf
+++ b/terraform/network.tf
@@ -1,0 +1,18 @@
+locals {
+  # Address block used by the GKE cluster and other regional resources.
+  # Addresses in 10.0.0.0/8 outside of this block are reserved for future
+  # expansion.
+  cluster_subnet_block = "10.64.0.0/11"
+}
+
+# A VPC is a global resource in GCP, however, subnets inside it are regional
+# and so are created for each region by the gke module. See gke/network.tf
+resource "google_compute_network" "network" {
+  provider = google-beta
+  # Add prefix to support existing multi-env GCP projects
+  name = "${var.environment}-network"
+  # We will always have to create a subnet for the cluster anyways, so the auto
+  # networks would never be used. This frees up the upper half of the 10.0.0.0/8
+  # block, where these auto subnets would otherwise be allocated.
+  auto_create_subnetworks = false
+}
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -116,6 +116,24 @@ provider "google-beta" {
   project = var.gcp_project
 }
 
+# Activate some services which the deployment will require. Note that while this
+# can save you from having to go to the GCP console and manually activate them,
+# often a new deploy will still fail due to a race condition between the service
+# being activated and terraform trying to use it. This could probably be greatly
+# ameliorated by adding manual dependencies between the service activation and
+# the resources, but the maintenance cost of that exceeds the benefit. For now,
+# simply retrying the apply after a minute should let it continue.
+
+resource "google_project_service" "compute" {
+  provider = google-beta
+  service  = "compute.googleapis.com"
+}
+
+resource "google_project_service" "container" {
+  provider = google-beta
+  service  = "container.googleapis.com"
+}
+
 provider "aws" {
   # aws_s3_bucket resources will be created in the region specified in this
   # provider.
@@ -146,6 +164,8 @@ module "gke" {
   gcp_region      = var.gcp_region
   gcp_project     = var.gcp_project
   machine_type    = var.machine_type
+  network         = google_compute_network.network.self_link
+  base_subnet     = local.cluster_subnet_block
 }
 
 # For each peer data share processor, we will receive ingestion batches from two
--- a/terraform/modules/gke/gke.tf
+++ b/terraform/modules/gke/gke.tf
@@ -19,6 +19,10 @@ variable "machine_type" {
 }
 
 resource "google_container_cluster" "cluster" {
+  # The google-beta provider is needed for some features we're using below, such
+  # as VPC-native clusters.
+  provider = google-beta
+
   name = "${var.resource_prefix}-cluster"
   # Specifying a region and not a zone here gives us a regional cluster, meaning
   # we get cluster masters across multiple zones.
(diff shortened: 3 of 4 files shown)


terraform/network.tf
```diff
@@ -0,0 +1,20 @@
+locals {
+  # Allocated per-region subnets, keyed by region
+  gcp_region_subnets = {
+    # Start at 2 to reserve some subnets for special uses if needed
+    "us-west1" = cidrsubnet("10.0.0.0/8", 3, 2)
```
tgeoghegan, line 5: I don't understand why we need this map. Does every env running in us-west1 share 10/8?

tgeoghegan (commented):
I'm pleased to see how small this change is, which doesn't mean it didn't take a bunch of work and research and experimentation, but I'm glad to see it's pretty straightforward and easy to follow. What is your sense of the risk level of this change? What things are likely to stop working? And if we deploy this to a cluster, is it easy to roll back?

terraform/network.tf
yuriks: Sorry, I should've probably explained this better in the PR. This is for if we want to run in more than one region later on:

Each environment in the project gets it's own VPC (being created in this file) from virtue of being a completely separate TF state/etc. VPCs are global resources and aren't associated with any particular location. They're also entirely separated networks, so, say, 10.0.0.0/8 in each VPC is subnet independent and they can't reach each other.

*Subnets* however are regional. Only hosts in the subnet's region can get addresses from it, so those need to be specific for a region. Therefore, inside the environment's VPC, I need to carve out blocks for at least each region we're running in, so that subnets can be created for each of them (in gke/network.tf).

However, in this case, our TF code actually only supports one region due to the way it's set up right now, but this is something I want to fix. (I'd started to do it for this PR, but it ended up being too much churn, so I split it off.) If we have more than one region, then it's probably quite likely that we'll want the regions to be able to see and communicate with each other, so they have to be on the same VPC. Therefore, it's not possible (at least cleanly) for us to use entirely separate tfvars/state storage for each region. At the same time, the multiple regions (when it's viable to support them in our TF setup) need to use separate address blocks so they don't conflict with each other, and so there needs to be a table of region -> address block somewhere.

---

I don't love the `3, 2` either. [They are, respectively](https://www.terraform.io/docs/configuration/functions/cidrsubnet.html) `newbits` and `netnum`. So the 3 means we want to turn the `/8` into a `/11` (which means we get 2^3=8 divisions of our `/8`), and, of those, choose the 2nd subnet. If we had multiple entries in the table it'd look something like this:

```
gcp_region_subnets = {
  # Reserved: 10.0.0.0/11
  # Reserved: 10.32.0.0/11
  "us-west1" = cidrsubnet("10.0.0.0/8", 3, 2) # 10.64.0.0/11
  "us-east1" = cidrsubnet("10.0.0.0/8", 3, 3) # 10.96.0.0/11
  "europe-west1" = cidrsubnet("10.0.0.0/8", 3, 4) # 10.128.0.0/11
  # etc.
  "europe-west2" = cidrsubnet("10.0.0.0/8", 3, 7) # 10.224.0.0/11
  # Now we've used up the entire address space.
}
```

I *think* there was a reason I didn't also use the `cidr` module like I did in gke/network.tf, but I can't remember what it was anymore. I'll update the change to use it and push a new version if there's no issue, since I think it should be a bit clearer than this.

yuriks:
Re. the risk of the change: As far as I know, it's unlikely to break anything. Individual pods never got their own external IPs anyway, only nodes did, and we don't run anything directly on nodes, so there's no chance we were relying on those external IPs for anything other than SSHing into nodes (the host. Getting a shell inside a pod using `kubectl` continues to work fine.) As far as I know no one in our team has any usecase for that other than for troubleshooting, for which we'll want a bastion VM or other SSH gateway in the VPC, but that's outside of the GKE cluster anyway. For outgoing connections (which we definitely do a bunch of), everything in the subnet will use the NAT as the default gateway, so that should also work fine.

The GKE cluster automatically adds firewall rules to allow the nodes to access all related subnets (k8s pods/services and the control plane) so there aren't any connectivity issues talking between nodes. However, if someone is manually creating Compute VMs in their project outside of the GKE cluster, then those would not be able to talk to the cluster unless they're also attached to the new VPC created here (instead of `default`) and a firewall rule is added. I'm not really aware of anyone doing this (and in any case I think it'd be outside of scope for this TF to support arbitrary untracked one-off additions people are making to their projects.)

**All that said**, it *is* a quite churny change: It requires re-creating the cluster (you can't just flip a cluster to be private since it involves pretty substantial changes in how the control plane servers are set up by GKE) and the same would be the case to rollback.

I assumed that since we branched off for narnia it's fine to do these sorts of breaking changes in main (and in fact, probably a good idea to do them now while we can!) but if recreating existing clusters is a problem then we can discuss a strategy to roll this out. Let me know what you think.

yuriks:
Pushed a new version with a few very minor formatting tweaks and also using the `cidr` module instead. It is a bit more obvious, but also significantly more verbose I suppose. Still fine as far as I'm concerned. PTAL

terraform/network.tf
tgeoghegan: > This is for if we want to run in more than one region later on
I think we shouldn't try to solve that problem (and take on the burden of testing and maintaining the solution) until we have that problem. In the meantime we should aim to keep everything as simple as possible.

tgeoghegan (approved):

PR #221, opened 2020-11-20
Misc. hardening improvements

yuriks:
Enables a few easy-win hardening settings in GKE. Shielded nodes were suggested by the Atredis report and the rest are other things I noticed along the way. See the individual commit messages for more info.

### DIFF
--- a/terraform/modules/gke/gke.tf
+++ b/terraform/modules/gke/gke.tf
@@ -28,6 +28,11 @@ resource "google_container_cluster" "cluster" {
   # https://www.terraform.io/docs/providers/google/r/container_cluster.html#remove_default_node_pool
   remove_default_node_pool = true
   initial_node_count       = 1
+
+  release_channel {
+    channel = "REGULAR"
+  }
+
   # We opt into a VPC native cluster because they have several benefits (see
   # https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips). Enabling
   # this networking_mode requires an ip_allocation_policy block and the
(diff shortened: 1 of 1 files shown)


tgeoghegan:
One question about the cluster update feature: our application can't function unless the Kubernetes API is available, because we need to schedule facilitator jobs to do work. So does that mean we'll have 20-30 minutes of downtime during cluster upgrades? Or does the upgrade roll out across the managed GKE cluster so that there's always >1 node that can handle k8s API requests?

tgeoghegan (approved):
No other questions!

yuriks:
For zonal clusters, you do get apiserver downtime while upgrading. I'm actually not sure how long this takes but I just triggered a manual cluster upgrade in one of my test clusters to verify that.

For regional clusters however, which we are using, GKE updates one apiserver (out of 3) at a time, and so there is no apiserver downtime for upgrades.

tgeoghegan:
That's great, thank you for checking on that.

yuriks:
Update on the update timing:

```
$ gcloud container operations list --filter="TYPE:UPGRADE_MASTER TARGET:prio-yks-pvtk8s-b-cluster"
NAME                              TYPE            LOCATION  TARGET                     STATUS_MESSAGE  STATUS  START_TIME                      END_TIME
operation-1605919819970-477916e2  UPGRADE_MASTER  us-west1  prio-yks-pvtk8s-b-cluster                  DONE    2020-11-21T00:50:19.970912839Z  2020-11-21T01:16:36.087060015Z
```

This was a regional upgrade and took about 26m. Since there's 3 zonal apiservers to upgrade, in order, I'd roughly estimate downtime for a zonal cluster to be about a 1/3rd of that, so on the order of ~10m.

PR #227, opened 2020-11-24
Remove deploy-tool and deploy-tool related strings

aaomidi:
We need to also do a new release to get the bits of the manifest-updater and deploy-operator onto dockerhub.

This PR is also potentially going to require us to change the required checks for merging.


terraform/main.tf
```diff
@@ -343,15 +343,6 @@ output "gke_kubeconfig" {
   value = "Run this command to update your kubectl config: gcloud container clusters get-credentials ${module.gke.cluster_name} --region ${var.gcp_region} --project ${var.gcp_project}"
 }
 
-output "specific_manifests" {
```
tgeoghegan, line 346: I suspect you can also delete some of the outputs in the `data_share_processor` module that are only used to construct this output

terraform/Makefile
```diff
@@ -92,7 +92,7 @@ apply: prep ## Have terraform do the things. This will cost money.
 		-input=false \
 		-refresh=true \
 		-var-file="$(VARS)"
-	echo "If that succeeded, run 'terraform output --json | (cd ../deploy-tool/ ; go run main.go )'"
+	echo "If that succeeded, navigate to deploy-operator, run 'make deploy; make install'. Then use 'kubectl apply -f crds.yml' on the file generated by terraform."
```
tgeoghegan, line 95: I'm concerned about the proliferation of tools we have to remember to run *after* `make apply`. jsha has been looking into using a Helm chart to deploy Prometheus; could we do something similar for the deploy-operator and the CRDs? If not, let's add these `make` and `kubectl` steps to the `make apply` target.

terraform/main.tf
```diff
@@ -343,15 +343,6 @@ output "gke_kubeconfig" {
...
-  value = { for v in module.data_share_processors : v.data_share_processor_name => {
-    kubernetes-namespace = v.kubernetes_namespace
-    certificate-fqdn     = v.certificate_fqdn
-    specific-manifest    = v.specific_manifest
-    }
-  }
-}
-
 output "use_test_pha_decryption_key" {
```
tgeoghegan, line 346: This value is consumed by `deploy-tool` and is used when setting up paired test envs, so that one of them uses predictable keys. Does the deploy operator have a similar feature?

tgeoghegan (changes requested):

tgeoghegan (changes requested):
Cross posting from slack:
How are the facilitator jobs going to use the rotated keys? Right now we create batch signing keys and packet decryption keys Kubernetes secrets in Terraform, and provide the names of the secrets to `workflow-manager` via the `bsk-secret-name` and `pdks-secret-name` parameters, which in turns plugs the relevant secrets into the environment of the `facilitator` jobs it spawns. However I don't think that'll work anymore. I think what's missing is that `workflow-manager` needs to use the Kubernetes API to
 - discover what the latest batch signing key k8s secret is
 - discover ALL available packet decryption key k8s secrets
...and then plug those into the spawned `facilitator` jobs' environment. Fortunately you've already decorated the keys appropriately with the `type` label, so it should be straightforward for any `workflow-manager` to find the right keys! I think that should be done as part of this PR, because we can't delete `deploy-tool` until this is an end-to-end functional replacement.

PR #230, opened 2020-11-25
Add Prometheus Helm chart and basic metrics.

jsha:
Note: This does not yet use a regional GCE Persistent Disk even though that was recommended in #194. I wanted to get something published that was simple and demonstrated actual metrics flowing through. Plumbing through a setting to use the regional disk may take longer and can be done in parallel.


terraform/main.tf
```diff
@@ -293,3 +293,19 @@ output "specific_manifests" {
...
+
+  set {
+    name  = "server.ingress.enabled"
+    value = "true"
+  }
+
+  # Values available for overriding are listed in https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/values.yaml
+  set {
+    name  = "server.ingress.hosts[0]"
+    value = "prometheus.prio.crud.net"
```
tgeoghegan, line 309: Should this value be interpolated from something per-environment?

tgeoghegan (changes requested):
If not a persistent disk, where are metrics stored right now? The filesystem of the container that Prometheus runs in?

Would you mind filing issues for follow on work that you're aware of like configuring durable storage, setting up Grafana, and finding a Rust Prometheus client to make it easier to hand this work off when we all come back from the Native American Heritage Day long weekend?

jsha:
> If not a persistent disk, where are metrics stored right now?

They're stored on a persistent disk (which is created by the Helm chart). It's just not regional (afaict).

> Would you mind filing issues for follow on work that you're aware of like configuring durable storage, setting up Grafana, and finding a Rust Prometheus client to make it easier to hand this work off when we all come back from the Native American Heritage Day long weekend?

Will do!

terraform/main.tf
jsha: Good catch. I removed this for now; we don't actually want to export Prometheus by default.

tgeoghegan (approved):

PR #243, opened 2020-11-30
workflow-manager: use batch ID in intake job name

### LINKED ISSUE #231: intake-batch Kubernetes job names must incorporate batch UUID
At the moment, intake-batch jobs have names like `i-<truncated aggregation ID>-<date>`, yielding job names like `i-kittens-seen-2020-11-25-01-23`. This causes collisions if multiple batches for a single aggregation arrive in the same minute. The job name needs to include at least some of the batch UUID to ensure uniqueness (or at least give us a better hope of uniqueness). The tricky thing is fitting a unique component as well as some human-readable information in the meager 63 characters that Kubernetes affords us!

tgeoghegan:
Intake Kubernetes jobs must incorporate the batch UUID name or
workflow-manager will not schedule jobs if more than one batch arrives
with the same timestamp.

Resolves #231

### DIFF
--- a/workflow-manager/main.go
+++ b/workflow-manager/main.go
@@ -483,14 +483,36 @@ func fmtTime(t time.Time) string {
 	return t.Format("2006/01/02/15/04")
 }
 
-// idForJobName generates a job name-safe string from an aggregationID.
+// jobNameForBatchPath generates a name for the Kubernetes job that will intake
+// the provided batch. The name will incorporate the aggregation ID, batch UUID
+// and batch timestamp while being a legal Kubernetes job name.
+func intakeJobNameForBatchPath(path *batchPath) string {
+	// Kubernetes job names must be valid DNS identifiers, which means they are
+	// limited to 63 characters in length and also what characters they may
+	// contain. Intake job names are like:
+	// i-<aggregation name fragment>-<batch UUID fragment>-<batch timestamp>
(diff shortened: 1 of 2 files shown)


workflow-manager/main.go
```diff
@@ -483,14 +483,36 @@ func fmtTime(t time.Time) string {
...
+	// The batch timestamp is 16 characters, and the 'i' and '-'es take up
+	// another 4, leaving 43. We take the '-'es out of the UUID and use half of
+	// it, hoping that this plus the date will provide enough entropy to avoid
+	// collisions. Half a UUID is 16 characters, leaving 27 for the aggregation
+	// ID fragment.
+	// For example, we might get:
+	// i-com-apple-EN-verylongnameth-0f0f0f0f0f0f0f0f-2006-01-02-15-04
+	return fmt.Sprintf("i-%s-%s-%s",
+		aggregationJobNameFragment(path.aggregationID, 27),
+		strings.ReplaceAll(path.ID, "-", "")[:16],
```
aaomidi, line 503: For a tiny bit more entropy, we could use the last 16 chars of the UUID which is "more random" than the first 16 chars, mainly since the first few bits of the UUID are static: https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)

aaomidi (approved):

PR #245, opened 2020-12-01
Enable gs pagination

### LINKED ISSUE #167: workflow-manager: implement cursoring for GS APIs

aaomidi:
Fixes #167

### DIFF
--- a/workflow-manager/main.go
+++ b/workflow-manager/main.go
@@ -421,24 +421,26 @@ func (b *bucket) listFilesGS(ctx context.Context) ([]string, error) {
 	var output []string
 	it := bkt.Objects(ctx, query)
 
+	// Use the paginated API to list bucket contents, as otherwise we would only get the first 1,000 objects in the bucket.
+	// As in the S3 case above, we get 100 results at a time to ensure we exercise the pagination case.
 	// https://cloud.google.com/storage/docs/json_api/v1/objects/list
-	// We can have up to: 1000
 	p := iterator.NewPager(it, 100, "")
-
+	var objects []*storage.ObjectAttrs
 	for {
-		var objects []*storage.ObjectAttrs
+		// NextPage will append to the objects slice
 		nextPageToken, err := p.NextPage(&objects)
 		if err != nil {
 			return nil, fmt.Errorf("storage.nextPage: %w", err)
 		}
 
-		for _, obj := range objects {
-			output = append(output, obj.Name)
-		}
-
 		if nextPageToken == "" {
+			// no more data
 			break
 		}
+	}
+
+	for _, obj := range objects {
+		output = append(output, obj.Name)
 	}
 
 	return output, nil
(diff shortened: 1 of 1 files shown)


aaomidi:
Testing with amir-lots-of-files on priobringup

workflow-manager/main.go
```diff
@@ -420,16 +420,27 @@ func (b *bucket) listFilesGS(ctx context.Context) ([]string, error) {
...
+		var objects []*storage.ObjectAttrs
+		nextPageToken, err := p.NextPage(&objects)
 		if err != nil {
-			return nil, fmt.Errorf("iterating on inputBucket objects from %q: %w", b.bucketName, err)
+			return nil, fmt.Errorf("storage.nextPage: %w", err)
+		}
+
+		for _, obj := range objects {
+			output = append(output, obj.Name)
+		}
```
tgeoghegan, line 437: [`Pager.NextPage`](https://godoc.org/google.golang.org/api/iterator#Pager.NextPage) is documented to append the next page's contents to the slice passed in. So I think you could save yourself this inner `for` loop and all the copying it does if you instead declare `var objects` outside the outer loop, and pass `objects` into `p.NextPage` until `nextPageToken == ""`.

workflow-manager/main.go
```diff
@@ -420,16 +420,27 @@ func (b *bucket) listFilesGS(ctx context.Context) ([]string, error) {
 	log.Printf("looking for ready batches in gs://%s as (ambient service account)", b.bucketName)
 	var output []string
 	it := bkt.Objects(ctx, query)
+
+	// https://cloud.google.com/storage/docs/json_api/v1/objects/list
+	// We can have up to: 1000
```
tgeoghegan, line 425: I appreciate the link to the API docs! However the comment is a bit cryptic. 1000 what? Also the value we passed below is 100, so they seem disconnected. I'm assuming you used 100 to mirror what we do in the S3 case, which is a good idea, but this comment should explain that. How about:

```
Use the paginated API to list bucket contents, as otherwise we would only get the first 1,000 objects in the bucket. As in the S3 case above, we get 100 results at a time to ensure we exercise the pagination case.
https://cloud.google.com/storage/docs/json_api/v1/objects/list
```

tgeoghegan (changes requested):

workflow-manager/main.go
aaomidi: That would work, the only concern I have is that the code needs to keep a LOT more objects in memory, but they're all small metadata anyway...

aaomidi: The iterator is a lazy iterator, so it can get rid of the objects it creates and just keep the names around, but I think that's just premature optimization at that point.

workflow-manager/main.go
```diff
@@ -420,16 +420,29 @@ func (b *bucket) listFilesGS(ctx context.Context) ([]string, error) {
...
+		// NextPage will append to the objects slice
+		nextPageToken, err := p.NextPage(&objects)
 		if err != nil {
-			return nil, fmt.Errorf("iterating on inputBucket objects from %q: %w", b.bucketName, err)
+			return nil, fmt.Errorf("storage.nextPage: %w", err)
 		}
-		output = append(output, attrs.Name)
+
+		if nextPageToken == "" {
+			// no more data
```
tgeoghegan, line 437: I appreciate this comment, I feel like any time an empty string is used as a sentinel it deserves a comment.

tgeoghegan (approved):

PR #247, opened 2020-12-01
workflow-manager: check aggregation inputs

### LINKED ISSUE #220: Ensure that all three inputs exist before scheduling aggregation job
To aggregate a batch, we need to ensure all three of ingestion batch, own validation and peer validation exist and are valid. `workflow-manager` should check this before scheduling an aggregation job.

[c.f. this TODO comment in workflow-manager::main.go](https://github.com/abetterinternet/prio-server/blob/1805f7193b6c84a267233a529434fb2c6db602e9/workflow-manager/main.go#L244)

tgeoghegan:
Make sure that all three of ingestion batch, own validation and peer
validation are present before including a batch in an aggregation job.

Resolves #220

### DIFF
--- a/workflow-manager/main.go
+++ b/workflow-manager/main.go
@@ -261,8 +261,24 @@ func main() {
 
 	log.Printf("found %d peer validations", len(peerValidationBatches))
 
-	// TODO: Check that both peer and own have something.
-	var aggregationBatches = peerValidationBatches
+	// Take the intersection of the sets of own validations and peer validations to get the list of
+	// batches we can aggregate.
+	// Go doesn't have sets, so we have to use a map[string]bool. We use the batch ID as the key to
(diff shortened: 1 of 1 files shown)


workflow-manager/main.go
```diff
@@ -261,8 +261,24 @@ func main() {
 
 	log.Printf("found %d peer validations", len(peerValidationBatches))
 
-	// TODO: Check that both peer and own have something.
-	var aggregationBatches = peerValidationBatches
+	// We can only aggregate a batch if we have our own validations and the peer's validations. To
+	// that end we construct a set of own validations, and check each peer validation against it. Go
```
aaomidi, line 265: nit: maybe make a mention that we're making an intersection of ownValidations and peerValidations to get the aggregationBatches as a tl;dr

aaomidi (approved):

PR #250, opened 2020-12-02
deal with non-singleton ingestors

### LINKED ISSUE #107: Ingestor servers are not necessarily singletons
The parameter exchange design had assumed that ingestion servers were or at least could pretend to be singletons, and hence use a single batch signing key (subject to key rotation) and a single AWS or GCP identity to communicate with all the data share processors. It turns out that Google will be operating one ingestion server per participating state. To enable that (sound, good) design choice, we need to extend the protocol so that data share processors can discover ingestion server batch keys and cloud credentials on a per-state basis.

tgeoghegan:
Google will have one ingestion server per locality, using distinct batch
signing keys and GCP service accounts. We now will check for the
ingestor global manifest, and if it cannot be found, will fall back to a
specific manifest at a path "<instance-name>-manifest.json", i.e.,
"us-co-g-enpa-manifest.json". If there is a global manifest but it is
not a valid manifest, we still error out.

resolves #107

### DIFF
--- a/terraform/modules/data_share_processor/data_share_processor.tf
+++ b/terraform/modules/data_share_processor/data_share_processor.tf
@@ -46,10 +46,6 @@ variable "ingestor_manifest_base_url" {
   type = string
 }
 
-variable "ingestor_gcp_service_account_email" {
-  type = string
-}
-
 variable "peer_share_processor_aws_account_id" {
   type = string
 }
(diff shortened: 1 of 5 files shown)


terraform/modules/data_share_processor/data_share_processor.tf
```diff
@@ -96,7 +92,44 @@ variable "kms_keyring" {
...
+# We need the ingestion server's manifest so that we can discover the GCP
+# service account it will use to upload ingestion batches. Some ingestors
+# (Apple) are singletons, and advertise a single global manifest which contains
+# the single GCP SA used by that server to send ingestion batches. Others
+# (Google) operate a distinct instance per locality, using a distinct GCP SA for
+# each. We first check for a global manifest, and fall back to a specific
+# manifest if it cannot be found. A Terraform data.http block would cause the
+# entire plan or apply to fail if the specified URL can't be loaded, so we use
+# this data.external block to check the HTTP status.
+data "external" "global_manifest_http_status" {
```
aaomidi, line 104: I love this

aaomidi (approved):

Issue #251, opened 2020-12-02
workflow-manager: default to not pushing any metrics


jsha:
Right now workflow-manager uses a reasonable default value for its pushgateway to send metrics to, which matches our deploy. However, we'd like workflow-manager to default to not pushing metrics, since other deploys may not do them the same way we do. We should add a flag to specify the pushgateway to workflow-manager, and if it's empty, don't push metrics.

PR #252, opened 2020-12-02
Add Prometheus metrics to facilitator

jsha:
This adds some very basic metrics - number of intake-batch and aggregate jobs started. The main intent is to set up the scaffolding for pushing metrics from facilitator. This adds a variable for the address of the pushgateway. If not address is provided (the default), no metrics are pushed. A similar scheme should be adopted for workflow-manager (#251).

This incidentally factors out some of the arguments to BatchIntaker and BatchAggregator to reduce nesting a bit.

### DIFF
--- a/facilitator/src/bin/facilitator.rs
+++ b/facilitator/src/bin/facilitator.rs
@@ -1,6 +1,7 @@
 use anyhow::{anyhow, Context, Result};
 use chrono::{prelude::Utc, NaiveDateTime};
 use clap::{App, Arg, ArgMatches, SubCommand};
+use log::info;
 use prio::encrypt::PrivateKey;
 use prometheus::{register_counter, register_counter_vec, Counter};
 use ring::signature::{
@@ -723,11 +724,16 @@ fn intake_batch(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
         batch_signing_key: batch_signing_key_from_arg(sub_matches)?,
     };
 
+    let batch_id: &str = sub_matches.value_of("batch-id").unwrap();
+    let batch_id: Uuid = Uuid::parse_str(batch_id).unwrap();
+
+    let date: &str = sub_matches.value_of("date").unwrap();
+    let date: NaiveDateTime = NaiveDateTime::parse_from_str(date, DATE_FORMAT).unwrap();
+
     let mut batch_intaker = BatchIntaker::new(
         &sub_matches.value_of("aggregation-id").unwrap(),
-        &Uuid::parse_str(sub_matches.value_of("batch-id").unwrap()).unwrap(),
-        &NaiveDateTime::parse_from_str(&sub_matches.value_of("date").unwrap(), DATE_FORMAT)
-            .unwrap(),
+        &batch_id,
+        &date,
         &mut intake_transport,
         &mut peer_validation_transport,
         &mut own_validation_transport,
@@ -741,7 +747,9 @@ fn intake_batch(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
     .unwrap();
     intake_started.inc();
 
+    info!("beginning intake");
     let result = batch_intaker.generate_validation_share();
+    info!("done");
 
     let intake_finished = register_counter_vec!(
         "intake_jobs_finished",
--- a/facilitator/src/bin/facilitator.rs
+++ b/facilitator/src/bin/facilitator.rs
@@ -324,7 +324,6 @@ impl<'a, 'b> AppArgumentAdder for App<'a, 'b> {
 }
 
 fn main() -> Result<(), anyhow::Error> {
-    use log::info;
     env_logger::init();
 
     let args: Vec<String> = std::env::args().collect();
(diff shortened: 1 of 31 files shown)


facilitator/src/bin/facilitator.rs
```diff
@@ -617,7 +616,28 @@ fn main() -> Result<(), anyhow::Error> {
         ("aggregate", Some(sub_matches)) => aggregate(sub_matches),
         ("lint-manifest", Some(sub_matches)) => lint_manifest(sub_matches),
         (_, _) => Ok(()),
+    };
+
+    // Once we've run the subcommand, whether it succeeds or fails, we want to push any metrics
+    // we have collected.
+    if let Some(pushgateway) = matches.value_of("pushgateway") {
+        if pushgateway != "" {
```
tgeoghegan, line 624: nit: [You can add a condition to a pattern match](https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html#extra-conditionals-with-match-guards) so that it'll match `Some` and check for the empty string. I'm not sure if it'll work in a `let` statement though.

facilitator/src/bin/facilitator.rs
```diff
@@ -709,17 +729,43 @@ fn intake_batch(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
         batch_signing_key: batch_signing_key_from_arg(sub_matches)?,
     };
 
+    let batch_id: &str = sub_matches.value_of("batch-id").unwrap();
+    let batch_id: Uuid = Uuid::parse_str(batch_id).unwrap();
```
tgeoghegan, line 733: nit: is the explicit `: Uuid` needed here? The compiler can't infer it from the return type of `Uuid::parse_str`? same goes for the `NaiveDateTime::parse_from_str` call just below.

facilitator/src/bin/facilitator.rs
```diff
@@ -709,17 +729,43 @@ fn intake_batch(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
...
         &mut intake_transport,
         &mut peer_validation_transport,
         &mut own_validation_transport,
         is_first_from_arg(sub_matches),
     )?;
-    batch_intaker.generate_validation_share()
+
+    let intake_started: Counter = register_counter!(
+        "intake_jobs_begun",
+        "Number of intake-batch jobs that started (on the facilitator side)"
```
tgeoghegan, line 750: nit: I think we should remove the `(on the facilitator side)` suffix from this and other counters. I think whether or not this was a facilitator or a PHA can be inferred from whether the whole env `is_first` or not.

tgeoghegan (approved):

facilitator/src/bin/facilitator.rs
jsha: I chose to be explicit here to emphasize the change in type for the shadowed variable - from `&str` to `Uuid`. Definitely open to other styles. For instance, we could name the second one `batch_uuid`.

facilitator/src/bin/facilitator.rs
jsha: Yeah, that was my first instinct, but I decided

```
if let Some(pushgateway) if !pushgateway.is_empty() = matches.value_of("pushgateway") {
```

was stylistically too confusing. Happy to go the other way on this though; what do you think?

facilitator/src/bin/facilitator.rs
jsha: In this case I am distinguishing the variable from `intake_jobs_started` on the workflow-manager side. In other words, we have one metric for "how many jobs did workflow-manager create," and "how many jobs did the facilitator actually start working on." They'll almost always be the same, but it seemed like a very simple first metric to add.

facilitator/src/bin/facilitator.rs
tgeoghegan: I prefer doing it in a single statement to avoid the extra indentation of the nested if statement, and I think statements that you have to squint at a bit to make sense of are table stakes when you write Rust.

facilitator/src/bin/facilitator.rs
tgeoghegan: You're right, I do like being explicit about shadowing the variable.

facilitator/src/bin/facilitator.rs
tgeoghegan: Oh, I understand now. I was thinking about "facilitator" as in "not a PHA", but you meant "facilitator" as in "not workflow-manager". This is fine as is.

facilitator/src/bin/facilitator.rs
jsha: Oh, funny, I actually looked this up the other day and found it's not yet implemented, then promptly forgot: https://github.com/rust-lang/rust/issues/51114

facilitator/src/bin/facilitator.rs
jsha: Oh, funny, I actually looked this up the other day and found it's not yet implemented, then promptly forgot: https://github.com/rust-lang/rust/issues/51114

So, guard statements (the "if !pushgateway.is_empty()") are not yet supported on `if let`.

tgeoghegan: Shame! in that case what you have is just fine.

Issue #151, opened 2020-11-04
Retry job creation in workflow manager as appropriate


tgeoghegan:
While doing some testing, I had a `workflow-manager` job fail with this log output:
```
2020/11/04 17:50:10 starting /workflow-manager version timg/robust-fake-ingestor+c156cd89 - Wed 04 Nov 2020 05:29:30 PM UTC. Args: [--is-first=false --k8s-namespace narnia --k8s-service-account narnia-ingestor-2-workflow-manager --ingestor-input s3://us-west-1/prio-hatfield-narnia-ingestor-2-ingestion --ingestor-identity arn:aws:iam::338276578713:role/prio-hatfield-narnia-ingestor-2-bucket-role --own-validation-input gs://prio-hatfield-narnia-ingestor-2-own-validation --peer-validation-input s3://us-west-1/prio-hatfield-narnia-ingestor-2-peer-validation --peer-validation-identity arn:aws:iam::338276578713:role/prio-hatfield-narnia-ingestor-2-bucket-role --bsk-secret-name hatfield-narnia-ingestor-2-batch-signing-key --pdks-secret-name hatfield-narnia-ingestion-packet-decryption-key --intake-batch-config-map narnia-ingestor-2-intake-batch-config --aggregate-config-map narnia-ingestor-2-aggregate-config --facilitator-image us.gcr.io/prio-atredis-oct-2020/prio-facilitator:latest]
2020/11/04 17:50:10 looking for ready batches in s3://prio-hatfield-narnia-ingestor-2-ingestion as arn:aws:iam::338276578713:role/prio-hatfield-narnia-ingestor-2-bucket-role
2020/11/04 17:50:14 fetched token from http://metadata.google.internal:80/computeMetadata/v1/instance/service-accounts/default/identity?audience=sts.amazonaws.com/338276578713
2020/11/04 17:50:14 ready: kittens-seen/2020/11/04/17/47/a023beed-a03c-4122-a063-f94f558e7110
2020/11/04 17:50:14 ready: kittens-seen/2020/11/04/17/48/8f0af102-caa2-4fcf-9650-80ab2b3e615b
2020/11/04 17:50:14 ready: kittens-seen/2020/11/04/17/49/a5065a9c-0fb6-4532-95a8-bd0e4c31cb46
2020/11/04 17:50:14 ready: kittens-seen/2020/11/04/17/43/d2e1d3d8-e2b3-4455-beb5-aedec8975285
[...]

PR #255, opened 2020-12-02
Restructure code to separate out different components

aaomidi:
Doing this pre #151 to make the retry-logic code easier to write and reason about with

### DIFF
--- a/workflow-manager/batchpath/batchpath.go
+++ b/workflow-manager/batchpath/batchpath.go
@@ -1,0 +1,119 @@
+package batchpath
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/letsencrypt/prio-server/workflow-manager/utils"
+)
+
+type BatchPath struct {
+	AggregationID  string
+	dateComponents []string
+	ID             string
+	Time           time.Time
+	metadata       bool
+	avro           bool
+	sig            bool
+}
+
+type List []*BatchPath
+
+func (bpl List) Len() int {
+	return len(bpl)
+}
+
+func (bpl List) Less(i, j int) bool {
+	return bpl[i].Time.Before(bpl[j].Time)
+}
+
+func (bpl List) Swap(i, j int) {
+	bpl[i], bpl[j] = bpl[j], bpl[i]
+}
+
+func New(batchName string) (*BatchPath, error) {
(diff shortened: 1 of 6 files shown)


workflow-manager/batchpath/batchpath.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (bpl List) Less(i, j int) bool {
+	return bpl[i].Time.Before(bpl[j].Time)
+}
+
+func (bpl List) Swap(i, j int) {
+	bpl[i], bpl[j] = bpl[j], bpl[i]
+}
+
+func New(batchName string) (*BatchPath, error) {
```
tgeoghegan, line 38: nit: exported methods should have doc strings. I know the original code lacked substantial comments or docs, but since you're going to the trouble of cleaning this up, now might be the time.

workflow-manager/bucket/bucket.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
+	aws_session "github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/sts"
+	"github.com/letsencrypt/prio-server/workflow-manager/tokenfetcher"
+	"google.golang.org/api/iterator"
+)
+
+type bucket struct {
```
tgeoghegan, line 21: nit: Should this be `Bucket` so it's visible outside of this module?

tgeoghegan (approved):

workflow-manager/bucket/bucket.go
aaomidi: I didn't see anything that needs to see the actual bucket struct outside of the bucket module, so I went with the principle of data hiding.

I think if we end up needing the Bucket outside of this module then we can make it visible?

workflow-manager/batchpath/batchpath.go
aaomidi: Yep! I should add those comments. Thanks for reminding me!

workflow-manager/bucket/bucket.go
aaomidi: Eh actually since we're using the Bucket struct outside it feels odd to keep it unexported. I'll make it exported.

PR #260, opened 2020-12-03
facilitator: retry AWS S3 requests

tgeoghegan:
During Narnia testing, we encountered cases where we failed to upload
validation batches to S3 because the ingestion batches were _just_ big
enough so that roughly ten seconds would elapse between the
CreateMultipartUpload call and the first UploadPart. We had previously
set the Hyper connection pool idle timeout to 10 seconds to avoid having
S3 close connections on us after 20 seconds, so sometimes we would lose
the race with Hyper and have uploads spuriously fail. Our fix is to
detect the case where the txn fails due to the connection getting closed
(which manifests as RusotoError::HttpDispatch) and try up to two more
times in that case. We don't retry in the face of other errors to avoid
abusing Amazon API endpoints.

Example of the error encountered:

HttpDispatch(HttpDispatchError { message: "Error during dispatch: connection closed before message completed" })

### DIFF
--- a/facilitator/src/transport/s3.rs
+++ b/facilitator/src/transport/s3.rs
@@ -13,7 +13,7 @@ use rusoto_core::{
         AutoRefreshingProvider, AwsCredentials, ContainerProvider, CredentialsError,
         InstanceMetadataProvider, ProfileProvider, ProvideAwsCredentials, Secret, Variable,
     },
-    ByteStream, Region,
+    ByteStream, Region, RusotoError, RusotoResult,
 };
 use rusoto_s3::{
     AbortMultipartUploadRequest, CompleteMultipartUploadRequest, CompletedMultipartUpload,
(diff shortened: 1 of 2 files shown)


facilitator/src/transport/s3.rs
```diff
@@ -192,13 +195,33 @@ impl Transport for S3Transport {
         info!("get {}/{} as {:?}", self.path, key, self.iam_role);
         let mut runtime = basic_runtime()?;
         let client = (self.client_provider)(&self.path.region, self.iam_role.clone())?;
-        let get_output = runtime
-            .block_on(client.get_object(GetObjectRequest {
+
+        let mut retries = 0;
+        let get_output = loop {
+            match runtime.block_on(client.get_object(GetObjectRequest {
```
jsha, line 204: Style-wise, I think it's really valuale to avoid multi-line inputs to match statements. I recommend assigning the output of `runtime.block_on` to a variable, and then doing `match result {`

facilitator/src/transport/s3.rs
```diff
@@ -192,13 +198,33 @@ impl Transport for S3Transport {
...
                 bucket: self.path.bucket.to_owned(),
                 key: [&self.path.key, key].concat(),
                 ..Default::default()
-            }))
-            .context("error getting S3 object")?;
+            })) {
+                Ok(get_output) => break Ok(get_output),
```
jsha, line 209: Style nit: The `Ok(get_output)` and `Err(err)` branches of this match can consolidated into a single catch-all `result => result` branch that comes after the `Err(RusotoError..)` one.

facilitator/src/transport/s3.rs
```diff
@@ -192,13 +198,33 @@ impl Transport for S3Transport {
...
+                Err(RusotoError::HttpDispatch(err)) => {
+                    retries += 1;
+                    if retries == ATTEMPT_COUNT {
```
jsha, line 212:
```suggestion
                    if retries >= ATTEMPT_COUNT {
```

Always good to be paranoid just in case you somehow increment twice.

facilitator/src/transport/s3.rs
```diff
@@ -394,20 +456,41 @@ impl TransportWriter for MultipartUploadWriter {
...
-                        bucket: self.bucket.to_string(),
-                        key: self.key.to_string(),
-                        upload_id: self.upload_id.clone(),
-                        multipart_upload: Some(CompletedMultipartUpload {
-                            parts: Some(mem::take(&mut self.completed_parts)),
-                        }),
-                        ..Default::default()
+        let mut retries = 0;
+        let completed_parts = mem::take(&mut self.completed_parts);
+        loop {
```
jsha, line 461: The duplication of these retry loops strongly suggests factoring out a "do_with_retries" function that takes a closure, and runs it in the loop specified below.

jsha (changes requested):

facilitator/src/transport/s3.rs
tgeoghegan: Extremely cool, I did not know you could do that!

facilitator/src/transport/s3.rs
tgeoghegan: This was less painful than I expected it to be and it's much better this way.

jsha (approved):

PR #262, opened 2020-12-03
workflow-manager shouldn't use prometheus by default

### LINKED ISSUE #251: workflow-manager: default to not pushing any metrics

aaomidi:
Fixes #251


tgeoghegan (approved):
I like the NoopCounter! Before merging, I suggest improving the commit message and PR description to include some more context on what this change does and why.

jsha:
One thing to note: The approach I took in https://github.com/abetterinternet/prio-server/pull/252 was to to add a --pushgateway flag. If it's empty, we don't push prometheus metrics; if it exists, we push them to the address specified.

I don't have a strong opinion about which approach to take, but it would be valuable to have them both be the same. For instance, if someone needs to use a different address for the pushgateway, and they can do that in facilitator but not in workflow-manager, that would be annoying.

Also: The only thing you really need to turn off in non-prometheus mode is `push.New(pushGateway, "workflow-manager").Gatherer(prometheus.DefaultGatherer).Push()`. It's totally legit and fine to create and increment Prometheus counters even if you're not going to be delivering them. The downside is that you don't get to use NoopCounters for testing that counters get incremented when you expect.

aaomidi:
Oh interesting. I didn't really know how the library works @jsha. I'll tackle this in a bit of a cleaner way.

I'm not sure if the NoopCounter is going to be super useful for us, but we may want to use another counter implementation in our integration testing to verify the operations took place successfully at the end. So I think I'll keep the idea of using these counters around.

jsha:
FWIW, you can write to Prometheus counters and then extract their values for testing, but it's quite annoying to do. See these helps in Boulder: https://github.com/letsencrypt/boulder/blob/main/test/test-tools.go#L160-L215. It's also possible the story has gotten easier in the time since we wrote those.

PR #267, opened 2020-12-04
deploy-tool: only create manifests if they don't already exist

aaomidi:


deploy-tool/main.go
```diff
@@ -185,7 +190,20 @@ func generateAndDeployKeyPair(namespace, keyName string, keyMarshaler privateKey
 	return p256Key, nil
 }
 
+func manifestExists(fqdn, dsp string) bool {
+	// Remove the locality name from the FQDN
+	path := fmt.Sprintf("https://%s/%s-manifest.json", fqdn, dsp)
+
+	resp, err := http.Get(path)
+	if err != nil {
+		log.Printf("error when getting manifest %s: %s", path, err)
```
tgeoghegan, line 199: I don't think we need to log here since this isn't really an error condition. Could this just `return false`?

terraform/main.tf
```diff
@@ -396,6 +396,10 @@ output "specific_manifests" {
   }
 }
 
+output "own_manifest_url" {
```
tgeoghegan, line 399: nit: `own_manifest_base_url`, to connote that this isn't the full URL to a manifest but rather the URL relative to which manifests may be found.

tgeoghegan (approved):
Please change the log in the error case in `manifestExists`, then go ahead and merge.

tgeoghegan:
Looks like there are also some linter issues that should be addressed.

deploy-tool/main.go
aaomidi: Hmm, non 200s don't return an error. This error means something actually broke. I'll actually convert it to a fatalf since the operator probably doesn't want it to continue messing up the environment.

PR #268, opened 2020-12-04
Terraform hotfixes

tgeoghegan:
While bringing up new environments for integration testing, I encountered a couple of bugs in Terraform which this commit fixes.

### DIFF
--- a/terraform/modules/manifest/manifest.tf
+++ b/terraform/modules/manifest/manifest.tf
@@ -132,5 +132,6 @@ output "bucket" {
 }
 
 output "base_url" {
-  value = local.domain_name
+  # local.domain_name is a fully qualified DNS name, ending in '.'
+  value = trimsuffix(local.domain_name, ".")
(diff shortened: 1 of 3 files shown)


terraform/modules/manifest/manifest.tf
```diff
@@ -132,5 +132,6 @@ output "bucket" {
 }
 
 output "base_url" {
-  value = local.domain_name
+  # local.domain_name is a fully qualified DNS name, ending in '.'
+  value = trimsuffix(local.domain_name, ".")
```
aaomidi, line 136: It might even make sense to add the https scheme here and then update

https://github.com/abetterinternet/prio-server/blob/main/deploy-tool/main.go#L92
and
https://github.com/abetterinternet/prio-server/blob/main/deploy-tool/main.go#L196

But what this is so far looks good too!

aaomidi (approved):

PR #269, opened 2020-12-06
s3: try WebIdentityProvider before instance metadata

tgeoghegan:
We observed that every single S3 transaction being done by NCI's
facilitator containers was taking at least 30 seconds, which led me to
suspect a timeout somewhere. I think I was able to reproduce the problem
by moving my ~/.aws/ directory aside and running facilitator locally.
Sure enough, it takes 30 seconds for Rusoto to give up on connecting to
169.254.169.254 (the AWS instance metadata endpoint). The crude fix here
is to check the web identity provider before instance metadata.

### DIFF
--- a/facilitator/src/transport/s3.rs
+++ b/facilitator/src/transport/s3.rs
@@ -920,10 +920,10 @@ async fn chain_provider_credentials(
     if let Ok(creds) = provider.container_provider.credentials().await {
         return Ok(creds);
     }
-    if let Ok(creds) = provider.instance_metadata_provider.credentials().await {
+    if let Ok(creds) = provider.webidp_provider.credentials().await {
         return Ok(creds);
     }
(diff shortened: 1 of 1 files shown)


jsha (approved):
Great quick fix. Makes me wonder: Does the equivalent code in the Go AWS SDK time out earlier, or do some other preliminary check, or what?

tgeoghegan:
On the Go/`workflow-manager` side, [we assume that the presence of an `own`, `peer`, etc. `-identity` flag means we want to use Web Identity and directly use a `WebIdentityRoleProviderWithToken`](https://github.com/abetterinternet/prio-server/blob/3a666e6213e061c3250d8019e3ad885d87391ba4/workflow-manager/bucket/bucket.go#L92), short circuiting any timeout trying to access the metadata service.

Issue #14, opened 2020-09-21
Metrics and alerting


tgeoghegan:
Figure out what kind of metrics we want to emit from the server, what conditions to alert on, and where alerts should go. For instance, the facilitator could encounter failures because of bad data emitted by an ingestion server, so perhaps we should figure out how to route such alerts to the other organisation.

tgeoghegan:
We have some basic metrics and alerts in place. Any further work will be tracked in more specific tickets or PRs.

PR #270, opened 2020-12-06
facilitator: improve aggregation job logs

tgeoghegan:
Have the aggregation jobs log rejected packets, and also log what inputs
and outputs they are using, analogously to some very helpful logging in
the intake-batch job.

### DIFF
--- a/facilitator/src/aggregation.rs
+++ b/facilitator/src/aggregation.rs
@@ -9,6 +9,7 @@ use crate::{
 };
 use anyhow::{anyhow, Context, Result};
 use chrono::NaiveDateTime;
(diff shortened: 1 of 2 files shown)


jsha (approved):
Looks great. I'll just note that this would be a great thing to add a metric for (as a separate PR). Happy to give description on how, if needed.

tgeoghegan:
Agreed on metrics, I'll mention #14 here so we don't lose track of that observation. I want to include this in the version of `facilitator` we release to NCI later today, but hope to spend lots of time in the coming week or two adding new metrics like this.

PR #273, opened 2020-12-07
terraform: pin prod-us to 0.5.0 images

tgeoghegan:
We pin the prod-us deployment of the application to version 0.5.0 of
`prio-workflow-manager` and `prio-facilitator`. This commit also plumbs
arguments for container image versions and image repository all the way
from kubernetes.tf to main.tf.

### DIFF
--- a/terraform/modules/kubernetes/kubernetes.tf
+++ b/terraform/modules/kubernetes/kubernetes.tf
@@ -11,8 +11,7 @@ variable "environment" {
 }
 
 variable "container_registry" {
-  type    = string
-  default = "letsencrypt"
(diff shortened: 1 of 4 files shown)


terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -10,19 +10,13 @@ variable "environment" {
   type = string
 }
 
-variable "container_registry" {
```
aaomidi, line 13: Seems like this was being referred to in the module

aaomidi (approved):

PR #275, opened 2020-12-07
Simplify call to generate_ingestion_sample.

jsha:
clap has a macro value_t that lets us skip much of the boilerplate:
https://docs.rs/clap/2.33.3/clap/macro.value_t.html


tgeoghegan (approved):
Great find!

PR #279, opened 2020-12-11
attempt to change aws account id to string

hostirosti:


facilitator/src/manifest.rs
```diff
@@ -58,7 +58,7 @@ pub struct DataShareProcessorGlobalManifest {
 pub struct DataShareProcessorServerIdentity {
     /// The numeric account ID of the AWS account this data share processor will
     /// use to access peer cloud resources.
-    aws_account_id: u64,
+    aws_account_id: String,
```
tgeoghegan, line 61: Per discussion in Slack, this value isn't always needed, so I think the thing to do would be `Option<String>`, which would enable NCI to not include this value at all. That means the test cases below would need `Some("12345678901234567")`. We should then also bump the expected `format` value to 1, to reflect the change in format since Narnia.

tgeoghegan (changes requested):

facilitator/src/manifest.rs
hostirosti: made some changes according to your commit. Not sure I made the right ones.

jsha:
While we're on the subject, @tgeoghegan do you think there's any benefit from making this field contain an ARN instead of an account ID?

hostirosti:
@jsha ARN wouldn't work here, we need the account_id for the identity federation that is used to write to NCI buckets.

tgeoghegan:
> While we're on the subject, @tgeoghegan do you think there's any benefit from making this field contain an ARN instead of an account ID?

I'm inferring that you propose having ISRG create a single AWS role for writing to NCI's peer validation buckets, and have all data share processors assume that role (right now we create a distinct AWS role for each data share processor). I'm not sure if that's better from NCI's point of view: they have no insight into what policies are associated with a role in a remote account so it's tantamount to granting access to the entire account from their point of view.

I'm also extremely reluctant to make any changes to manifest formats as the coordinated deploys and test cycles are very tedious and painful.

PR #285, opened 2020-12-17
Terraform: metrics and alerting

tgeoghegan:
- Configures workflow-manager and facilitator to emit metrics to
   Prometheus in prod-us
 - Adds Grafana to the cluster and configures Prometheus as a data
   source
 - Configures Prometheus to send alerts to VictorOps
 - Documents how to add dashboards to Grafana and how to access
   Prometheus and Grafana cluster private endpoints using `kubectl
   port-forward`.


terraform/main.tf
```diff
@@ -431,8 +431,81 @@ provider "helm" {
   }
 }
 
+# Prometheus alertmanager must be hooked up to VictorOps to deliver alerts,
```
aaomidi, line 434: Could the monitoring stuff live inside another module? I'm trying to cleanup `main.tf` to basically only invoke other modules.

terraform/main.tf
```diff
@@ -431,8 +431,81 @@ provider "helm" {
...
+# which requires an API key. We don't want to check that key into git, so
+# similarly to the batch signing and packet decryption keys, we make a dummy
+# k8s secret here and fill in a real value later. See terraform/README.md.
+resource "kubernetes_secret" "prometheus_alertmanager_config" {
+  metadata {
+    name = "prometheus-alertmanager-config"
+    # Prometheus runs in the default namespace
+    namespace = "default"
```
aaomidi, line 442: Hmm, does it have to run in the default namespace? Would making a monitoring namespace make sense here?

aaomidi (approved):

tgeoghegan:
Good call! It was quite easy to move things into a separate TF module and k8s namespace for `monitoring`.

PR #286, opened 2020-12-18
Integration tester skeleton

aaomidi:
TODO:

- [x] Add better logging
- [x] Add CI/CD pipelines (doing this in another PR)
- [x] Readme

Integration tester skeleton will replace the facilitator code that we're running on terraform to generate samples. This code is responsible for finding all the various environment information by reading  a manifest file, and then generate the suitable samples for it.

#241 depends on this PR.


facilitator/src/bin/facilitator.rs
```diff
@@ -1031,7 +1031,8 @@ fn batch_signing_key_from_arg(matches: &ArgMatches) -> Result<BatchSigningKey> {
         .value_of("batch-signing-private-key-identifier")
         .unwrap();
     Ok(BatchSigningKey {
-        key: EcdsaKeyPair::from_pkcs8(&ECDSA_P256_SHA256_ASN1_SIGNING, &key_bytes)?,
+        key: EcdsaKeyPair::from_pkcs8(&ECDSA_P256_SHA256_ASN1_SIGNING, &key_bytes)
```
aaomidi, line 1034: I can probably do this in another PR if we don't want to take it here

tgeoghegan: No, this a much better error message, and is useful to have here. However I would just do `.context("failed to parse batch signing key")?`. There's no need for a closure since you're not capturing any variables here.

integration-tester/cmd/root.go
```diff
@@ -0,0 +1,81 @@
...
+	rootCmd.PersistentFlags().StringVarP(&serviceAccountName, "service-account-name", "", "", "The service account name for kubernetes")
+
+	rootCmd.PersistentFlags().StringVarP(&facilitatorImage, "facilitator-image", "", "", "The image for the facilitator")
+	rootCmd.PersistentFlags().StringVarP(&pushGateway, "push-gateway", "", "", "The push gateway")
+	rootCmd.PersistentFlags().StringVarP(&peerIdentity, "peer-identity", "", "", "The peer identity")
+	rootCmd.PersistentFlags().StringVarP(&awsAccountId, "aws-account-id", "", "", "The aws account ID")
+}
+
+func Execute() {
+	// Allow running from explorer
```
tgeoghegan, line 68: explorer? Like Windows explorer?

integration-tester/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,77 @@
+package kubernetes
```
tgeoghegan, line 1: I added a Kubernetes module to `workflow-manager` in #292 that shares a little bit of functionality with this file, namely the bits that get a client. It'd be nice to share modules between different go targets.

integration-tester/tester/logic.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return err
+	}
+
+	job := t.createJob(manifest, bsk, pdk)
+
+	return kubernetes.ScheduleJob(t.namespace, job)
+}
+
+func (t *Tester) createJob(manifest *m.DataShareSpecificManifest, bsk, pdk *corev1.Secret) *batchv1.Job {
```
tgeoghegan, line 36: You'll need to introduce something that deletes these jobs when they're completed, or we'll run into the same problems we did with intake jobs where we have too many completed jobs lying around.

integration-tester/tester/logic.go
```diff
@@ -0,0 +1,157 @@
...
+	defer r.Body.Close()
+
+	err = json.NewDecoder(r.Body).Decode(dsm)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode body %s: %v", r.Body, err)
+	}
+	return dsm, err
+}
+
+func (t *Tester) getValidPDK(manifest *m.DataShareSpecificManifest) (*corev1.Secret, error) {
```
tgeoghegan, line 115: I think go style frowns on abbreviations like this, so `getValidPacketDecryptionKey` would be better. Also what makes a key valid? From the implementation, it looks like that means whatever key is currently advertised in a manifest, but it'd be good to explain that in a doccomment.

integration-tester/tester/logic.go
```diff
@@ -0,0 +1,157 @@
...
+		val, ok := secretMap[name]
+
+		if ok {
+			return &val, nil
+		}
+	}
+	return nil, fmt.Errorf("unable to find a suitable packet decryption key - manifest was: %s", manifest.PacketEncryptionKeyCSRs)
+}
+
+func (t *Tester) getValidBSK(manifest *m.DataShareSpecificManifest) (*corev1.Secret, error) {
```
tgeoghegan, line 132: Similar comment to `getValidPDK`: expand the acronym and document the basis on which a key is deemed valid

tgeoghegan (changes requested):
Can you help me understand what the goal of this PR is? It looks like what's achieved is to build an image that can replace the `sample-maker` cronjobs and doesn't need to use fixed packet encryption keys. Is that accurate? If so, then do you want to include the Terraform changes to replace the `sample-makers` with this in this PR or in a subsequent one? Basically I'd love to have a sense of what your broader plan is so I know how this PR fits in with that vision.

integration-tester/cmd/root.go
aaomidi: At this point I have no idea what I meant by this comment 😅

integration-tester/kubernetes/kubernetes.go
aaomidi: I'll wait for that to get merged in and pull that module in here

integration-tester/tester/logic.go
aaomidi: Yep!

integration-tester/cmd/root.go
aaomidi: Ah this is just some code I share between projects. Irrelevant to this one.

aaomidi:
> Is that accurate? If so, then do you want to include the Terraform changes to replace the sample-makers with this in this PR or in a subsequent one? Basically I'd love to have a sense of what your broader plan is so I know how this PR fits in with that vision.

Yep that's accurate.

I'm looking to make this PR not change anything in terraform yet. I'm planning to update terraform either in #241 or in its own PR (I'll probably make it in its own PR, #241 is a major change).

The work in #241 showed that an integration-tester is necessary since terraform is eventually not going to be responsible for making the various testing keys and we want to test with actual manifest files and the actual deploy-operator work.

integration-tester/kubernetes/kubernetes.go
aaomidi: Yep it's fine! Just when building in docker the entire context of the git repo needs to be sent to the build agent, rather than this project alone.

tgeoghegan (changes requested):
My big concerns here are the keys that the ingestion batch maker will use. It should not be using the data share processor's batch signing key (ingestors have their own batch signing key), and it should be discovering the packet encryption keys via the data share processor specific manifests instead of peering into Kubernetes secrets or using hard coded keys.

integration-tester/tester/types.go
```diff
@@ -0,0 +1,47 @@
+package tester
```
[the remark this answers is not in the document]
aaomidi: `root.go` and `run.go` is just how cobra & viper recommend their project layouts. `main.go` just needs to exist to bootstrap the application.

I don't necessarily agree that putting everything in `main.go` makes it more straightforward.

Right now, for example, if we find out we would rather pass something as a config value, or environment variable into workflow-manager there would be a lot of refactoring that would need to happen. There's also dozens of lines of `log.FatalF` for error management that we can skip over with this layout. These extra steps just make it really easy to follow the structure of the application once familiarity with viper & cobra is established.

`root` being the command registry.
`run` being the one specific command.

I do agree with logic.go and types.go though. That's just personal preference of keeping basic type definitions outside of the core logic of the module. But I can consolidate them into a single file and put them under `run.go`.

integration-tester/tester/logic.go
```diff
@@ -0,0 +1,255 @@
...
+				Spec: corev1.PodSpec{
+					RestartPolicy:                "Never",
+					ServiceAccountName:           t.serviceAccountName,
+					AutomountServiceAccountToken: &trueP,
+					Containers: []corev1.Container{
+						{
+							Name:  "integration-tester",
+							Image: t.facilitatorImage,
+							Env:   env,
+							Args: []string{
```
aaomidi, line 134: Let's just double check to make sure this makes sense

integration-tester/tester/logic.go
```diff
@@ -0,0 +1,255 @@
...
+			return "", fmt.Errorf("error when parsing the packet encryption csr: %v", err)
+		}
+
+		return publicKey, nil
+	}
+	return "", fmt.Errorf("no packet encryption CSRs in manifest: %v", manifest)
+}
+
+func getBase64PublicKeyFromCSR(pemCsr string) (string, error) {
+	pemBlock, rest := pem.Decode([]byte(pemCsr))
```
aaomidi, line 207: I'll write a test for this piece

integration-tester/tester/logic.go
```diff
@@ -0,0 +1,255 @@
...
+	phaManifest, err := ReadDataShareProcessorSpecificManifest(t.phaManifestUrl)
+	if err != nil {
+		return err
+	}
+	facilManifest, err := ReadDataShareProcessorSpecificManifest(t.facilManifestUrl)
+	if err != nil {
+		return err
+	}
+
+	bsk, err := t.getValidBatchSigningKey(ownManifest)
```
tgeoghegan, line 40: nit: expand this to `batchSigningKey`

integration-tester/tester/logic.go
tgeoghegan: Based on what's currently in `kubernetes.tf`, I think you have `own` and `peer` reversed. It's possible that `facilitator generate-ingestion-sample` is just wrong in its treatment of `own` and `peer` vs. facilitator and PHA.

I'm also not sure about the identity arguments, because I'm not sure where `facilManifest.IngestionIdentity` or `phaManifest.IngestionIdentity` come from, or what their values will be. But consider carefully whether they need values at all, and what their values should be based on whether the bucket being written to is a GCS or S3 bucket. The scheme for all these buckets and identities is documented [here](https://docs.google.com/document/d/1MdfM3QT63ISU70l63bwzTrxr93Z7Tv7EDjLfammzo6Q/edit#heading=h.m8nz1l5m6znk).

integration-tester/tester/logic.go
```diff
@@ -0,0 +1,255 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+// ReadDataShareProcessorSpecificManifest retrieves a manifest.DataShareProcessorSpecificManifest from the given url
+func ReadDataShareProcessorSpecificManifest(url string) (manifest.DataShareProcessorSpecificManifest, error) {
```
tgeoghegan, line 159: Are these functions ever used outside this module? Do they need to be public?

integration-tester/tester/logic.go
```diff
@@ -0,0 +1,255 @@
...
+
+	if !ok {
+		return "", fmt.Errorf("certificate request public key was not an ecdsa.PublicKey")
+	}
+	publicKeyBytes := elliptic.Marshal(ecdsaPublicKey.Curve, ecdsaPublicKey.X, ecdsaPublicKey.Y)
+
+	return base64.StdEncoding.EncodeToString(publicKeyBytes), nil
+}
+
+func (t *Tester) getValidBatchSigningKey(manifest manifest.IngestorGlobalManifest) (*corev1.Secret, error) {
```
tgeoghegan, line 230: Suggest renaming this type to `IngestorGlobal` so that then it reads as `manifest.IngestorGlobal`.

tgeoghegan (changes requested):

facilitator/src/sample.rs
```diff
@@ -133,7 +129,10 @@ pub fn generate_ingestion_sample(
                             .encode_simple(&data)
                             .context("failed to encode data")?;
 
-                        let r_pit = fake_server.choose_eval_at();
+                        // Hardcoded r_pit value
+                        // Some other options: 81133978, 998314904, 1103857271
+                        // Value obtained by running an instance of libprio::Server
```
tgeoghegan, line 134: Rather than list other magic numbers, this comment should concisely explain why we're using a constant. "running an instance of libprio::Server" is kind of vague, too.

facilitator/src/bin/facilitator.rs
```diff
@@ -455,8 +455,8 @@ fn main() -> Result<(), anyhow::Error> {
                         .required(true),
                 )
                 .arg(
-                    Arg::with_name("facilitator-ecies-private-key")
-                        .long("facilitator-ecies-private-key")
+                    Arg::with_name("facilitator-ecies-public-key")
+                        .long("facilitator-ecies-public-key")
                         .env("FACILITATOR_ECIES_PRIVATE_KEY")
```
tgeoghegan, line 460: The environment variable names should match the parameter names.

aaomidi: Oh whoops

facilitator/src/bin/facilitator.rs
```diff
@@ -2,7 +2,7 @@ use anyhow::{anyhow, Context, Result};
 use chrono::{prelude::Utc, NaiveDateTime};
```
tgeoghegan, line 2: These changes to `facilitator` make sense, but they will break existing paired test envs until the corresponding Terraform changes land

PR #292, opened 2020-12-28
publisher/subscriber task model

tgeoghegan:
This pull request implements a publisher/subscriber task model as we discussed on December 23rd. Note that I use the word _task_ to disambiguate from Kubernetes _jobs_.

 - `workflow-manager` is still a cronjob and still scans buckets for work to do, but now schedules tasks by publishing to PubSub topics. We use distinct topics for intake and aggregation.
 - `facilitator` has two new entry points, `intake-batch-worker` and `aggregate-worker`, that poll PubSub subscriptions in a loop for tasks to execute.
 - On the Terraform side, we now create a PubSub topic and subscription for intake and aggregation for each ingestor/locality pair. Correspondingly, we create a Kubernetes deployment for each subscription that dequeues tasks.
 - The deployments run a number of persistent replicas instead of having many, many individual jobs. The deployments can be scaled independently from each other by tuning the desired replica count.
 - We now have a staging environment, composed of the `staging-pha` and `staging-facil` environments for which I added .tfvars files. I'd like future releases of the application to bake in staging for some interval of time (at least long enough to run an aggregation) but I have no concrete plans around release flows or testing plans at the moment.
- Experimental support for task queues based on publishing to an SNS topic and pulling messages from an SQS queue subscribed to that topic. These implementations compile but have not been tested.

### DIFF
--- a/facilitator/src/bin/facilitator.rs
+++ b/facilitator/src/bin/facilitator.rs
@@ -13,13 +13,14 @@ use uuid::Uuid;
 
 use facilitator::{
     aggregation::BatchAggregator,
-    config::{Identity, ManifestKind, StoragePath},
+    config::{Identity, ManifestKind, StoragePath, TaskQueueKind},
     intake::BatchIntaker,
     manifest::{
         DataShareProcessorGlobalManifest, IngestionServerManifest, PortalServerGlobalManifest,
         SpecificManifest,
     },
     sample::{generate_ingestion_sample, SampleOutput},
+    task::{AggregationTask, AwsSqsTaskQueue, GcpPubSubTaskQueue, IntakeBatchTask, TaskQueue},
     transport::{
         GCSTransport, LocalFileTransport, S3Transport, SignableTransport, Transport,
         VerifiableAndDecryptableTransport, VerifiableTransport,
@@ -37,14 +38,6 @@ fn date_validator(s: String) -> Result<(), String> {
     NaiveDateTime::parse_from_str(&s, DATE_FORMAT)
         .map(|_| ())
         .map_err(|e| format!("{} {}", s, e.to_string()))
-}
-
-fn b64_validator(s: String) -> Result<(), String> {
-    if s == "not-a-real-key" {
-        return Err("'not-a-real-key'. Run deploy-tool to generate secrets".to_string());
-    }
-    base64::decode(s).map_err(|e| format!("decoding base64: {}", e))?;
-    Ok(())
 }
 
 fn uuid_validator(s: String) -> Result<(), String> {
@@ -75,6 +68,8 @@ trait AppArgumentAdder {
     fn add_packet_decryption_key_argument(self: Self) -> Self;
 
     fn add_gcp_service_account_key_file_argument(self: Self) -> Self;
+
+    fn add_task_queue_arguments(self: Self) -> Self;
 }
 
 const SHARED_HELP: &str = "Storage arguments: Any flag ending in -input or -output can take an \
@@ -242,8 +237,7 @@ impl<'a, 'b> AppArgumentAdder for App<'a, 'b> {
                 .help(leak_string(format!(
                     "Batch signing public key for the {}",
                     entity.str()
-                )))
-                .validator(b64_validator),
+                ))),
         )
         .arg(
             Arg::with_name(entity.suffix("-public-key-identifier"))
@@ -268,8 +262,7 @@ impl<'a, 'b> AppArgumentAdder for App<'a, 'b> {
                     batch signing private key to be used by this server when \
                     sending messages to other servers.",
                 )
-                .required(true)
-                .validator(b64_validator),
+                .required(true),
         )
         .arg(
             Arg::with_name("batch-signing-private-key-identifier")
@@ -301,7 +294,6 @@ impl<'a, 'b> AppArgumentAdder for App<'a, 'b> {
                 .multiple(true)
                 .min_values(1)
                 .use_delimiter(true)
-                .validator(b64_validator)
                 .required(true),
         )
     }
@@ -319,6 +311,60 @@ impl<'a, 'b> AppArgumentAdder for App<'a, 'b> {
                     account found in the GKE metadata service will be used for \
                     authentication or impersonation.",
                 ),
+        )
+    }
+
+    fn add_task_queue_arguments(self: App<'a, 'b>) -> App<'a, 'b> {
+        self.arg(
+            Arg::with_name("task-queue-kind")
+                .long("task-queue-kind")
+                .env("TASK_QUEUE_KIND")
+                .help("kind of task queue to use")
+                .possible_value(leak_string(TaskQueueKind::GcpPubSub.to_string()))
+                .possible_value(leak_string(TaskQueueKind::AwsSqs.to_string()))
+                .required(true),
+        )
+        .arg(
+            Arg::with_name("task-queue-name")
+                .long("task-queue-name")
+                .env("TASK_QUEUE_NAME")
+                .help("Name of queue from which tasks should be pulled.")
+                .long_help(
+                    "Name of queue from which tasks should be pulled. On GCP, \
+                    a PubSub subscription ID. On AWS, an SQS queue URL.",
+                )
+                .required(true),
+        )
+        .arg(
+            Arg::with_name("task-queue-identity")
+                .long("task-queue-identity")
+                .env("TASK_QUEUE_IDENTITY")
+                .help("Identity to assume when accessing task queue"),
+        )
+        .arg(
+            Arg::with_name("gcp-project-id")
+                .long("gcp-project-id")
+                .env("GCP_PROJECT_ID")
+                .help("Project ID for GCP PubSub topic")
+                .long_help(
+                    "The GCP Project ID in which the PubSub topic implementing \
+                    the work queue was created. Required if the task queue is \
+                    a PubSub topic.",
+                ),
+        )
+        .arg(
+            Arg::with_name("pubsub-api-endpoint")
+                .long("pubsub-api-endpoint")
+                .env("PUBSUB_API_ENDPOINT")
+                .help("API endpoint for GCP PubSub")
+                .default_value("https://pubsub.googleapis.com")
+                .help("API endpoint for GCP PubSub. Optional."),
+        )
+        .arg(
+            Arg::with_name("sqs-region")
+                .long("sqs-region")
+                .env("SQS_REGION")
+                .help("AWS region in which to use SQS"),
         )
     }
 }
@@ -406,8 +452,7 @@ fn main() -> Result<(), anyhow::Error> {
                             "Base64 encoded ECIES private key for the PHA \
                             server",
                         )
-                        .required(true)
-                        .validator(b64_validator),
+                        .required(true),
                 )
                 .arg(
                     Arg::with_name("facilitator-ecies-private-key")
@@ -418,8 +463,7 @@ fn main() -> Result<(), anyhow::Error> {
                             "Base64 encoded ECIES private key for the \
                             facilitator server",
                         )
-                        .required(true)
-                        .validator(b64_validator),
+                        .required(true),
                 )
                 .add_batch_signing_key_arguments()
                 .arg(
@@ -607,6 +651,42 @@ fn main() -> Result<(), anyhow::Error> {
                     )
             )
         )
+        .subcommand(
+            SubCommand::with_name("intake-batch-worker")
+                .about(format!("Consume intake batch tasks from a queue, validating an input share (from an ingestor's bucket) and emit a validation share.\n\n{}", SHARED_HELP).as_str())
+                .add_instance_name_argument()
+                .add_is_first_argument()
+                .add_gcp_service_account_key_file_argument()
+                .add_packet_decryption_key_argument()
+                .add_batch_public_key_arguments(Entity::Ingestor)
+                .add_batch_signing_key_arguments()
+                .add_manifest_base_url_argument(Entity::Ingestor)
+                .add_storage_arguments(Entity::Ingestor, InOut::Input)
+                .add_manifest_base_url_argument(Entity::Peer)
+                .add_storage_arguments(Entity::Peer, InOut::Output)
+                .add_storage_arguments(Entity::Own, InOut::Output)
+                .add_task_queue_arguments()
+        )
+        .subcommand(
+            SubCommand::with_name("aggregate-worker")
+                .about(format!("Consume aggregate tasks from a queue.\n\n{}", SHARED_HELP).as_str())
+                .add_instance_name_argument()
+                .add_is_first_argument()
+                .add_gcp_service_account_key_file_argument()
+                .add_manifest_base_url_argument(Entity::Ingestor)
+                .add_storage_arguments(Entity::Ingestor, InOut::Input)
+                .add_batch_public_key_arguments(Entity::Ingestor)
+                .add_manifest_base_url_argument(Entity::Own)
+                .add_storage_arguments(Entity::Own, InOut::Input)
+                .add_manifest_base_url_argument(Entity::Peer)
+                .add_storage_arguments(Entity::Peer, InOut::Input)
+                .add_batch_public_key_arguments(Entity::Peer)
+                .add_manifest_base_url_argument(Entity::Portal)
+                .add_storage_arguments(Entity::Portal, InOut::Output)
+                .add_packet_decryption_key_argument()
+                .add_batch_signing_key_arguments()
+                .add_task_queue_arguments()
+        )
         .get_matches();
 
     let result = match matches.subcommand() {
@@ -614,30 +694,13 @@ fn main() -> Result<(), anyhow::Error> {
         // various parameters are present and valid, so it is safe to use
         // unwrap() here.
         ("generate-ingestion-sample", Some(sub_matches)) => generate_sample(sub_matches),
-        ("intake-batch", Some(sub_matches)) => intake_batch(sub_matches),
-        ("aggregate", Some(sub_matches)) => aggregate(sub_matches),
+        ("intake-batch", Some(sub_matches)) => intake_batch_subcommand(sub_matches),
+        ("intake-batch-worker", Some(sub_matches)) => intake_batch_worker(sub_matches),
+        ("aggregate", Some(sub_matches)) => aggregate_subcommand(sub_matches),
+        ("aggregate-worker", Some(sub_matches)) => aggregate_worker(sub_matches),
         ("lint-manifest", Some(sub_matches)) => lint_manifest(sub_matches),
         (_, _) => Ok(()),
     };
-
-    // Once we've run the subcommand, whether it succeeds or fails, we want to push any metrics
-    // we have collected.
-    if let Some(pushgateway) = matches.value_of("pushgateway") {
-        if pushgateway != "" {
-            info!("pushing metrics to {}", pushgateway);
-            let jobname: &str = matches.subcommand().0;
-            prometheus::push_metrics(
-                jobname,
-                prometheus::labels! {},
-                pushgateway,
-                prometheus::gather(),
-                None,
-            )
-            .map(|_| info!("done pushing metrics"))
-            .map_err(|e| error!("error pushing metrics: {}", e))
-            .ok();
-        }
-    }
 
     result
 }
@@ -692,7 +755,12 @@ fn generate_sample(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
     Ok(())
 }
 
-fn intake_batch(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
+fn intake_batch(
+    aggregation_id: &str,
+    batch_id: &str,
+    date: &str,
+    sub_matches: &ArgMatches,
+) -> Result<(), anyhow::Error> {
     let mut intake_transport = intake_transport_from_args(sub_matches)?;
 
     // We need the bucket to which we will write validations for the
@@ -723,14 +791,12 @@ fn intake_batch(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
         batch_signing_key: batch_signing_key_from_arg(sub_matches)?,
     };
 
-    let batch_id: &str = sub_matches.value_of("batch-id").unwrap();
     let batch_id: Uuid = Uuid::parse_str(batch_id).unwrap();
 
-    let date: &str = sub_matches.value_of("date").unwrap();
     let date: NaiveDateTime = NaiveDateTime::parse_from_str(date, DATE_FORMAT).unwrap();
 
     let mut batch_intaker = BatchIntaker::new(
-        &sub_matches.value_of("aggregation-id").unwrap(),
+        &aggregation_id,
         &batch_id,
         &date,
         &mut intake_transport,
@@ -739,30 +805,69 @@ fn intake_batch(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
         is_first_from_arg(sub_matches),
     )?;
 
+    batch_intaker.generate_validation_share()
+}
+
+fn intake_batch_subcommand(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
+    intake_batch(
+        sub_matches.value_of("aggregation-id").unwrap(),
+        sub_matches.value_of("batch-id").unwrap(),
+        sub_matches.value_of("date").unwrap(),
+        sub_matches,
+    )
+}
+
+fn intake_batch_worker(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
     let intake_started: Counter = register_counter!(
         "intake_jobs_begun",
         "Number of intake-batch jobs that started (on the facilitator side)"
     )
-    .unwrap();
-    intake_started.inc();
-
-    let result = batch_intaker.generate_validation_share();
+    .context("failed to register metrics counter for started intakes")?;
 
     let intake_finished = register_counter_vec!(
         "intake_jobs_finished",
         "Number of intake-batch jobs that finished (on the facilitator side)",
         &["status"]
     )
-    .unwrap();
-    match result {
-        Ok(_) => intake_finished.with_label_values(&["success"]).inc(),
-        Err(_) => intake_finished.with_label_values(&["error"]).inc(),
+    .context("failed to register metrics counter for finished intakes")?;
+
+    let mut queue = intake_task_queue_from_args(sub_matches)?;
+
+    loop {
+        if let Some(task_handle) = queue.dequeue()? {
+            info!("dequeued task: {:?}", task_handle);
+            intake_started.inc();
+            let result = intake_batch(
+                &task_handle.task.aggregation_id,
+                &task_handle.task.batch_id,
+                &task_handle.task.date,
+                sub_matches,
+            );
+
+            match result {
+                Ok(_) => {
+                    intake_finished.with_label_values(&["success"]).inc();
+                    queue.acknowledge_task(task_handle)?;
+                }
+                Err(err) => {
+                    error!("error while processing task {:?}: {:?}", task_handle, err);
+                    intake_finished.with_label_values(&["error"]).inc();
+                    queue.nacknowledge_task(task_handle)?;
+                }
+            }
+        }
     }
 
-    result
+    // unreachable
 }
 
-fn aggregate(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
+fn aggregate(
+    aggregation_id: &str,
+    start: &str,
+    end: &str,
+    batches: Vec<(&str, &str)>,
+    sub_matches: &ArgMatches,
+) -> Result<()> {
     let instance_name = sub_matches.value_of("instance-name").unwrap();
     let is_first = is_first_from_arg(sub_matches);
 
@@ -788,7 +893,7 @@ fn aggregate(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
                 .batch_signing_public_keys()?
         }
         (_, Some(private_key), Some(private_key_identifier)) => {
-            public_key_map_from_arg(private_key, private_key_identifier)
+            public_key_map_from_arg(private_key, private_key_identifier)?
         }
         _ => {
             return Err(anyhow!(
@@ -821,7 +926,7 @@ fn aggregate(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
                 .batch_signing_public_keys()?
         }
         (Some(public_key), Some(public_key_identifier), _) => {
-            public_key_map_from_arg(public_key, public_key_identifier)
+            public_key_map_from_arg(public_key, public_key_identifier)?
         }
         _ => {
             return Err(anyhow!(
@@ -853,25 +958,7 @@ fn aggregate(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
     // portal server.
     let batch_signing_key = batch_signing_key_from_arg(sub_matches)?;
 
-    let batch_ids: Vec<Uuid> = sub_matches
-        .values_of("batch-id")
-        .context("no batch-id")?
-        .map(|v| Uuid::parse_str(v).unwrap())
-        .collect();
-    let batch_dates: Vec<NaiveDateTime> = sub_matches
-        .values_of("batch-time")
-        .context("no batch-time")?
-        .map(|s| NaiveDateTime::parse_from_str(&s, DATE_FORMAT).unwrap())
-        .collect();
-    if batch_ids.len() != batch_dates.len() {
-        return Err(anyhow!(
-            "must provide same number of batch-id and batch-date values"
-        ));
-    }
-
-    let start: &str = sub_matches.value_of("aggregation-start").unwrap();
     let start: NaiveDateTime = NaiveDateTime::parse_from_str(start, DATE_FORMAT).unwrap();
-    let end: &str = sub_matches.value_of("aggregation-end").unwrap();
     let end: NaiveDateTime = NaiveDateTime::parse_from_str(end, DATE_FORMAT).unwrap();
 
     let mut own_validation_transport = VerifiableTransport {
@@ -887,10 +974,17 @@ fn aggregate(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
         batch_signing_key,
     };
 
-    let batch_info: Vec<_> = batch_ids.into_iter().zip(batch_dates).collect();
+    let mut parsed_batches: Vec<(Uuid, NaiveDateTime)> = Vec::new();
+    for raw_batch in batches.iter() {
+        let uuid = Uuid::parse_str(raw_batch.0).context("batch ID is not a UUID")?;
+        let date = NaiveDateTime::parse_from_str(raw_batch.1, DATE_FORMAT)
+            .context("batch date is not in expected format")?;
+        parsed_batches.push((uuid, date));
+    }
+
     let mut aggregator = BatchAggregator::new(
         instance_name,
-        &sub_matches.value_of("aggregation-id").unwrap(),
+        aggregation_id,
         &start,
         &end,
         is_first,
@@ -900,27 +994,84 @@ fn aggregate(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
         &mut aggregation_transport,
     )?;
 
+    aggregator.generate_sum_part(&parsed_batches)
+}
+
+fn aggregate_subcommand(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
+    let batch_ids: Vec<&str> = sub_matches
+        .values_of("batch-id")
+        .context("no batch-id")?
+        .collect();
+    let batch_dates: Vec<&str> = sub_matches
+        .values_of("batch-time")
+        .context("no batch-time")?
+        .collect();
+
+    if batch_ids.len() != batch_dates.len() {
+        return Err(anyhow!(
+            "must provide same number of batch-id and batch-date values"
+        ));
+    }
+    let batch_info: Vec<_> = batch_ids.into_iter().zip(batch_dates).collect();
+
+    aggregate(
+        &sub_matches.value_of("aggregation-id").unwrap(),
+        sub_matches.value_of("aggregation-start").unwrap(),
+        sub_matches.value_of("aggregation-end").unwrap(),
+        batch_info,
+        sub_matches,
+    )
+}
+
+fn aggregate_worker(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
     let aggregation_started: Counter = register_counter!(
         "aggregation_jobs_begun",
         "Number of aggregation jobs that started (on the facilitator side)"
     )
-    .unwrap();
-    aggregation_started.inc();
-
-    let result = aggregator.generate_sum_part(&batch_info);
-
+    .context("failed to register counter for started aggregations")?;
     let aggregation_finished = register_counter_vec!(
         "aggregation_jobs_finished",
         "Number of aggregation jobs that finished (on the facilitator side)",
         &["status"]
     )
-    .unwrap();
-    match result {
-        Ok(_) => aggregation_finished.with_label_values(&["success"]).inc(),
-        Err(_) => aggregation_finished.with_label_values(&["error"]).inc(),
+    .context("failed to register counter for finished aggregations")?;
+
+    let mut queue = aggregation_task_queue_from_args(sub_matches)?;
+
+    loop {
+        if let Some(task_handle) = queue.dequeue()? {
+            info!("dequeued task: {:?}", task_handle);
+            aggregation_started.inc();
+
+            let batches: Vec<(&str, &str)> = task_handle
+                .task
+                .batches
+                .iter()
+                .map(|b| (b.id.as_str(), b.time.as_str()))
+                .collect();
+            let result = aggregate(
+                &task_handle.task.aggregation_id,
+                &task_handle.task.aggregation_start,
+                &task_handle.task.aggregation_end,
+                batches,
+                sub_matches,
+            );
+
+            match result {
+                Ok(_) => {
+                    aggregation_finished.with_label_values(&["success"]).inc();
+                    queue.acknowledge_task(task_handle)?;
+                }
+                Err(err) => {
+                    error!("error while processing task {:?}: {:?}", task_handle, err);
+                    aggregation_finished.with_label_values(&["error"]).inc();
+                    queue.nacknowledge_task(task_handle)?;
+                }
+            }
+        }
     }
 
-    result
+    // unreachable
 }
 
 fn lint_manifest(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
@@ -1008,10 +1159,10 @@ fn is_first_from_arg(matches: &ArgMatches) -> bool {
 fn public_key_map_from_arg(
     key: &str,
     key_identifier: &str,
-) -> HashMap<String, UnparsedPublicKey<Vec<u8>>> {
+) -> Result<HashMap<String, UnparsedPublicKey<Vec<u8>>>> {
     // UnparsedPublicKey::new doesn't return an error, so try parsing the
     // argument as a private key first.
-    let key_bytes = base64::decode(key).unwrap();
+    let key_bytes = decode_base64_key(key)?;
     let public_key = match EcdsaKeyPair::from_pkcs8(&ECDSA_P256_SHA256_ASN1_SIGNING, &key_bytes) {
         Ok(priv_key) => UnparsedPublicKey::new(
             &ECDSA_P256_SHA256_ASN1,
@@ -1022,11 +1173,11 @@ fn public_key_map_from_arg(
 
     let mut key_map = HashMap::new();
     key_map.insert(key_identifier.to_owned(), public_key);
-    key_map
+    Ok(key_map)
 }
 
 fn batch_signing_key_from_arg(matches: &ArgMatches) -> Result<BatchSigningKey> {
-    let key_bytes = base64::decode(matches.value_of("batch-signing-private-key").unwrap()).unwrap();
+    let key_bytes = decode_base64_key(matches.value_of("batch-signing-private-key").unwrap())?;
     let key_identifier = matches
         .value_of("batch-signing-private-key-identifier")
         .unwrap();
@@ -1054,7 +1205,7 @@ fn intake_transport_from_args(matches: &ArgMatches) -> Result<VerifiableAndDecry
         matches.value_of("ingestor-manifest-base-url"),
     ) {
         (Some(public_key), Some(public_key_identifier), _) => {
-            public_key_map_from_arg(public_key, public_key_identifier)
+            public_key_map_from_arg(public_key, public_key_identifier)?
         }
         (_, _, Some(manifest_base_url)) => IngestionServerManifest::from_https(
             manifest_base_url,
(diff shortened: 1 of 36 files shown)


workflow-manager/task/task.go
```diff
@@ -0,0 +1,156 @@
...
+	Enqueue(task Task, completion func(error))
+	// Stop blocks until all tasks passed to Enqueue() have been enqueued in the
+	// underlying system, and all completion functions pased to Enqueue() have
+	// returned, and so it is safe to exit the program without losing any tasks.
+	Stop()
+}
+
+// CreatePubSubTopic creates a PubSub topic with the provided ID. Returns error
+// on failure.
+func CreatePubSubTopic(ctx context.Context, project string, topicID string) error {
```
hostirosti, line 82: I'd create the topics with Terraform and not in the workflow-manager.

workflow-manager/task/task.go
```diff
@@ -0,0 +1,156 @@
...
+		}()
+		return
+	}
+
+	// Publish() returns immediately, giving us a handle to the result that we
+	// can block on to see if publishing succeeded. The PubSub client
+	// automatically retries for us, so we just keep the handle so the caller
+	// can do whatever they need to after successful publication and we can
+	// block in Stop() until all tasks have been enqueued
+	res := e.topic.Publish(e.context, &pubsub.Message{Data: jsonTask})
```
hostirosti, line 142: Why don't you put the entire work (Publish + waiting for response) into the goroutine?

workflow-manager/task/task.go
tgeoghegan: That's the plan -- this exists only to support local development cases, working with the PubSub emulator, since it is not going to be managed by Terraform and you can't use `gcloud` to create topics or subscriptions.

workflow-manager/task/task.go
tgeoghegan: You're right, it makes the error handling much simpler to do everything in the goroutine.

workflow-manager/task/task.go
```diff
@@ -0,0 +1,222 @@
...
+		return fmt.Errorf("pubsub.CreateSubscription: %w", err)
+	}
+
+	return nil
+}
+
+// GCPPubSubEnqueuer implements Enqueuer using GCP PubSub
+type GCPPubSubEnqueuer struct {
+	topic     *pubsub.Topic
+	context   context.Context
```
jsha, line 116: https://golang.org/pkg/context/#pkg-overview

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

workflow-manager/task/task.go
```diff
@@ -0,0 +1,222 @@
...
+func (a Aggregation) MarshalToJson() ([]byte, error) {
+	return json.Marshal(a)
+}
+
+// Batch represents a batch included in an aggregation task
+type Batch struct {
+	// ID is the batch ID. Typically a UUID.
+	ID string `json:"id"`
+	// Time is the timestamp on the batch, in UTC, with minute precision,
+	// formatted like "2006/01/02/15/04"
```
jsha, line 53: Since we have specific requirements for the timestamp format, this would be a nice thing to enforce at the API boundary between the `task` package and `main`. For instance, instead of letting `main` directly instantiate `Batch`, you could provide a function that builds `Batches`. That function would take a `*time.Time` (Go's internal time type), and format it in the expected wire format.

Alternately, you could make the field of type `SlashyMinuteTime` (a type alias for `*time.Time), and define a custom JSON marshaller for `SlashyMinuteTime`. That would let you continue instantiating `Batch` directly, while also enforcing correct wire format.

workflow-manager/task/task.go
```diff
@@ -0,0 +1,222 @@
...
+	if err != nil {
+		completion(fmt.Errorf("failed to publish task %+v: %w", task, err))
+		return
+	}
+
+	completion(nil)
+}
+
+func (e *AWSSNSEnqueuer) Stop() {
+	// No-op since sns.Publish() blocks until completion or failure
```
jsha, line 221: In theory a future refactor could make it so that `Enqueue` is called concurrently in the main thread. In that situation, there could be inflight `Enqueue` calls when `.Stop()` is called, and this would be incorrect. I think it makes sense to go ahead and add a WaitGroup to `AWSNSEnqueue`, even though you'll be doing `.Add(1)` and `.Done()` in the same function with no goroutines.

workflow-manager/utils/utils.go
```diff
@@ -8,3 +22,62 @@ func Index(isFirst bool) int {
...
+	return Clock{now: now}
+}
+
+// Now returns the current time according to this clock
+func (c *Clock) Now() time.Time {
+	if c.now.IsZero() {
+		return time.Now()
+	}
+	return c.now
+}
```
jsha, line 47: This approach is also implemented by https://github.com/jmhodges/clock, in case you would prefer to use that.

workflow-manager/utils/utils.go
```diff
@@ -8,3 +22,62 @@ func Index(isFirst bool) int {
...
+// New() functions for various AWS services. If identity contains a valid role
+// ARN, the config will use a web identity role provider for that role.
+func AWSClientConfig(region, identity string) (*aws_session.Session, *aws.Config, error) {
+	sess, err := aws_session.NewSession()
+	if err != nil {
+		return nil, nil, fmt.Errorf("making AWS session: %w", err)
+	}
+
+	config := aws.NewConfig().
+		WithRegion(region)
```
jsha, line 74: This line break is a little odd - I'd join them.

workflow-manager/utils/utils.go
```diff
@@ -8,3 +22,62 @@ func Index(isFirst bool) int {
...
+	roleProvider := stscreds.NewWebIdentityRoleProviderWithToken(
+		stsSTS, identity, roleSessionName, tokenfetcher.NewTokenFetcher(audience))
+
+	return credentials.NewCredentials(roleProvider), nil
+}
+
+// AWSClientConfig returns a (Session, Config) pair suitable for passing to the
+// New() functions for various AWS services. If identity contains a valid role
+// ARN, the config will use a web identity role provider for that role.
+func AWSClientConfig(region, identity string) (*aws_session.Session, *aws.Config, error) {
```
jsha, line 67: I think it's best to have a very high bar for `util` packages, lest they become a crowded dumping ground. I think in this case there's a clear enough theme to put this function in its own `aws` package (where it would become `aws.ClientConfig`).

jsha (changes requested):
Nice work! Thanks for cranking this out. It will be a big improvement over our current system.

Review on 8a19bea.

Why is the task marker defined using the namespace of job names, which are themselves a transform of batch paths? It seems like it would be better for task markers to live alongside the batch paths they correspond to. For instance, an intake batch `2021/01/01/01/01/kittens-seen.batch` would get a corresponding `2021/01/01/01/01/kittens-seen.batch.inprogress`. Or, if you didn't want to pollute the input bucket, a corresponding file in a different input bucket, or with a different prefix.

For aggregation batches it's a bit messier, but naming the `.inprogress` file after the expected output of the aggregation batch would probably work well.

terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -331,9 +494,9 @@ resource "kubernetes_cron_job" "sample_maker" {
               args = [
                 "--pushgateway", var.pushgateway,
                 "generate-ingestion-sample",
-                "--own-output", var.ingestion_bucket,
-                "--peer-output", var.test_peer_ingestion_bucket,
-                "--peer-identity", var.remote_peer_validation_bucket_identity,
+                "--peer-output", var.ingestion_bucket,
+                "--own-output", var.test_peer_ingestion_bucket,
+                "--own-identity", var.remote_peer_validation_bucket_identity,
```
jsha, line 499: Out of curiosity, why did peer-output and own-output switch for the sample-maker deployment?

workflow-manager/task/task.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// Enqueuer allows enqueuing tasks.
+type Enqueuer interface {
+	// Enqueue enqueues a task to be executed later. The provided completion
+	// function will be invoked once the task is either successfully enqueued or
+	// some unretryable error has occurred. A call to Stop() will not return
+	// until completion functions passed to any and all calls to Enqueue() have
+	// returned.
+	Enqueue(task Task, completion func(error))
```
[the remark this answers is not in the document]
tgeoghegan: Yes, and also so that we can avoid storing a `context.Context` in the `GCPPubSubEnqueuer` struct per your other piece of feedback.

terraform/modules/kubernetes/kubernetes.tf
tgeoghegan: While testing, I found that they were wrong, and were causing packet decryption errors because the packet encryption keys were getting mixed up. It presumably had been wrong for some time, presumably since some change I made hastily in December without testing it through a paired test env.

terraform/main.tf
```diff
@@ -315,7 +321,9 @@ locals {
       ingestor                                = pair[1]
       kubernetes_namespace                    = kubernetes_namespace.namespaces[pair[0]].metadata[0].name
       packet_decryption_key_kubernetes_secret = kubernetes_secret.ingestion_packet_decryption_keys[pair[0]].metadata[0].name
-      ingestor_manifest_base_url              = var.ingestors[pair[1]]
+      ingestor_manifest_base_url              = var.ingestors[pair[1]].manifest_base_url
+      intake_batch_replica_count              = var.ingestors[pair[1]].localities[pair[0]].intake_batch_replica_count
```
hostirosti, line 325: minor nit on naming. Are those the replica counts for the intake batch and aggregate workers? If yes, `intake_worker_count` and `aggregate_worker_count` might be more descriptive.

terraform/modules/data_share_processor/data_share_processor.tf
```diff
@@ -419,6 +435,10 @@ module "kubernetes" {
   container_registry                      = var.container_registry
   workflow_manager_version                = var.workflow_manager_version
   facilitator_version                     = var.facilitator_version
+  intake_tasks_queue                      = module.pubsub.intake_batch_queue
```
hostirosti, line 438: again minor nit around consistent `intake_tasks_queue` vs `intake_batch_queue` --> I'd just use `intake_queue`

terraform/modules/gke/gke.tf
```diff
@@ -84,9 +84,9 @@ resource "google_container_node_pool" "worker_nodes" {
   name               = "${var.resource_prefix}-node-pool"
   location           = var.gcp_region
   cluster            = google_container_cluster.cluster.name
-  initial_node_count = 3
+  initial_node_count = 4
   autoscaling {
-    min_node_count = 2
+    min_node_count = 4
```
hostirosti, line 89: what's the reason for going up to 4 here?

terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -298,6 +321,146 @@ resource "kubernetes_cron_job" "workflow_manager" {
...
+        container {
+          name  = "facile-container"
+          image = "${var.container_registry}/${var.facilitator_image}:${var.facilitator_version}"
+          args  = ["intake-batch-worker"]
+          resources {
+            # Batch intake is single-threaded, so we request an entire CPU for
+            # maximum throughput. We never expect to see batches larger than
+            # 3-400 MB, and batches are processed in memory, so request enough
+            # for whole ingestion batches plus a safety margin.
+            requests {
```
hostirosti, line 354: Did you check what the used resources on small batches is? 500Mi and 1.0 CPU seems quite high imho. I'd start lower (but I don't know rust and how memory hungry it is). However, for CPU I'd go lower since you can bin pack more workers on one node this way and they'll use what they can get, if they can't get more things will just take a bit longer.

terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -298,6 +321,146 @@ resource "kubernetes_cron_job" "workflow_manager" {
...
+              memory = "500Mi"
+              cpu    = "1.0"
+            }
+            limits {
```
hostirosti, line 358: With having requests and limits so close to each other you limit the possibility for bin packing and better resource utilization. The limits are just there to basically tame possible run-away's and drive them into OOMs or starve them of hogging CPU. The requests are there to help a bit the bin packing if you know the lower end of resource requirements well.

terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -298,6 +321,146 @@ resource "kubernetes_cron_job" "workflow_manager" {
...
+      spec {
+        service_account_name = module.account_mapping.kubernetes_account_name
+        container {
+          name  = "facile-container"
+          image = "${var.container_registry}/${var.facilitator_image}:${var.facilitator_version}"
+          args  = ["aggregate-worker"]
+          resources {
+            # As in the intake-batch case, aggregate jobs are single threaded
+            # and need to fit whole ingestion batches into memory.
+            requests {
```
hostirosti, line 423: Same as above, I'd go to the minimal requirements for small batches on the request side to start with.

workflow-manager/main.go
```diff
@@ -115,23 +113,99 @@ func main() {
...
+		log.Fatalf("--task-queue-kind, --intake-tasks-topic and --aggregate-tasks-topic are required")
+	}
+
+	var intakeTaskEnqueuer task.Enqueuer
+	var aggregationTaskEnqueuer task.Enqueuer
+
+	switch *taskQueueKind {
+	case "gcp-pubsub":
+		if *gcpPubSubProjectID == "" {
+			log.Fatal("--gcp-pubsub-project-id is required for task-queue-kind=gcp-pubsub")
```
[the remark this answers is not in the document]
hostirosti: Agree, no need to separate those or at least default to `gcp-project-id` if you want to keep the flexibility.

workflow-manager/main.go
```diff
@@ -115,23 +113,99 @@ func main() {
...
+		}
+
+		if *gcpPubSubCreatePubSubTopics {
+			if err := task.CreatePubSubTopic(
+				context.Background(),
```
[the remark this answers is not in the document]
hostirosti: I think it's advised against that if my memory doesn't fail me. The context should be kept as small as possible (if you need it to live across multiple function call yes, but not across the lifetime of a process (it times out)).

workflow-manager/main.go
```diff
@@ -140,32 +214,99 @@ func main() {
...
-	log.Printf("found %d own validations", len(ownValidationBatches))
+	scheduleTasks(
+		*isFirst,
+		utils.DefaultClock(),
+		intakeFiles, ownValidationFiles, peerValidationFiles,
+		existingJobs,
+		intakeTaskEnqueuer, aggregationTaskEnqueuer,
+		ownValidationBucket,
+		maxAgeParsed, aggregationPeriodParsed, gracePeriodParsed,
+	)
```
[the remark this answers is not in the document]
hostirosti: +1

workflow-manager/task/task.go
```diff
@@ -0,0 +1,222 @@
...
+
+	"cloud.google.com/go/pubsub"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/sns"
+)
+
+// Task is a task that can be enqueued into an Enqueuer
+type Task interface {
+	// MarshalToJson marshals self into a JSON message
+	MarshalToJson() ([]byte, error)
```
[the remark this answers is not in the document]
hostirosti: No need for the json.Marshaler interface imho.

Unless you have a special encoding you can just use the default encoder 

Sample in an API context
```
w.Header().Set("Content-Type", "application/json; charset=UTF-8")
w.WriteHeader(http.StatusCreated)
if err := json.NewEncoder(w).Encode(a);
```

workflow-manager/task/task.go
hostirosti: +100

facilitator/src/bin/facilitator.rs
```diff
@@ -900,27 +1013,101 @@ fn aggregate(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
...
+            );
+
+            match result {
+                Ok(_) => {
+                    aggregation_finished.with_label_values(&["success"]).inc();
+                    queue.acknowledge_task(task_handle)?;
+                }
+                Err(err) => {
+                    error!("error while processing task {}: {:?}", task_handle, err);
+                    aggregation_finished.with_label_values(&["error"]).inc();
```
hostirosti, line 1086: This might be a place to nack the pubsub message or send it to a deadletter queue.

facilitator/src/task/pubsub.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        let handle = TaskHandle {
+            task: T::from_json(&task_json)?,
+            acknowledgment_id: received_messages[0].ack_id.clone(),
+        };
+
+        Ok(Some(handle))
+    }
+
+    fn acknowledge_task(&mut self, handle: TaskHandle<T>) -> Result<()> {
```
hostirosti, line 148: fn for nack or send to dead_letter queue is missing. I recommend a dead-letter queue, in that case you might want to differentiate between non-recoverable errors and recoverable errors. For non-recoverable errors you send to the dead-letter queue (which is a way to help investigate those issues without OOM (capacity-wise) the system with unprocessable tasks. For recoverable errors it would be good to keep a retry counter (when limit reached they are sent to a dead-letter queue as well) -- multiple ways you could implement that, either you ack the message and queue it with an updated retry counter, or you use your maker files to keep track of the counter.

workflow-manager/task/task.go
hostirosti: Remove ctx from function call and create a separate context here:

```
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
defer cancel()
```

workflow-manager/task/task.go
aaomidi: TIL

workflow-manager/main_test.go
```diff
@@ -49,3 +58,461 @@ func TestAggregationJobNameFragment(t *testing.T) {
...
+				}
+				if !foundExpectedMarker {
+					t.Errorf("did not find expected task marker among %q", ownValidationBucket.writtenObjectKeys)
+				}
+			}
+		})
+	}
+}
+
+func TestScheduleAggregationTasks(t *testing.T) {
```
aaomidi, line 244: I love the thoroughness in testing here

terraform/modules/fake_server_resources/fake_server_resources.tf
```diff
@@ -15,7 +15,7 @@ variable "sum_part_bucket_writer_email" {
 }
 
 variable "ingestors" {
-  type = map(string)
+  type = list(string)
```
aaomidi, line 18: Are we sure we're not going to need the actual ingestor addresses here in the future?

workflow-manager/task/task.go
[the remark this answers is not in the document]
tgeoghegan: I wound up adding a different method `Marker() string` to the `Task` interface that returns the task marker, per jsha's suggestion that we don't need to use the abbrevated Kubernetes job names for this, so now the `Task` interface doesn't need to mention JSON at all.

workflow-manager/task/task.go
tgeoghegan: Go type aliases are a bit of a pain because you have to type assert back to the original thing to use any of its methods, but I still like the idea. I introduced `type Timestamp time.Time` with a custom JSON formatter. I'm curious though: is there a reason for it to be an alias to `*time.Time` instead of just `time.Time`?

workflow-manager/utils/utils.go
tgeoghegan: Had I known, I would have used it, but I'm happy with this `Clock`.

workflow-manager/utils/utils.go
tgeoghegan: Refactored this into the `ClientConfig` function in the `aws` package.

terraform/modules/kubernetes/kubernetes.tf
tgeoghegan: The real batches we've seen thus far are much, much smaller than this. On average, about 3 KB for Apple and 135 KB for Google across MD and CT. You're right that since we no longer rely on the container resource requests to limit how many jobs run at once, we can set these much lower. I'll set 0.1 CPU and 50 MB of memory.

facilitator/src/bin/facilitator.rs
```diff
@@ -739,30 +807,86 @@ fn intake_batch(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
...
+            );
+
+            match result {
+                Ok(_) => {
+                    intake_finished.with_label_values(&["success"]).inc();
+                    queue.acknowledge_task(task_handle)?;
+                }
+                Err(err) => {
+                    error!("error while processing task {}: {:?}", task_handle, err);
+                    intake_finished.with_label_values(&["error"]).inc();
```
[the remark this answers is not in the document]
tgeoghegan: PubSub will delete a message if it is not acked after 7 days, and will try to redeliver the message with exponential backoff until then. I like the idea of relying on PubSub for retries -- it saves me the trouble of writing retry code. I have not set up any dead letter queue because I don't know what I would do with it, to be honest. Email someone? Alert about the job we're giving up on?

We could nack messages much more aggressively then 7 days. There's no point to retrying an intake job if the aggregation window it corresponds to is long gone, for instance. I think we could achieve that by looking at the message redelivery count in the PubSub API response, but I'd like to leave that for a future change: #295

terraform/modules/fake_server_resources/fake_server_resources.tf
tgeoghegan: If we do, then we can change this back to a richer representation.

facilitator/src/bin/facilitator.rs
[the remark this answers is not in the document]
tgeoghegan: I'm reading up on this more and it seems like dead letter topics and more aggressive nacks are (1) a good idea and (2) pretty simple to wire up so I'll just do them in this PR.

facilitator/src/gcp_oauth.rs
```diff
@@ -0,0 +1,291 @@
...
+        if let Some(token) = &self.impersonated_account_token {
+            if !token.expired() {
+                return Ok(token.token.clone());
+            }
+        }
+
+        let service_account_to_impersonate = self.account_to_impersonate.clone().unwrap();
+        // API reference:
+        // https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken
+        let request_url = format!("https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{}:generateAccessToken",
```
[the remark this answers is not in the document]
tgeoghegan: `ureq` treats URLs as `&str` so I think what we've got is good enough. We could pick a URL crate and be more rigorous but that's outside of the scope of this PR.

workflow-manager/main.go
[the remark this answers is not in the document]
jsha: Good point about not making the context span the whole length of the program; I withdraw my comment.

workflow-manager/task/task.go
jsha: > Go type aliases are a bit of a pain because you have to type assert back to the original thing to use any of its methods, but I still like the idea.

For type aliases, you're actually doing a type conversion, not a type assertion, which is both safer and syntactically somewhat cleaner:

https://www.sohamkamani.com/golang/type-assertions-vs-type-conversions/

Short version: a type assertion is fallible and done at runtime. A type conversion is infallible and checked at compile time.

> is there a reason for it to be an alias to *time.Time instead of just time.Time?

Nope. I can never remember whether it's typically passed as a pointer or not, so I took a guess.

facilitator/src/gcp_oauth.rs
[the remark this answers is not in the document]
jsha: FYI, ureq 2.0 is just released and adds a way to pass a pre-parsed `Url` to create a request: https://docs.rs/ureq/2.0.0/ureq/struct.Agent.html#method.request_url.

If you wanted to be really cautious about building this URL you could parse the constant part, and then use [`Url::path_segments_mut` ](https://docs.rs/url/2.2.0/url/struct.Url.html#method.path_segments_mut) to append a path segment. But I think that's probably overkill for this specific use case, and as Tim said, beyond the scope of this PR.

terraform/modules/gke/gke.tf
tgeoghegan: That was accidental, I'll remove this portion of the change

facilitator/src/bin/facilitator.rs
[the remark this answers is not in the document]
tgeoghegan: So I think I understand all this better now. We now create a dead letter topic for each subscription, which causes PubSub to [redeliver messages up to 5 times](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions#deadletterpolicy) until they are ACKed. The [default retry policy](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions#retrypolicy) means PubSub will use exponential backoff to wait between 10 seconds and 10 minutes to redeliver unacked messages. So that means the default retention policy of 7 days should be irrelevant -- six delivery attempts (1 + 5 redeliveries) should have gone by long before then, and messages that fail to be processed six times go to the dead letter queue for later examination by humans.

I'd also misunderstood how nacking works -- I thought a nack would make PubSub forward a message to the dead letter queue, but instead of short circuits the visibility timeout and allows a message to be redelivered sooner.

facilitator/src/bin/facilitator.rs
```diff
@@ -321,6 +324,47 @@ impl<'a, 'b> AppArgumentAdder for App<'a, 'b> {
...
+                ),
+        )
+        .arg(
+            Arg::with_name("pubsub-api-endpoint")
+                .long("pubsub-api-endpoint")
+                .env("PUBSUB_API_ENDPOINT")
+                .help("API endpoint for GCP PubSub")
+                .long_help(
+                    "API endpoint for GCP PubSub. Optional. Uses \
+                    https://pubsub.googleapis.com by default.",
```
jsha, line 362: It would be good to express this using `.default_value(...)`

facilitator/src/task/pubsub.rs
```diff
@@ -0,0 +1,187 @@
...
+            // subscription, the PubSub API will wait about 90 seconds to send
+            // an HTTP 200 with an empty JSON body. We set a higher read timeout
+            // than usual to allow for this.
+            .timeout_connect(10_000) // ten seconds
+            .timeout_read(180_000) // three minutes
+            .set("Content-Type", "application/json")
+            .send_json(ureq::json!({
+                // Dequeue one task at a time
+                "maxMessages": 1
+            }));
```
jsha: This repeated block (shared with function below) should be factored out into `self.send_json(read_timeout, json)`

facilitator/src/task.rs
```diff
@@ -0,0 +1,91 @@
...
+use serde::Deserialize;
+use std::fmt::Debug;
+
+pub use pubsub::PubSubTaskQueue;
+
+/// Represents a task that can be assigned to a worker
+pub trait Task: Debug + Sized {
+    /// Attempts to deserialize a task structure from the provided JSON string
+    fn from_json(json: &[u8]) -> Result<Self>;
+}
```
jsha: Like in Go, I think you can get want you want from pre-existing types:

```rust
pub trait Task: Debug + Sized + serde::de::DeserializeOwned {
}
```

Then you can take advantage of `from_reader`: https://docs.serde.rs/serde/de/trait.DeserializeOwned.html

And `&[u8]` implements Read: https://doc.rust-lang.org/std/primitive.slice.html#impl-Read

facilitator/src/task.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// Once the task has been successfully completed, the TaskHandle should be
+    /// passed to acknowledge_task to permanently remove the task. If
+    /// acknowledge_task is never called, the task will eventually be
+    /// re-delivered via dequeue().
+    fn dequeue(&mut self) -> Result<Option<TaskHandle<T>>>;
+
+    /// Signal to the task queue that the task has been handled and should be
+    /// removed from the queue.
+    fn acknowledge_task(&mut self, handle: TaskHandle<T>) -> Result<()>;
+}
```
jsha: Style-wise, I generally put traits at the top of the file so they're the first thing people read. That's particularly useful when the trait holds the relevant doccomments.

jsha (commented):
Reviewing the Rust part of this now. It's very nice!

facilitator/src/bin/facilitator.rs
```diff
@@ -857,24 +820,60 @@ fn intake_batch_subcommand(sub_matches: &ArgMatches) -> Result<(), anyhow::Error
...
+                    intake_finished.with_label_values(&["success"]).inc();
+                    queue.acknowledge_task(task_handle)?;
+                }
+                Err(err) => {
+                    error!("error while processing task {:?}: {:?}", task_handle, err);
+                    intake_finished.with_label_values(&["error"]).inc();
+                }
+            }
+
+            // Once we've handled teh task, whether it succeeds or fails, we
```
jsha, line 861: *the

facilitator/src/bin/facilitator.rs
```diff
@@ -1064,32 +1043,67 @@ fn aggregate_subcommand(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
...
+                        prometheus::labels! {},
+                        pushgateway,
+                        prometheus::gather(),
+                        None,
+                    )
+                    .map(|_| info!("done pushing metrics"))
+                    .map_err(|e| error!("error pushing metrics: {}", e))
+                    .ok();
+                }
+            }
```
jsha, line 1108: This block, which is duplicated between the intake-batch and aggregate workers, should be factored out into a function.

facilitator/src/bin/facilitator.rs
```diff
@@ -1295,3 +1308,90 @@ fn transport_for_path(
...
+    }
+    base64::decode(s).context("decoding key from base64")
+}
+
+// You can't make a trait object out of a trait that is generic in another trait
+// (as would be the case for TaskQueue<T: Task>) because such traits are not
+// "object safe" [1], so we can't write a function like
+// fn task_queue_from_args<T: Task>() -> Result<Box<dyn TaskQueue<T>>>.
+// To work around this we manually provide specializations on
+// task_queue_from_args for IntakeBatchTask and AggregationTask.
```
jsha, line 1328: Since you're providing specializations, do you really need the trait object? Seems like you could return `Result<Box<TaskQueue<IntakeBatchTask>>>` for this one, and `Result<Box<TaskQueue<AggregateTask>>>` for the other.

I think you could collapse these two into a single implementation that way:

```rust
fn task_queue_from_args<T>(matches: &ArgMatches) -> Result<Box<TaskQueue<T>>>
```

jsha (changes requested):

facilitator/src/bin/facilitator.rs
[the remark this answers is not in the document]
hostirosti: Correct nack is a shortcut to have it redelivered and retried without having to wait for the Ack timeout. The 7 days retention is if you have replay enabled or if your workers are so busy that they don't get to it earlier. So there is still a value of having a retention policy.

facilitator/src/bin/facilitator.rs
tgeoghegan: I don't think I understand your comment. I _do_ return `Result<Box<dyn TaskQueue<IntakeBatchTask>>>` from the one and `Result<Box<dyn TaskQueue<AggregationTask>>>` from the other. And I tried to write almost exactly the function signature you suggest, though with `Result<Box<dyn TaskQueue<T>>>` and `T: Task`. You can't do it, because `TaskQueue` is not an object safe trait. I'm pretty dumb, though, so maybe there's some easy way to do it that I missed.

facilitator/src/bin/facilitator.rs
tgeoghegan: Elsewhere we discussed moving the facilitator to a pull/scrape model for Prometheus metrics, so rather than refactor this and test it, I'm just going to delete it, since that's what the subsequent change to add a metrics HTTP endpoint would do anyway.

facilitator/src/gcp_oauth.rs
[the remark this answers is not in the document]
tgeoghegan: Good idea though: #296

facilitator/src/task/pubsub.rs
tgeoghegan: Done; I added the `nackowledge_task` method to the `TaskQueue` trait and implemented it here and in `AwsSqsTaskQueue`.

facilitator/src/bin/facilitator.rs
jsha: I'm saying: I don't think you need the `dyn` (which makes it a trait object). And actually you may not need the `Box` either, if it was only there for the purpose of making a trait object.

jsha: Oops, sorry, I realized my mistake here. What makes this a trait object is that `TaskQueue` is a trait, not the `dyn`. The `dyn` is just syntactic sugar (gradually becoming required) to emphasize to the reader that `TaskQueue` is a trait and therefore we're using a trait object here.

So, code is fine as it stands and now I understand what you were getting at. Sorry for the misleading / confusing comments!

tgeoghegan: No worries, this is actually quite confusing! I also tried to ameliorate this by changing `TaskHandle` to not have any type parameters, and use `Task` as a trait object, but that didn't work out because then you can't get a trait object from a `Sized`, and if you're not `Sized`, you can't implement `serde::de::DeserializedOwned`... kind of a mess really, so I'm happy to live with these two ugly functions.

facilitator/src/task/pubsub.rs
tgeoghegan: I refactored this into the `send_json_request` function in the `http` module. There's more boilerplate around `ureq` usage that could be cleaned up, but I think step #1 there is to move to `ureq` > 2.0.0, and I don't want to do that in this PR, so I filed #297

aaomidi (approved):
:shipit:

facilitator/src/gcp_oauth.rs
```diff
@@ -260,16 +262,19 @@ impl OauthTokenProvider {
...
-            .timeout_connect(10_000) // ten seconds
-            .timeout_read(10_000) // ten seconds
-            .set("Authorization", &auth)
-            .set("Content-Type", "application/json")
-            .send_json(ureq::json!({
+        let request = ureq::post(&request_url)
+            .set(
+                "Authorization",
+                &format!("Bearer {}", self.ensure_default_account_token()?),
+            )
```
jsha, line 269: Can this be passed as `token_provider: self`?

facilitator/src/task.rs
```diff
@@ -1,22 +1,39 @@
...
+    /// re-delivered via dequeue().
+    fn dequeue(&mut self) -> Result<Option<TaskHandle<T>>>;
+
+    /// Signal to the task queue that the task has been handled and should be
+    /// removed from the queue.
+    fn acknowledge_task(&mut self, handle: TaskHandle<T>) -> Result<()>;
+
+    /// Signal to the task queue that the task was not handled and should be
+    /// retried later.
+    fn nacknowledge_task(&mut self, handle: TaskHandle<T>) -> Result<()>;
```
jsha, line 31: Love this name :-D

jsha (approved):
Looks good!

facilitator/src/gcp_oauth.rs
tgeoghegan: No, because here we want to call `ensure_default_account_token()` and not `ensure_oauth_token()` as most callers of `send_json_request()`. In fact passing in `token_provider: self` would cause an infinite recursion! Since this is the only callsite that wants this behaviour, I think it's easier to add the special logic here instead of making `send_json_request()` more complicated.

facilitator/src/task.rs
hostirosti: It's easy to mix up though, ack vs nack is easy to distinguish, acknowledge vs nacknowledge less so.

workflow-manager/bucket/bucket.go
```diff
@@ -58,51 +64,61 @@ func (b *Bucket) ListFiles(ctx context.Context) ([]string, error) {
...
-	stsSTS := sts.New(sess)
-	roleSessionName := ""
-	roleProvider := stscreds.NewWebIdentityRoleProviderWithToken(
-		stsSTS, identity, roleSessionName, tokenfetcher.NewTokenFetcher(audience))
-
-	return credentials.NewCredentials(roleProvider), nil
+	return s3.New(sess, config), nil
 }
 
 func (b *Bucket) listFilesS3(ctx context.Context) ([]string, error) {
```
hostirosti, line 104: ctx doesn't seem to be used --> drop

workflow-manager/bucket/bucket.go
```diff
@@ -123,7 +139,40 @@ func (b *Bucket) listFilesS3(ctx context.Context) ([]string, error) {
 	return output, nil
 }
 
-func (b *Bucket) listFilesGS(ctx context.Context) ([]string, error) {
+func (b *Bucket) writeTaskMarkerS3(ctx context.Context, marker string) error {
```
hostirosti, line 142: ctx doesn't seem to be used --> drop

workflow-manager/bucket/bucket.go
```diff
@@ -123,7 +139,40 @@ func (b *Bucket) listFilesS3(ctx context.Context) ([]string, error) {
...
+
+	// Deliberately ignore the result, we only care if the write succeeds
+	if _, err := svc.PutObject(input); err != nil {
+		return fmt.Errorf("storage.PutObject: %w", err)
+	}
+
+	return nil
+}
+
+func (b *Bucket) gcsClient(ctx context.Context) (*storage.Client, error) {
```
hostirosti, line 175: I'd drop passing ctx here and use 
```
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
defer cancel()
```
to get the context for obtaining the client. The NewClient context is just needed for the client creation, for subsequent calls you use a new / different context.

workflow-manager/bucket/bucket.go
```diff
@@ -132,17 +181,26 @@ func (b *Bucket) listFilesGS(ctx context.Context) ([]string, error) {
 		return nil, fmt.Errorf("storage.newClient: %w", err)
 	}
 
+	return client, nil
+}
+
+func (b *Bucket) listFilesGS(ctx context.Context) ([]string, error) {
+	client, err := b.gcsClient(ctx)
```
hostirosti, line 188: I recommend creating a GCSService calls that lives across the lifetime of the process and holds the GCS client. This way you don't need to create a new client for every GCS call (which is inefficient). The client can be long lived.

workflow-manager/bucket/bucket.go
```diff
@@ -163,3 +221,37 @@ func (b *Bucket) listFilesGS(ctx context.Context) ([]string, error) {
 
 	return output, nil
 }
+
+func (b *Bucket) writeTaskMarkerGS(ctx context.Context, marker string) error {
+	client, err := b.gcsClient(ctx)
```
hostirosti, line 226: same as above.

workflow-manager/bucket/bucket.go
```diff
@@ -163,3 +221,37 @@ func (b *Bucket) listFilesGS(ctx context.Context) ([]string, error) {
...
+	log.Printf("writing task marker to gs://%s/%s as (ambient service account)",
+		b.bucketName, marker)
+
+	if b.dryRun {
+		log.Printf("dry run, skipping marker write")
+		return nil
+	}
+
+	object := bkt.Object(marker)
+	writer := object.NewWriter(ctx)
```
hostirosti, line 242: Create a specific context for this operation
```
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
defer cancel()
```

workflow-manager/main.go
```diff
@@ -177,29 +316,27 @@ func main() {
...
-		ShouldRequeue: wferror.IsTransientErr,
-		// 5 is an arbitrary number... it seemed right
-		MaxTries: 5,
-		// No need to wait between retries
-		TimeBetweenTries: 0,
-	}
-
-	if err := r.Start(); err != nil {
+	err = enqueueAggregationTasks(
+		context.Background(),
```
hostirosti, line 324: You can get the Background context anywhere, no need to pass per fn param. Also keep the context as narrow as possible and required

workflow-manager/main.go
```diff
@@ -288,290 +425,162 @@ func groupByAggregationID(batches batchpath.List) aggregationMap {
...
-			return fmt.Errorf("starting job for batch %s: %w", batch, err)
-		}
-		intakesStarted.Inc()
-	}
 	return nil
 }
 
-func startIntakeJob(
+func enqueueIntakeTasks(
 	ctx context.Context,
```
hostirosti, line 519: Remove ctx from the param list

workflow-manager/task/task.go
```diff
@@ -0,0 +1,253 @@
...
+	topic     *pubsub.Topic
+	waitGroup sync.WaitGroup
+	dryRun    bool
+}
+
+// NewGCPPubSubEnqueuer creates a task enqueuer for a given project and topic
+// in GCP PubSub. If dryRun is true, no tasks will actually be enqueued. Clients
+// should re-use a single instance as much as possible to enable batching of
+// publish requests.
+func NewGCPPubSubEnqueuer(ctx context.Context, project string, topicID string, dryRun bool) (*GCPPubSubEnqueuer, error) {
```
hostirosti, line 152: remove ctx from fn param list

workflow-manager/task/task.go
```diff
@@ -0,0 +1,253 @@
...
+	client, err := pubsub.NewClient(ctx, project)
```
hostirosti, line 153: Use a narrow ctx here:
```
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
defer cancel()
```

workflow-manager/task/task.go
```diff
@@ -0,0 +1,253 @@
...
+		return nil, fmt.Errorf("pubsub.NewClient: %w", err)
+	}
+
+	return &GCPPubSubEnqueuer{
+		topic:  client.Topic(topicID),
+		dryRun: dryRun,
+	}, nil
+}
+
+func (e *GCPPubSubEnqueuer) Enqueue(ctx context.Context, task Task, completion func(error)) {
```
hostirosti, line 164: You use an external ctx if you want to group multiple operations in one ctx that you can cancel if any of the participants errors and thus release resources in all participating systems. If there is no grouping of operations is required you can create the context here in the function and don't need to pass it as a fn param.

```
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
defer cancel()
```

workflow-manager/task/task.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	return &AWSSNSEnqueuer{
+		service:  sns.New(session, config),
+		topicARN: topicARN,
+		dryRun:   dryRun,
+	}, nil
+}
+
+func (e *AWSSNSEnqueuer) Enqueue(ctx context.Context, task Task, completion func(error)) {
```
hostirosti, line 219: Same as above

hostirosti (changes requested):
The use of context in the go parts needs some cleanup since current use potentially cause unexpected / intended behaviour especially in long lived instances.

tgeoghegan:
> The use of context in the go parts needs some cleanup since current use potentially cause unexpected / intended behaviour especially in long lived instances.

Yes, we can always add context parameters back to these things if at some point in the future we opt to get fancy and build context trees. In the meantime, I'll do the simple thing.

workflow-manager/bucket/bucket.go
```diff
@@ -123,26 +139,68 @@ func (b *Bucket) listFilesS3(ctx context.Context) ([]string, error) {
...
+
+	return nil
+}
+
+func (b *Bucket) gcsClient() (*storage.Client, error) {
 	if b.identity != "" {
 		return nil, fmt.Errorf("workflow-manager doesn't support non-default identity %q for GS Bucket %q", b.identity, b.bucketName)
 	}
-	client, err := storage.NewClient(ctx)
+	client, err := storage.NewClient(context.Background())
```
hostirosti, line 179: technically possible but really would encourage to use a timing out ctx here.
```
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
	defer cancel()
```

workflow-manager/bucket/bucket.go
```diff
@@ -123,26 +139,68 @@ func (b *Bucket) listFilesS3(ctx context.Context) ([]string, error) {
...
+		return nil, err
+	}
+
 	bkt := client.Bucket(b.bucketName)
 	query := &storage.Query{Prefix: ""}
 
 	log.Printf("looking for ready batches in gs://%s as (ambient service account)", b.bucketName)
 	var output []string
-	it := bkt.Objects(ctx, query)
+	it := bkt.Objects(context.Background(), query)
```
hostirosti, line 198: same as above

workflow-manager/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,78 @@
...
+// error on failure.
+func (c *Client) ListJobs() (map[string]batchv1.Job, error) {
+	jobs := map[string]batchv1.Job{}
+
+	// The jobs list API is paginated. We request 1000 entries at a time. If
+	// there are more entries, the response will contain a continue token to
+	// provide on subsequent requests.
+	continueToken := ""
+	for {
+		jobsList, err := c.client.BatchV1().Jobs(c.namespace).List(context.Background(), metav1.ListOptions{
```
hostirosti, line 58: technically possible but really would encourage to use a timing out ctx here.
```
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
	defer cancel()
```

workflow-manager/task/task.go
```diff
@@ -0,0 +1,253 @@
...
+	waitGroup sync.WaitGroup
+	dryRun    bool
+}
+
+// NewGCPPubSubEnqueuer creates a task enqueuer for a given project and topic
+// in GCP PubSub. If dryRun is true, no tasks will actually be enqueued. Clients
+// should re-use a single instance as much as possible to enable batching of
+// publish requests.
+func NewGCPPubSubEnqueuer(project string, topicID string, dryRun bool) (*GCPPubSubEnqueuer, error) {
+	client, err := pubsub.NewClient(context.Background(), project)
```
hostirosti, line 153: technically possible but really would encourage to use a timing out ctx here.
```
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
	defer cancel()
```

workflow-manager/task/task.go
```diff
@@ -0,0 +1,253 @@
...
+			completion(nil)
+			return
+		}
+
+		// Publish() returns immediately, giving us a handle to the result that we
+		// can block on to see if publishing succeeded. The PubSub client
+		// automatically retries for us, so we just keep the handle so the caller
+		// can do whatever they need to after successful publication and we can
+		// block in Stop() until all tasks have been enqueued
+		res := e.topic.Publish(context.Background(), &pubsub.Message{Data: jsonTask})
```
hostirosti, line 185: technically possible but really would encourage to use a timing out ctx here.
```
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
	defer cancel()
``` 

Plus you want to group the two usages of context (Publish + Get)

tgeoghegan:
@hostirosti  I jumped the gun on merging this change, but addressed your comments about timeouts in #304

MODNXS-cell:
> :shipit: 

Hold fast, for surely I come quickly

PR #304, opened 2021-01-08
workflow-manager: add timeouts to context.Contexts

tgeoghegan:
use context.Context to set a 30s timeout on GCP and Kubernetes API
calls.

### DIFF
--- a/workflow-manager/task/task.go
+++ b/workflow-manager/task/task.go
@@ -5,7 +5,6 @@
 package task
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"log"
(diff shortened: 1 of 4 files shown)


aaomidi (approved):

workflow-manager/task/task.go
```diff
@@ -182,8 +186,10 @@ func (e *GCPPubSubEnqueuer) Enqueue(task Task, completion func(error)) {
 		// automatically retries for us, so we just keep the handle so the caller
 		// can do whatever they need to after successful publication and we can
 		// block in Stop() until all tasks have been enqueued
-		res := e.topic.Publish(context.Background(), &pubsub.Message{Data: jsonTask})
-		if _, err := res.Get(context.Background()); err != nil {
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
aaomidi, line 189: I see the 30 * time.Second repeated a bunch, maybe we can clean this up and make it a const.

PR #313, opened 2021-01-15
workflow-manager: relax timeout on GCS object list

tgeoghegan:
We observed workflow-manager failing in the `us-md-apple` instance
because we were hitting timeouts enumerating the contents of GCS
buckets. This is because the `list` API on buckets is paginated
(https://cloud.google.com/storage/docs/json_api/v1/buckets/list), which
the Go SDK handles by returning an `ObjectIterator` from
`BucketHandle.Objects()`. However you can only provide a
`context.Context` to the initial call to `Objects()`, and each
subsequent call to `ObjectIterator.next()` is subject to the same
initial timeout. There's no way to reset the timeout on a
`context.Context` that I can find, so I believe the thing to do is to
raise the timeout for this particular operation, which we understand
actually does take longer than 30 seconds, to 120 seconds.

Additionally, while looking into correct handling of timeouts in GCS, I
found mention in Google documentation that timeouts should not be
applied when calling the various `NewClient()` functions from the GCP
SDK. Accordingly those calls are amended to use `context.Background()`.

### DIFF
--- a/workflow-manager/bucket/bucket.go
+++ b/workflow-manager/bucket/bucket.go
@@ -1,10 +1,12 @@
 package bucket
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"log"
 	"strings"
+	"time"
 
 	leaws "github.com/letsencrypt/prio-server/workflow-manager/aws"
 	"github.com/letsencrypt/prio-server/workflow-manager/utils"
@@ -173,8 +175,9 @@ func (b *Bucket) writeTaskMarkerS3(marker string) error {
 }
 
 func (b *Bucket) gcsClient() (*storage.Client, error) {
-	ctx, cancel := utils.ContextWithTimeout()
-	defer cancel()
+	// Google documentation advises against timeouts on client creation
+	// https://godoc.org/cloud.google.com/go#hdr-Timeouts_and_Cancellation
+	ctx := context.Background()
 
 	if b.identity != "" {
 		return nil, fmt.Errorf("workflow-manager doesn't support non-default identity %q for GS Bucket %q", b.identity, b.bucketName)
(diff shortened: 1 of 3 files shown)


workflow-manager/bucket/bucket.go
```diff
@@ -189,7 +192,9 @@ func (b *Bucket) gcsClient() (*storage.Client, error) {
 }
 
 func (b *Bucket) listFilesGS() ([]string, error) {
-	ctx, cancel := utils.ContextWithTimeout()
+	// This timeout has to cover potentially numerous roundtrips to the
+	// paginated API for listing objects, so we use a longer timeout than usual.
+	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
```
aaomidi, line 197: This is a critical part of the application right? Does having *any* timeout even make sense here? Maybe it's a sign that something is wrong if it fails, but if the application can't reasonably recover from this timeout being cancelled would it make sense to make it even larger to handle variations in networking/downtime of gs? (e.g. something like 10 minutes)

aaomidi (approved):

workflow-manager/bucket/bucket.go
tgeoghegan: We already will retry 10 minutes later because `workflow-manager` will run again. The point of a timeout here is so that we don't hang forever if the GCS server just never replies but also never closes the connection: we'd rather fail obviously with an error than just have a particular run of the task never exit.

jsha: I agree we can afford to make this higher. That said, I also think it's time to change how we list files so we list by a pattern that matches the things we're looking for, so we don't have to paginate through so much unwanted stuff.

jsha (approved):

tgeoghegan:
I bumped the timeout to 10 minutes, just in case we get buckets with perversely large object counts. So long as we have _some_ timeout value, we'll catch workflow-manager jobs that get stuck.

PR #316, opened 2021-01-16
Limiter signal channel

aaomidi:


tgeoghegan (changes requested):
This logic should be implemented within `GCPPubSubEnqueuer`, since (1) only its implementation is actually asynchronous and (2) you shouldn't need to repeat the logic at all callsites to `Enqueue()`. Also, can you provide a link to a reference describing this pattern? I'm not familiar with it, and it'd be good to include a comment explaining what this does, why, and how.

aaomidi:
I didn't follow any specific pattern with this PR and wanted to quickly implement some basic queueing and ratelimits. I don't think this is the optimal way of doing this. I will come up with a better way of doing this.

Alongside ratelimiting this, I'm going to take a look at why this enqueue was making us run out of memory. I think the library we're using potentially has a goroutine leak somewhere.

aaomidi:
Going to close this PR and write a better one

jsha:
A goroutine leak wouldn't explain this because workflow-manager isn't a long-lived process. It runs once through and exits.

Also a reminder that goroutines are very cheap - about 2.7kB each. So if we're running out of memory with 10k goroutines, it's probably because we have some per-goroutine data that's largeish (maybe a GCP client?). The effect is the same, of course, but I think it's useful to be cautious about terminology, since "goroutines are cheap" is one of the core working concepts of Go.

Go has some nice built-in memory profiling that I'm happy to help walk through sometime.

PR #321, opened 2021-01-19
Create a work limiter that limits intake and aggregation task creation

aaomidi:
After merging #313 we noticed that our worker pods are running out of memory. It seems like fixing our GCS listing code caused us to create so many goroutines that we were regularly getting OOMed.

This PR intends to provide a long term solution that can be used by workflow-manager and other packages for limiting the creation of resources based upon a number of "workers".

I want to make sure we're fine with the generation philosophy on how this works before I finish the TODO tasks.

TODO:
- [x] Make the worker count something we can configure with runtime parameters
- [x] Expose the number of tickets in progress for reporting purposes - maybe this can be done with a simple len() call on the workers channel
- [x] Probably write some tests for this


tgeoghegan:
Porting over [some discussion from the previous PR](https://github.com/abetterinternet/prio-server/pull/316#issuecomment-763130780):

>Also a reminder that goroutines are very cheap - about 2.7kB each. So if we're running out of memory with 10k goroutines, it's probably because we have some per-goroutine data that's largeish (maybe a GCP client?). The effect is the same, of course, but I think it's useful to be cautious about terminology, since "goroutines are cheap" is one of the core working concepts of Go.

I'm not 100% certain what Go's closure capture semantics are (when does it capture a reference vs. a copy of objects), but off the top of my head, the other per-Goroutine allocations would be the JSON encoding of a task (which should just be three ish lines of text, so maybe a few hundred bytes per goroutine). I agree we should profile to get a better idea of where the memory footprint is.

workflow-manager/limiter/limiter.go
```diff
@@ -0,0 +1,69 @@
...
+	worker := <-l.workers
+	return &Ticket{
+		limiter: l,
+		worker:  worker,
+		dirty:   int32(0),
+	}
+}
+
+// Execute Executes a function with the work ticket
+func (t *Ticket) Execute(job func()) error {
```
tgeoghegan, line 45: Am I ever going to call `RequestTicket()` without immediately calling `Execute()`? I think you could simplify the interface by having a single method `limiter.Execute(job func()) (*Ticket, error)` that blocks on an available worker (by reading from `l.workers`) and then executes the provided `job`. Then you don't need the atomic test on `ticket.Dirty` because there is no way to call `Execute()` twice on a single `Ticket`.

workflow-manager/limiter/limiter.go
```diff
@@ -0,0 +1,69 @@
...
+
+	atomic.AddInt32(&t.dirty, 1)
+	atomic.AddInt32(&t.limiter.inProgress, 1)
+	job()
+
+	return nil
+}
+
+// CompleteWork signals that the execution for this particular task is complete
+func (t *Ticket) Done() error {
```
tgeoghegan, line 58: nit/subjective: something feels odd to me about `Ticket` having a `Limiter` member since you get a `Ticket` from a `Limiter`, so it feels like parent/child relationship is being inverted here. I wonder if having a method `(l *Limiter) Done(ticket *Ticket) error` would work better?

workflow-manager/limiter/limiter.go
```diff
@@ -0,0 +1,69 @@
...
+// Ticket is the structure that allows you to schedule jobs
+type Ticket struct {
+	limiter *Limiter
+	worker  int32
+	dirty   int32
+}
+
+// New creates a new instance of Limiter that limits how many functions can get executed without the other ones completing first
+func New(maxWorkers int32) *Limiter {
+	l := &Limiter{make(chan int32, maxWorkers), 0}
```
tgeoghegan, line 24: I'm dumb and concurrency is hard -- I think this merits a comment explaining that Go channels created with a capacity are blocking, and that we rely on this behavior to make the read from `l.workers` in `RequestTicket()` block until a worker is available.

workflow-manager/limiter/limiter.go
```diff
@@ -0,0 +1,69 @@
...
+
+	for i := int32(0); i < maxWorkers; i++ {
+		l.workers <- i
```
tgeoghegan, line 27: IIUC, it doesn't matter what values we fill the channel with so long as there are `maxWorkers` of them. Do we fill them with distinct values just so we could someday log what worker a ticket is using?

workflow-manager/limiter/limiter.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"fmt"
+	"sync/atomic"
+)
+
+// Limiter is similar to sync.Pool, however, it limtis how many shared
+// objects (in this case workers) there can be
+type Limiter struct {
+	workers    chan int32
+	inProgress int32
```
tgeoghegan, line 12: We update `inProgress`as tickets come and go, but never use it, right? Is this here just so we could someday log or monitor how many tickets are in flight?

workflow-manager/limiter/limiter.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
+
+// Execute Executes a function with the work ticket
+func (t *Ticket) Execute(job func()) error {
+	if atomic.LoadInt32(&t.dirty) > 0 {
+		return fmt.Errorf("did not schedule job - ticket was dirty")
+	}
+
+	atomic.AddInt32(&t.dirty, 1)
```
tgeoghegan, line 50: You should define an enum for the possible values of `dirty` that make it clear what they mean (e.g., `0 = idle, 1 = working, 2 = done`), and then use `atomic.StoreInt32` to set the state to exactly what you want.

workflow-manager/limiter/limiter.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// CompleteWork signals that the execution for this particular task is complete
+func (t *Ticket) Done() error {
+	if atomic.LoadInt32(&t.dirty) > 1 {
+		return fmt.Errorf("did not complete job - job already completed")
+	}
+	// Add enough to be considered completed
+	atomic.AddInt32(&t.dirty, 2)
+	atomic.AddInt32(&t.limiter.inProgress, -1)
```
tgeoghegan, line 64: I'm concerned when I see three atomic operations in a row, because while each read or write will be atomic with respect to each other or with respect to other calls to `Done()`, I suspect your critical section is actually the entire method. Consider this sequence of operations:

- Thread 1 and thread 2 share a reference to a `Ticket` which has `dirty = 1` and `worker = 12`
- Thread 1 calls `ticket.Done()`, gets past the atomic test on line 59, then gets preempted before it can execute line 63
- Thread 2 calls `ticket.Done()`, gets past the atomic test on line 59 (because thread 1 hasn't yet incremented `ticket.Dirty`), then continues to the end of the method, sending `12` into `ticket.limiter.workers`, then gets preempted
- Thread 1 runs again, and executes the remainder of `Done()`, also sending `12` into `ticket.limiter.workers`.

So `t.limiter.workers` will have a duplicate value sent to it, and then eventually some other call to `Done()` will block forever because there's an extra value in the `workers` channel.

I think what you need here is to add a `sync.Mutex` member to the `Limiter` and use it to protect the entire body of this method, since the whole thing is the critical section. I think that will be more natural to do if you refactor `Done()` to be a method on `Limiter` instead of `Ticket`.

tgeoghegan (changes requested):

workflow-manager/limiter/limiter.go
aaomidi: Hmm, yeah. I kinda feel like that makes more sense but at the same time it feels like the "Ticket" is done. Maybe it would make more sense to pass the worker back to the limiter on some sort of channel.

workflow-manager/limiter/limiter.go
aaomidi: It just seemed like the simplest way to fill them up and giving each ticket a distinct value seemed like a nice design choice.

workflow-manager/limiter/limiter.go
aaomidi: Good call!

workflow-manager/limiter/limiter.go
aaomidi: Yeah, it's so we could potentially use them for debugging/monitoring.

workflow-manager/limiter/limiter.go
aaomidi: I think I can use atomic.CompareAndSwapInt32 to avoid using locks

tgeoghegan: Why avoid locks though? This isn't performance critical and I think it'll be much, much easier to get this right with a mutex.

jsha: I'm +1 for just using a lock.

aaomidi: But what would I be locking now? Doesn't the atomic function remove the need for it?

tgeoghegan: Atomics are harder to get right and harder to reason about, making it harder to maintain this code in the future. Note that in your most recent version, you have to add comments indicating which lines are and are not in the critical section. This would be vastly simpler if it just used a mutex to protect the entire method body, and in general we ought to aim to do the simplest thing that will solve the problem.

workflow-manager/limiter/limiter.go
tgeoghegan: Yep, just making sure I understand the significance of the values in the channel.

workflow-manager/limiter/limiter.go
```diff
@@ -0,0 +1,73 @@
...
+	job()
+	return nil
+}
+
+// Done signals that the execution for this particular task is complete
+// This call will fail if the Ticket has not been executed
+func (l *Limiter) Done(ticket *Ticket) error {
+	if !atomic.CompareAndSwapInt32(&ticket.status, working, done) {
+		return fmt.Errorf("did not complete job - job was either already completed or ticket was idle")
+	}
```
tgeoghegan, line 67: Another advantage of using a `sync.Mutex` is that `Mutex.Lock` cannot fail, which means that `Done()` would no longer need to return error and it would absolve its callers from doing error handling.

workflow-manager/limiter/limiter.go
tgeoghegan: I see your point. I would do `func (l *Limiter) Execute(job func(*Ticket)) error`, with the contract that `job` (or some other goroutine started by `job`) is expected to call `ticket.Done()`.

workflow-manager/limiter/limiter.go
aaomidi: Ahhh I see, fair!

jsha:
Remember that `chan` can hold any value. For instance, you could send a `Ticket` through a chan. So, one fun way to model this would be that the `Limiter` has a `chan Ticket`, which starts with N tickets. Tickets get pulled off the chan to run a task, and the same ticket gets put back on the chan when the task is done. This would be particularly sweet in Rust, where we could use ownership semantics to ensure that the caller truly "gave back" ownership of the Ticket. 

Note that with the current code, you'd have to "reset" the Ticket (set its status to done) before putting it back on the chan. But that leads me to the question: why does a Ticket need to track its status? Seems like it's to prevent double spending.

Are there simpler ways we can prevent double spending? One way would be to eliminate the round trip where the caller gets a ticket, then spends a ticket. Instead, the caller could give Limiter a pile of funcs and say "I want you to run these, but with a max concurrency of N. Tell me when you're done."

That would also have the benefit that you could encapsulate the `WaitGroup` in `Limiter` and calling code would have one less thing to worry about.

There are a few different ways to write this interface. The simplest would be to take a slice of funcs and have a blocking call that returns when all the work is done:

```go
func (l *Limiter) Run([]func())
```

I think that's perfectly adequate for this use case, and simplicity is very valuable. There are some cases where you might not know all the items up front, or might not want to build the whole slice at once. Then you could do:

```go
// Run a function concurrently.
// If there are <N functions running, Run will return immediately.
// If there are N functions running, Run will block until one finishes.
// Must not be called after `Finish()`.
func (l *Limiter) Run(func())
// Finish blocks until all currently running functions return.
func (l *Limiter) Finish()
```

Internally, Limiter would have a WaitGroup. It would look something like:

```
func (lim *Limiter) Run(f func()) {
  lim.start()
  go func() {
    defer lim.done()
    f()
  }()
}

func (lim *Limiter) start() {
  <-lim.workers
  lim.wg.Add(1)
}

type token struct{}

func (lim *Limiter) done() {
  lim.workers<-token{}
  lim.wg.Done()
}

func (lim *Limiter) Finish() {
  lim.wg.Wait()
}
```

Note that I also added a new `type token` in place of the worker ID that you're currently using in the channel. There are two ideas here: One is "You Ain't Gonna Need It (YAGNI)." If you're not using the worker ID concept right now, you may never use it. Better to omit it and then add it later when you actually need it, and have a better idea of the requirements.

The other idea is that types can be a form of documentation. By having a specific token type, people can see that it's not used anywhere else, and can see its possible semantics. So, for instance, the reader doesn't have to think about what happens if you run out of worker ids, or reuse a worker id.

jsha (changes requested):

workflow-manager/limiter/limiter.go
aaomidi: Huh actually with the changes to how tickets are done, I don't need locking or atomics at all. Technically I'm using atomics for inProgress stuff, but it seems like that's a non-critical part of the code.

jsha: Hah, yep. I wrote a whole novel below but you summed it up in a sentence. :-D

BTW I recommend pulling out the `inProgress` stuff. You're not using it right now, and "YAGNI" applies. Also, workflow-manager only pushes metrics when it's done with everything, so even if you added metrics export for that, you'd never see a nonzero `inProgress`.

aaomidi: I actually **really** enjoyed reading that - it's definitely a pattern I'm going to consider in the future. Maybe if we realize we need that in some part of workflow manager.

Concurrency is always difficult and I love reading and learning about the patterns others come up with. Going to definitely bookmark this one.

workflow-manager/limiter/limiter.go
aaomidi: Yesss! This was really clean.

aaomidi:
Realized a small issue that Done() can be called multiple times. Will push a fix out for that and address all the comments.

workflow-manager/limiter/limiter.go
```diff
@@ -0,0 +1,56 @@
...
+// Ticket is the structure that allows you to schedule jobs
+type Ticket struct {
+	worker int32
+	lock   sync.Mutex
+	dirty  bool
+}
+
+// New creates a new instance of Limiter that limits how many functions can get executed without the other ones completing first
+func New(maxWorkers int32) *Limiter {
+	// Channels are blocking in go, if you write to them, something needs to read from them
```
jsha, line 20: This is a [comma splice](https://www.grammarly.com/blog/comma-splice/).

```suggestion
	// Channels are blocking in go. If you write to them, something needs to read from them.
```

workflow-manager/limiter/limiter.go
```diff
@@ -0,0 +1,56 @@
...
+	// go allows us to create buffered channels, so that reading and writing from them is not blocking
+	// when the channel has items in the buffer, or has space in the buffer respectively
+	// using this feature of channels we can create a worker pool to execute tasks while limiting
+	// the resource creation that happens with each task
```
jsha, line 24: Nitpicking this comment a bit: The comment spends four lines explaining how a Go feature works, but doesn't actually explain how this code takes advantage of the feature. The comment makes sense now, while you're learning about channels, but when you come back in three months you'll wonder why you were explaining channels. I would recommend linking to https://golang.org/ref/spec#Channel_types for the details, and spending these lines explaining the _how_ of this code.

An example of a _how_ comment:

We use a buffered channel of N elements, and fill it with N tokens. Before running, a task must read a token from the channel, and after running, a task must write a token back to the channel. Since there are only N tokens, we guarantee that any tasks trying to start while there are already N tasks running will block until one of those tasks returns its token.

(Also, I prefer capitalization and periods in all comments that are full sentences. Otherwise it's hard to tell where one sentence ends and the next begins).

jsha (changes requested):

jsha (approved):

workflow-manager/task/task.go
```diff
@@ -163,41 +165,45 @@ func NewGCPPubSubEnqueuer(project string, topicID string, dryRun bool) (*GCPPubS
...
-		jsonTask, err := json.Marshal(task)
-		if err != nil {
-			completion(fmt.Errorf("marshaling task to JSON: %w", err))
-			return
-		}
-
-		if e.dryRun {
-			log.Printf("dry run, not enqueuing task")
+	e.limiter.Execute(func(ticket *limiter.Ticket) {
+		defer e.limiter.Done(ticket)
```
tgeoghegan, line 176: This function (the one passed to `Execute()` will return immediately after creating the goroutine on line 178, and so the ticket will be returned to the limiter before the potentially expensive work is done. I think this should be called inside the closure that is created on line 178.

aaomidi: Oh I completely didn't notice the goroutine there. Thanks!

hostirosti:
This seems like an overly complicated implementation to solve the issue. This can be done with simple channel semantics. You can basically us a buffered channel that accepts up to a certain number of pubsub publish requests, and then have a number of goroutines that pick from that channel. The enqueuer would publish to that channel, to save memory you could keep the buffer small and use cursors to retrieve the object listing from S3 or if it's not too much data load it into a lean data structure in memory and then publish to the channel from there.

aaomidi:
@hostirosti that's right - I was aiming for something more generic that could potentially be used in other prio go projects as we're already sharing modules between projects right now.

However it should be pretty simple to implement what you're suggesting. @tgeoghegan let me know if you'd prefer to go in that direction here.

workflow-manager/limiter/limiter_test.go
```diff
@@ -0,0 +1,42 @@
...
+	// This shouldn't be something we do in actual code but for test cases...eh it's fine?
+causality:
+	for {
+		select {
+		case <-counterChannel:
+			counter++
+		case <-time.After(time.Second):
+			break causality
+		}
+	}
```
tgeoghegan, line 37: I worry that this test is fragile due to the timing issue you mention, forces a one second delay in the test suite (these things add up as test suites grow) and doesn't prove that the remaining five jobs ever get to execute once the first five complete.

[Here's a Go playground](https://play.golang.org/p/UnMlkPcyg0U) that I think implements a more thorough test in its `main()` function. Note that we use signaling between the function and job goroutines to guarantee that we're blocked on the ticket from the first job being returned, and then verify that the other two jobs get to run after the first completes. There is still a 10 millisecond wait (`case <-time.After(10 * time.Millisecond)` on line 152) that we use to introduce a window of time during which we don't expect the second and third jobs to run, but I'm not sure how to get rid of that without doing some kind of Goroutine inspection that I don't know how to do.

I left in a ton of comments and `log` statements to make it obvious what is happening; you may wish to delete those if you opt to incorporate my test.

workflow-manager/limiter/limiter_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"time"
+)
+
+func TestTicketDoubleDone(t *testing.T) {
+	l := New(5)
+	l.Execute(func(ticket *Ticket) {
+		l.Done(ticket)
+		l.Done(ticket)
+	})
+	t.Log("l.Execute did not block")
```
tgeoghegan, line 14: How does a failure in this test manifest? The Go deadlock detector will eventually fire and abort the test? Does that let the rest of the test suite go through? I think it'd be better to do something like:
```
l := New(5)
doneChan := make(chan struct{})
go l.Execute(func(ticket *Ticket) {
    l.Done(ticket)
    l.Done(ticket)
    doneChan <- struct{}{}
})
for {
    select {
        case <-doneChan:
            // success
            return
        case <-time.After(time.Second):
            t.Fatal("timeout waiting for job to complete")
    }
}
```

tgeoghegan (changes requested):
I think Robert is right that it would be simpler to have put a channel in `GCPPubSubEnqueuer`, and [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) would dictate that we shouldn't build a generic solution to a problem until and unless the problem exists in more than one place. However I don't think `Limiter` is so terribly complicated that its existence imposes significant overhead, and having it separated out makes it easier to test, so I'm comfortable with what's here. I did leave some comments about test coverage and doccomments.

tgeoghegan (approved):
Sorry, I hadn't noticed this had new changes since I last reviewed. LGTM, though consider rebasing the change into a single commit before merging.

PR #324, opened 2021-01-20
[AWS] added retry logic to reach k8s api

hostirosti:
We are seeing a race condition in Fargate backed AWS EKS clusters when trying to reach the k8s api for listing jobs by the `workflow-manager`. To workaround this network path availability issue I added a retry logic. This is untested as of now.


workflow-manager/kubernetes/kubernetes.go
```diff
@@ -3,6 +3,11 @@ package kubernetes
 
 import (
 	"fmt"
+	"net"
+	"os"
+	"time"
+
+	"gopkg.in/retry.v1"
```
tgeoghegan, line 10: Good find, I wish we had found this package earlier! Would you mind deleting the (unused) `workflow-manager/retry` package in this PR?

workflow-manager/kubernetes/kubernetes.go
```diff
@@ -25,6 +30,29 @@ type Client struct {
 // it is not empty. If dryRun is true, then any destructive API calls will be
 // made with DryRun: All.
 func NewClient(namespace string, kubeconfigPath string, dryRun bool) (*Client, error) {
+	// Making sure endpoint is available before returning kubernetes client
```
tgeoghegan, line 33: Instead of polling for availability here, I'd prefer if we retried the call to `c.client.BatchV1.Jobs().List()` in `LostJobs()`, though only if the call fails due to the context deadline being exceeded. I think that would make us resilient to more kinds of failures, and also would rely less on implementation details of the Kubernetes client library.

workflow-manager/kubernetes/kubernetes.go
```diff
@@ -25,6 +30,29 @@ type Client struct {
...
+		)
+		timeout := time.Duration(1 * time.Second)
+		for a := retry.Start(strategy, nil); a.Next(); {
+			if _, err := net.DialTimeout("tcp", host+":"+port, timeout); err == nil {
+				available = true
+				break
+			}
+		}
+		if !available {
+			panic("InCluster k8s api endpoint not available")
```
tgeoghegan, line 52: This should return an error, so that it gets propagated up to `main()` and metrics can be incremented appropriately as in other failure modes.

tgeoghegan (changes requested):
Thank you for the pull request, @hostirosti! I left some comments. I'm happy to make these changes myself if you prefer.

workflow-manager/kubernetes/kubernetes.go
hostirosti: yes, can do.

workflow-manager/kubernetes/kubernetes.go
hostirosti: I see a challenge if we do it LostJobs() 
(1) this is purely to discover if the network is up and this needs to be done only once (when we get the client)
(2) The behaviour I discovered is that the call times out even if the network becomes available within the 30sec of the context deadline. Meaning we should use different timeouts which makes this complex and hard to maintain or very slow as an alternative. I'd prefer leaving this here. We can add more retry logic independently if to LostJobs if needed.

workflow-manager/kubernetes/kubernetes.go
hostirosti: Do we have an existing error that I can throw here or do you want to create a new one?

workflow-manager/kubernetes/kubernetes.go
tgeoghegan: Robert convinced me in Slack that since we'll soon delete this module anyway (because `workflow-manager` won't need to list Kubernetes jobs at all once the migration from the old mode is complete), we can live with this as it is.

tgeoghegan (approved):

PR #329, opened 2021-01-21
Scrape basic faciliator metrics

tgeoghegan:
This PR sets up basic metrics delivery from the `facilitator` for the `intake-batch` and `aggregate` deployments.

- I used [`warp`](https://crates.io/warp) to implement the metrics endpoint, which required spawning a task into a `tokio::runtime::Runtime`. `tokio` just released their 1.0 crate, so I updated our dependency, and also updated a number of other dependencies (such as `hyper` and `rusoto`) to move to versions which use `tokio` 1.0. We used to have several versions `hyper`, `rustls`, `tokio` and other dependencies compiled into the crate, but now we pull in far fewer distinct versions.
- When the `intake-batch-worker` or `aggregate-worker` subcommands are invoked, `facilitator` now listens on a configurable port for GET requests to the `/metrics` endpoint
- In Terraform, we now create a Kubernetes _service_ for each _deployment_. The services are annotated to allow Prometheus to discover metrics scrape targets.
 
Now, we can `kubectl port-forward` as instructed in `terraform/README/md` to get to the Prometheus console and BEHOLD! A graph:
image

### DIFF
--- a/facilitator/src/bin/facilitator.rs
+++ b/facilitator/src/bin/facilitator.rs
@@ -3,7 +3,6 @@ use chrono::{prelude::Utc, NaiveDateTime};
 use clap::{value_t, App, Arg, ArgMatches, SubCommand};
 use log::{error, info};
 use prio::encrypt::PrivateKey;
-use prometheus::{register_counter, register_counter_vec, Counter};
 use ring::signature::{
     EcdsaKeyPair, KeyPair, UnparsedPublicKey, ECDSA_P256_SHA256_ASN1,
     ECDSA_P256_SHA256_ASN1_SIGNING,
@@ -19,6 +18,7 @@ use facilitator::{
         DataShareProcessorGlobalManifest, IngestionServerManifest, PortalServerGlobalManifest,
         SpecificManifest,
     },
+    metrics::{start_metrics_scrape_endpoint, AggregateMetricsCollector, IntakeMetricsCollector},
     sample::{generate_ingestion_sample, SampleOutput},
     task::{AggregationTask, AwsSqsTaskQueue, GcpPubSubTaskQueue, IntakeBatchTask, TaskQueue},
     transport::{
@@ -70,6 +70,8 @@ trait AppArgumentAdder {
     fn add_gcp_service_account_key_file_argument(self: Self) -> Self;
 
     fn add_task_queue_arguments(self: Self) -> Self;
+
+    fn add_metrics_scrape_port_argument(self: Self) -> Self;
 }
 
 const SHARED_HELP: &str = "Storage arguments: Any flag ending in -input or -output can take an \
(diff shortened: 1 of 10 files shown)


aaomidi (approved):

facilitator/src/bin/facilitator.rs
```diff
@@ -1032,29 +1076,20 @@ fn aggregate_subcommand(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
...
-        "aggregation_jobs_finished",
-        "Number of aggregation jobs that finished (on the facilitator side)",
-        &["status"]
-    )
-    .context("failed to register counter for finished aggregations")?;
-
     let mut queue = aggregation_task_queue_from_args(sub_matches)?;
+    let metrics_collector = AggregateMetricsCollector::new()?;
+    let scrape_port = value_t!(sub_matches.value_of("metrics-scrape-port"), u16)?;
+    let _runtime = start_metrics_scrape_endpoint(scrape_port)?;
```
jsha, line 1088:
```suggestion
    start_metrics_scrape_endpoint(scrape_port)?;
```

jsha (approved):
Looks good!

Style-wise, I might put the intake metrics struct in `intake.rs` and the aggregate metrics struct in `aggregate.rs`, but that could go either way.

Also, the addition of another lifetimed field to BatchIntaker (etc) reminds me: is there a reason that BatchIntaker takes so many references, rather than owning all its values? Either way is fine, but owned values would let you avoid some of the `&'a` noise.

facilitator/src/bin/facilitator.rs
tgeoghegan: You have to keep a live reference to the runtime or it gets dropped, deallocated, and then tasks spawned onto it never run. Admittedly I should have a comment explaining that surprising behaviour. I'd also be very happy to learn that I'm just wrong.

tgeoghegan:
Thanks for the review!

> Style-wise, I might put the intake metrics struct in `intake.rs` and the aggregate metrics struct in `aggregate.rs`, but that could go either way.

I suspect that both those structs are going to get refactored, possibly as implementations of a trait, as I add more functionality and get a better sense of what's common to them.

> Also, the addition of another lifetimed field to BatchIntaker (etc) reminds me: is there a reason that BatchIntaker takes so many references, rather than owning all its values? Either way is fine, but owned values would let you avoid some of the `&'a` noise.

The intent was to avoid unnecessary copies. In `src/bin/facilitator.rs`, we could just move values into the `BatchIntaker`, but in tests, I often want to re-use something like a `SignableTransport` or a `BatchSigningKey` in multiple objects, so it's a matter of trading some extra `&` in one place for some extra `.clone()` calls in another.

facilitator/src/bin/facilitator.rs
jsha: Oh right, that makes sense with the comment on the function. You could potentially use [`std::mem::forget()`](https://doc.rust-lang.org/std/mem/fn.forget.html) to make this explicit. But I guess the current behavior actually does run the runtime's Drop when the enclosing functions exits, which is probably nicer and tidier.

Issue #296, opened 2021-01-06
facilitator: Safer URL construction


tgeoghegan:
[@aaomidi points out](https://github.com/abetterinternet/prio-server/pull/292/files#r552233907) that we could be using a more robust URL construction technique than `format!`. We should consider moving to `ureq` >= 2.0.0 so we can use the `url` crate.

tgeoghegan:
We're now on `ureq 2.0` and using `url` in several places, but there's still some places where we construct URLs with string interpolation. That might be fine, but we might also want to move everything to safer URL construction. The remaining task in this ticket would be to at least audit places we construct URLs using `format!` or the like and convince ourselves we're not worried about them, either because they're pretty fool-proof or because they end up being validated by a call to [`url::Url::parse`](https://docs.rs/url/2.2.0/url/struct.Url.html#method.parse).

Issue #297, opened 2021-01-06
facilitator: beef up `http` crate


tgeoghegan:
We have a few places where we interact with GCP APIs using `ureq` (`task/pubsub.rs`, `transport/gcs.rs`, and `gcp_oauth.rs`). There's lots of duplicated code for doing stuff like setting sensible timeouts, `Authorization: Bearer <blah>` headers, `Content-Type` headers, and so on. We should first see about moving to `ureq` > 2.0.0 to see if it removes the need for some of this boilerplate, then see if we can refactor the remaining common configuration stuff into functions in `http`.

tgeoghegan:
Relates to #296, which could be part of fixing this.

tgeoghegan:
Resolved by #350.

PR #330, opened 2021-01-23
Explore upgrading to ureq 2

aaomidi:
Upgrade to ureq2

- Addresses #297
- Will address #296 in the future

### DIFF
--- a/facilitator/Cargo.toml
+++ b/facilitator/Cargo.toml
@@ -7,9 +7,9 @@ version = "0.1.0"
 [dependencies]
 anyhow = "1.0"
 async-trait = "0.1"
-avro-rs = {version = "0.11.0", features = ["snappy"]}
+avro-rs = { version = "0.11.0", features = ["snappy"] }
 base64 = "0.12.3"
-chrono = {version = "0.4", features = ["serde"]}
+chrono = { version = "0.4", features = ["serde"] }
 clap = "2.33.3"
 derivative = "2.1.1"
 env_logger = "0.8.1"
@@ -24,25 +24,25 @@ prio = "0.2"
 prometheus = "0.11"
 rand = "0.7"
 regex = "1.4"
-ring = {version = "0.16.15", features = ["std"]}
-rusoto_core = {version = "^0.46", default_features = false, features = ["rustls"]}
-rusoto_s3 = {version = "^0.46", default_features = false, features = ["rustls"]}
-rusoto_sqs = {version = "^0.46", default_features = false, features = ["rustls"]}
-rusoto_sts = {version = "^0.46", default_features = false, features = ["rustls"]}
-serde = {version = "^1.0", features = ["derive"]}
+ring = { version = "0.16.15", features = ["std"] }
+rusoto_core = { version = "^0.46", default_features = false, features = ["rustls"] }
+rusoto_s3 = { version = "^0.46", default_features = false, features = ["rustls"] }
+rusoto_sqs = { version = "^0.46", default_features = false, features = ["rustls"] }
+rusoto_sts = { version = "^0.46", default_features = false, features = ["rustls"] }
+serde = { version = "^1.0", features = ["derive"] }
 serde_json = "1.0"
 structopt = "0.3"
 tempfile = "3.1.0"
 thiserror = "1.0"
-tokio = {version = "^1.0", features = ["full"]}
-ureq = {version = "^2.0", features = ["json"]}
+tokio = { version = "^1.0", features = ["full"] }
+ureq = { version = "^2.0", features = ["json"] }
 url = "2.2.0"
 urlencoding = "1.1.1"
-uuid = {version = "0.8", features = ["serde", "v4"]}
+uuid = { version = "0.8", features = ["serde", "v4"] }
 warp = "^0.3"
 
 [dev-dependencies]
 assert_matches = "1.4.0"
 mockito = "0.27.0"
-rusoto_mock = {version = "^0.46", default_features = false, features = ["rustls"]}
+rusoto_mock = { version = "^0.46", default_features = false, features = ["rustls"] }
 serde_test = "1.0"
--- a/facilitator/Cargo.toml
+++ b/facilitator/Cargo.toml
@@ -27,14 +27,16 @@ rusoto_core = { version = "0.45.0", default_features = false, features = ["rustl
 rusoto_s3 = { version = "0.45.0", default_features = false, features = ["rustls"] }
 rusoto_sqs = { version = "0.45.0", default_features = false, features = ["rustls"] }
 rusoto_sts = { version = "0.45.0", default_features = false, features = ["rustls"] }
-serde = { version = "1.0", features = ["derive"] }
+serde = { version = "^1.0", features = ["derive"] }
 serde_json = "1.0"
+lazy_static = "1.4.0"
 env_logger = "0.8.1"
 structopt = "0.3"
 tempfile = "3.1.0"
 thiserror = "1.0"
 tokio = { version = "0.2", features = ["rt-core", "io-util"] }
-ureq = { version = "1.5.2", features = ["json"] }
+ureq = { version = "^2.0", features = ["json"] }
+url = "2.2.0"
 urlencoding = "1.1.1"
 uuid = { version = "0.8", features = ["serde", "v4"] }
 
--- a/facilitator/Cargo.toml
+++ b/facilitator/Cargo.toml
@@ -1,17 +1,18 @@
 [package]
-name = "facilitator"
-version = "0.1.0"
 authors = ["Internet Security Research Group"]
 edition = "2018"
+name = "facilitator"
+version = "0.1.0"
 
 [dependencies]
-async-trait = "0.1"
 anyhow = "1.0"
-avro-rs = { version = "0.11.0", features = ["snappy"] }
+async-trait = "0.1"
+avro-rs = {version = "0.11.0", features = ["snappy"]}
 base64 = "0.12.3"
-chrono = { version ="0.4", features = ["serde"] }
+chrono = {version = "0.4", features = ["serde"]}
 clap = "2.33.3"
 derivative = "2.1.1"
+env_logger = "0.8.1"
 http = "^0.2"
 hyper = "^0.14"
 hyper-rustls = "^0.22"
@@ -23,27 +24,25 @@ prio = "0.2"
 prometheus = "0.11"
 rand = "0.7"
 regex = "1.4"
-ring = { version = "0.16.15", features = ["std"] }
-serde = { version = "^1.0", features = ["derive"] }
-rusoto_core = { version = "^0.46", default_features = false, features = ["rustls"] }
-rusoto_s3 = { version = "^0.46", default_features = false, features = ["rustls"] }
-rusoto_sqs = { version = "^0.46", default_features = false, features = ["rustls"] }
-rusoto_sts = { version = "^0.46", default_features = false, features = ["rustls"] }
+ring = {version = "0.16.15", features = ["std"]}
+rusoto_core = {version = "^0.46", default_features = false, features = ["rustls"]}
+rusoto_s3 = {version = "^0.46", default_features = false, features = ["rustls"]}
+rusoto_sqs = {version = "^0.46", default_features = false, features = ["rustls"]}
+rusoto_sts = {version = "^0.46", default_features = false, features = ["rustls"]}
+serde = {version = "^1.0", features = ["derive"]}
 serde_json = "1.0"
-lazy_static = "1.4.0"
-env_logger = "0.8.1"
 structopt = "0.3"
 tempfile = "3.1.0"
 thiserror = "1.0"
-ureq = { version = "^2.0", features = ["json"] }
+tokio = {version = "^1.0", features = ["full"]}
+ureq = {version = "^2.0", features = ["json"]}
 url = "2.2.0"
-tokio = { version = "^1.0", features = ["full"] }
 urlencoding = "1.1.1"
-uuid = { version = "0.8", features = ["serde", "v4"] }
+uuid = {version = "0.8", features = ["serde", "v4"]}
 warp = "^0.3"
 
 [dev-dependencies]
 assert_matches = "1.4.0"
 mockito = "0.27.0"
-rusoto_mock = { version = "^0.46", default_features = false, features = ["rustls"] }
+rusoto_mock = {version = "^0.46", default_features = false, features = ["rustls"]}
 serde_test = "1.0"
--- a/facilitator/src/config.rs
+++ b/facilitator/src/config.rs
@@ -3,9 +3,9 @@ use chrono::Duration;
 use once_cell::sync::Lazy;
 use regex::Regex;
 use rusoto_core::{region::ParseRegionError, Region};
-use serde::{de, export::Formatter, Deserialize, Deserializer, Serialize, Serializer};
+use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 use std::{
-    fmt::{self, Display},
+    fmt::{self, Display, Formatter},
     path::PathBuf,
     str::FromStr,
 };
@@ -44,7 +44,7 @@ impl S3Path {
 }
 
 impl Display for S3Path {
-    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> fmt::Result {
         write!(
             f,
             "s3://{}/{}/{}",
@@ -108,7 +108,7 @@ impl GCSPath {
 }
 
 impl Display for GCSPath {
-    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         write!(f, "gs://{}/{}", self.bucket, self.key)
     }
 }
@@ -190,7 +190,7 @@ impl FromStr for ManifestKind {
 }
 
 impl Display for ManifestKind {
-    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         match self {
             ManifestKind::IngestorGlobal => write!(f, "ingestor-global"),
             ManifestKind::IngestorSpecific => write!(f, "ingestor-specific"),
@@ -220,7 +220,7 @@ impl FromStr for TaskQueueKind {
 }
 
 impl Display for TaskQueueKind {
-    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         match self {
             TaskQueueKind::GcpPubSub => write!(f, "gcp-pubsub"),
             TaskQueueKind::AwsSqs => write!(f, "aws-sqs"),
--- a/facilitator/src/config.rs
+++ b/facilitator/src/config.rs
@@ -44,7 +44,7 @@ impl S3Path {
 }
 
 impl Display for S3Path {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> fmt::Result {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         write!(
             f,
             "s3://{}/{}/{}",
--- a/facilitator/src/gcp_oauth.rs
+++ b/facilitator/src/gcp_oauth.rs
@@ -3,11 +3,11 @@ use chrono::{prelude::Utc, DateTime, Duration};
 use jsonwebtoken::{encode, Algorithm, EncodingKey, Header};
 use serde::{Deserialize, Serialize};
 use std::{fmt, io::Read};
-use ureq::{Agent, AgentBuilder, Response};
+use ureq::{Agent, Response};
 
 use crate::http::{
-    prepare_request, send_json_request, Method, OauthTokenProvider, RequestParameters,
-    StaticOauthTokenProvider,
+    create_agent, prepare_request, send_json_request, Method, OauthTokenProvider,
+    RequestParameters, StaticOauthTokenProvider,
 };
 use url::Url;
 
@@ -18,7 +18,7 @@ fn default_oauth_token_url() -> Url {
 
 // API reference:
 // https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken
-fn generate_access_token_for_service_account(service_account_to_impersonate: &str) -> Result<Url> {
+fn access_token_url_for_service_account(service_account_to_impersonate: &str) -> Result<Url> {
     let request_url = format!("https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{}:generateAccessToken",
             service_account_to_impersonate);
 
@@ -107,7 +107,7 @@ pub(crate) struct GcpOauthTokenProvider {
     /// though the contained token may be expired. This will always be None if
     /// account_to_impersonate is None.
     impersonated_account_token: Option<OauthToken>,
-    /// The Agent will be used when making HTTP  requests to GCP APIs to fetch
+    /// The Agent will be used when making HTTP requests to GCP APIs to fetch
     /// Oauth tokens.
     agent: Agent,
 }
@@ -164,9 +164,7 @@ impl GcpOauthTokenProvider {
             }
             None => None,
         };
-        let agent = AgentBuilder::new()
-            .timeout(std::time::Duration::from_secs(10))
-            .build();
+        let agent = create_agent();
 
         Ok(GcpOauthTokenProvider {
             scope: scope.to_owned(),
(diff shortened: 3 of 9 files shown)


tgeoghegan:
What I think should be done in this PR to resolve #296 and #297:

- bump `ureq` dependency version in Cargo.toml to `ureq ^2.0`, using a [caret requirement](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements) to accept any `ureq 2.0.x`.
- fix existing `ureq` usage sites to adopt new API versions, in particular the error handling which now makes more idiomatic use of `Result<T, E>` and timeouts which are now set on an `Agent` via `AgentBuilder`, and for which there are now default values.
- use `Agent::request_url` instead of the `get()`, `post()`, etc. functions we do now, so that we more safely construct endpoint URLs
- identify usages of `ureq` in facilitator that repeat boilerplate (i.e., setting timeouts, setting authorization headers, logging) and factor them into functions in the `http` module.

tgeoghegan:
Also: I will caution you against writing a layer in the `http` module that attempts to abstract away `ureq`. You might end up having to reimplement a big part of `ureq`'s API surface to do that, and the goal here is to reduce repetition of code, not to conceal the fact that we use `ureq` or make it easy to move to another HTTP client implementation. So don't be shy about leaking the implementation detail of `ureq` from the functions in `src/http.rs`.

facilitator/src/gcp_oauth.rs
```diff
@@ -140,6 +140,19 @@ impl OauthTokenProvider {
         })
     }
 
+    pub(crate) fn new_from_token(token: &str) -> OauthTokenProvider {
```
tgeoghegan, line 143: This doesn't seem right -- `OauthTokenProvider` is responsible for obtaining the default account token from the GKE metadata service in `ensure_default_account_token`.

aaomidi: I wrote this to handle the test cases that were just passing empty tokens. I'm not entirely sure what we want to do with those. E.g. in the case here: https://github.com/abetterinternet/prio-server/blob/7661a8f6eb5e581876fb01ee4227479e292c9154/facilitator/src/transport/gcs.rs#L164

aaomidi: Any advice what would work well here?

tgeoghegan: Oh, yeah, I see why you'd need to do that. Off the top of my head, you could either:

1. refactor `StreamingTransferWriter` so that it takes an `OauthTokenProvider`, or rather a trait object of some new trait that the existing `OauthTokenProvider` would implement. Then in the tests, you would provide an instance of a dummy `OauthTokenProvider` that just returns a token parsed from a fixed string.
2. provide variants on whatever function you add in `http.rs` to handle constructing that can take either `OauthTokenProvider` or `&str`, and `StreamingTransferWriter` would use the latter variant.

(1) would be kinda tricky, the borrow checker will yell at you a lot about `Box` and `dyn`, and refactoring how the transport stuff works is probably beyond the scope of this task.

facilitator/src/transport/gcs.rs
```diff
@@ -61,30 +67,27 @@ impl Transport for GCSTransport {
...
-            ));
-        }
+        let mut url = STORAGE_API_BASE_URL
+            .clone()
+            .join("upload/storage/v1/b/")?
+            .join(self.path.bucket.as_str())?
+            .join("/o/")?
+            .join(encoded_key.as_str())?;
+
+        let url = url.query_pairs_mut().append_pair("alt", "media").finish();
```
tgeoghegan, line 78: You lost the comment explaining the purpose of the `alt: media` header.
```
 // Ensures response body will be content and not JSON metadata.
// https://cloud.google.com/storage/docs/json_api/v1/objects/get#parameters
```

facilitator/src/task/pubsub.rs
```diff
@@ -90,29 +91,24 @@ impl<T: Task> TaskQueue<T> for GcpPubSubTaskQueue<T> {
...
+        let request = HTTP.prepare_request(RequestParameters {
+            url: &Url::parse(&url)?,
+            method: "POST",
             token_provider: Some(&mut self.oauth_token_provider),
-            // Empirically, if there are no messages available in the
-            // subscription, the PubSub API will wait about 90 seconds to send
-            // an HTTP 200 with an empty JSON body. We set higher timeouts than
-            // usual to allow for this.
-            connect_timeout_millis: Some(180_000), // three minutes
-            read_timeout_millis: Some(180_000),    // three minutes
```
tgeoghegan, line 101: I think you will need to reintroduce the timeout parameters so that this particular HTTP request can wait longer than 10 seconds. Since timeouts are set on the `ureq::Agent`, I think this also supports my arguments against the global variable `HTTP` in `http.rs`.

facilitator/src/transport/gcs.rs
```diff
@@ -61,30 +67,27 @@ impl Transport for GCSTransport {
...
+        let request = HTTP.prepare_request(RequestParameters {
+            url: &url,
+            method: "GET",
+            token_provider: Some(&mut self.oauth_token_provider),
+            token: None,
+        });
+
+        let response = request
+            .call()
+            .context(anyhow!("failed to fetch object {} from GCS", url))?;
```
tgeoghegan, line 89:
```suggestion
            .context(format!("failed to fetch object {} from GCS", url))?;
```

facilitator/src/task/pubsub.rs
```diff
@@ -230,22 +225,24 @@ impl<T: Task> GcpPubSubTaskQueue<T> {
...
-        })?;
-        if http_response.error() {
-            return Err(anyhow!(
-                "failed to modify ack deadline on task {:?}: {:?}",
-                handle,
-                http_response
-            ));
-        }
+        )
+        .context(anyhow!(
```
tgeoghegan, line 242:
```suggestion
        .context(format!(
```

facilitator/src/http.rs
```diff
@@ -1,65 +1,109 @@
...
-    pub connect_timeout_millis: Option<u64>,
-    /// Timeout in milliseconds for reads. If unset, a default timeout of 10
-    /// seconds is applied.
-    pub read_timeout_millis: Option<u64>,
-    /// The JSON body to be sent in the request, constructed with ureq::json!
-    pub body: SerdeValue,
+    /// If the token_provider field isn't set, and this field is set, the request will
+    /// be sent with an "Authorization" header containing a bearer token obtained from
+    /// the OauthTokenProvider.
+    pub token: Option<&'a str>,
```
tgeoghegan, line 44: I think this is acceptable as is. However, having separate fields for `token_provider` and `token` is unfortunate, because now it's possible to provide both a `token_provider` and a `token`, but that's clearly invalid. I know your implementation will simply prefer the `token_provider`, but the principle of "make invalid states unrepresentable" suggests that we could find a better way. I have two different suggestions for improving this (in both cases you remove the `token` field on `RequestParameters`:

- you could define a new trait `OauthTokenProvider` that defines the method `ensure_oauth_token`, then refactor the existing `OauthTokenProvider` into an implementation of that trait and rename it `GcpOauthTokenProvider`, and then provide a struct like `StaticOauthTokenProvider` which would just wrap a `String` and return it from its implementation of `ensure_oauth_token`. Then the `token_provider` member of `RequestParameters` would be `Option<dyn OauthTokenProvider>` (or `Option<Box<dyn OauthTokenProvider>>`, whichever appeases the compiler) and you would always call `ensure_oauth_token` from `get_authenticated_request`. Callers would be responsible for instantiating a `StaticOauthTokenProvider` if they happen to have a token in hand already (or even better, you might be able to provide a [`std::convert::From<String>`](https://doc.rust-lang.org/std/convert/trait.From.html) implementation on `StaticOauthTokenProvider` so that they can pass their `String` and it will  automatically become an `OauthTokenProvider`!
- Make the `token_provider` field of `RequestParameters` into an [`FnMut`](https://doc.rust-lang.org/std/ops/trait.FnMut.html) that returns the token as a `String`. In the case where the caller has an `OauthTokenProvider`, they would provide a closure that calls `ensure_oauth_token()`. If the caller has an Oauth token as a string, they provide a closure that just returns that string.

Like I said, the current approach is fine, but I think either of the approaches above are a teeny bit better, and interesting Rust exercises.

tgeoghegan (changes requested):
I left a lot of comments, but I think this is definitely going in the right direction. I like the `prepare_request` pattern to deduplicate boilerplate but still allow the caller to insert whatever headers they need, or to then call either `send_bytes()` or `send_json()`.

facilitator/Cargo.toml
```diff
@@ -24,18 +24,20 @@ prometheus = "0.11"
...
 regex = "1.4"
 ring = { version = "0.16.15", features = ["std"] }
+serde = { version = "^1.0", features = ["derive"] }
 rusoto_core = { version = "^0.46", default_features = false, features = ["rustls"] }
 rusoto_s3 = { version = "^0.46", default_features = false, features = ["rustls"] }
 rusoto_sqs = { version = "^0.46", default_features = false, features = ["rustls"] }
 rusoto_sts = { version = "^0.46", default_features = false, features = ["rustls"] }
-serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+lazy_static = "1.4.0"
```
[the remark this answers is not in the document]
aaomidi: I did some investigations on once_cell, it seemed like it was going to add a lot of boilerplate:


For example (from their docs):
```
use std::{sync::Mutex, collections::HashMap};
use once_cell::sync::Lazy;

static GLOBAL_DATA: Lazy<Mutex<HashMap<i32, String>>> = Lazy::new(|| {
    let mut m = HashMap::new();
    m.insert(13, "Spica".to_string());
    m.insert(74, "Hoyten".to_string());
    Mutex::new(m)
});

fn main() {
    println!("{:?}", GLOBAL_DATA.lock().unwrap());
}
```

This just seemed to be a lot of code compared to 

```
lazy_static! {
    static ref STORAGE_API_BASE_URL: Url =
        Url::parse("https://storage.googleapis.com/").expect("unable to parse storage url");
}
```

And in the docs they do say that what they're doing is essentially what the lazy_static! macro does.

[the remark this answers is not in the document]
aaomidi: I'll experiment more with it as well

facilitator/src/config.rs
```diff
@@ -44,7 +44,7 @@ impl S3Path {
 }
 
 impl Display for S3Path {
-    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> fmt::Result {
```
aaomidi, line 47: Is this change fine? I had to do it since `export:Formatter` is no longer public in the `serde` crate.

facilitator/src/gcp_oauth.rs
```diff
@@ -262,29 +270,27 @@ impl OauthTokenProvider {
...
-            .build();
-        let http_response = send_json_request(JsonRequestParameters {
-            request: request,
-            body: ureq::json!({
+
+        let request = HTTP.prepare_request(RequestParameters {
+            url: &Url::parse(&request_url)?,
+            method: "POST",
+            token: None,
+            token_provider: Some(self),
```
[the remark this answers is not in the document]
aaomidi: Oh yeah whoops

facilitator/src/http.rs
```diff
@@ -1,65 +1,109 @@
...
-use ureq::{Request, Response, SerdeValue};
+use anyhow::{Context, Result};
+use lazy_static::lazy_static;
+use std::time::Duration;
+use ureq::{Agent, AgentBuilder, Error, Request, Response, SerdeValue};
+use url::{ParseError, Url};
 
 use crate::gcp_oauth::OauthTokenProvider;
+lazy_static! {
+    pub(crate) static ref HTTP: Http = Http::new();
```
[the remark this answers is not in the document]
aaomidi: The `Agent` has a connection pool we can utilize:

https://docs.rs/ureq/2.0.1/ureq/struct.Agent.html

https://docs.rs/ureq/2.0.1/ureq/struct.AgentBuilder.html

I can keep a private internal global state that I just clone for prepare_request, or make a new Agent on each prepare_request. It's a little unclear what the ureq API is expecting me to do.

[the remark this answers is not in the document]
aaomidi: Decision is to create a new agent on each prepare_request

facilitator/src/config.rs
tgeoghegan: This is interesting! OK, so [`std::fmt::Display::fmt`](https://doc.rust-lang.org/std/fmt/trait.Display.html#tymethod.fmt) wants a [`std::fmt::Formatter`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html). That's a concrete struct, not a trait, so how did this ever work? Intuitively the compiler should have rejected the `serde::export::Formatter`. I think what happened is that Serde used to have a module `export`, which [they removed recently](https://github.com/serde-rs/serde/commit/dd1f4b483ee204d58465064f6e5bf5a457543b54#diff-110ad2e16d30874ce5f0193e158b33b11289693a8ac60024f774317b92f4f857). That module contained the line `pub use lib::fmt::{self, Formatter};`, which re-exported `std::fmt::Formatter` as `serde::export::Formatter`. Finally it's `lib::fmt::Formatter` because [Serde's `lib.rs` re-exports all of `std::*` as `lib::*` within the Serde crate](https://github.com/serde-rs/serde/blob/b0c99ed761d638f2ca2f0437522e6c35ad254d93/serde/src/lib.rs#L152).

tl;dr: it was always wrong to use `serde::export::Formatter` and we were just lucky that it happened to alias to `std::fmt::Formatter`, so your change is correct. I would just add `Formatter` to the existing `use std::fmt{}` block at the top of this file so you don't need the fully qualified type name in the signatures of all the `fmt()` methods.

aaomidi: Awesome! Thank you for this!

facilitator/Cargo.toml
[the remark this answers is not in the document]
aaomidi: Well, I don't need lazy_static at all anymore

facilitator/src/http.rs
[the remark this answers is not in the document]
aaomidi: Moved to using a shared agent for different components of the code. e.g. pubsub, gcs transport, gcp_oauth get their own agent

facilitator/src/http.rs
```diff
@@ -1,65 +1,109 @@
...
-        Err(anyhow!("fetching {}: status {}", url, resp.status()))
-    } else {
-        resp.into_string()
-            .context(format!("reading body of {}", url))
+pub(crate) fn get_request(request: Request) -> Result<String> {
+    let response = request.call();
+    Ok(match response {
+        Err(error) => format!("{}", error),
+        Ok(response) => response.into_string().context("reading body failed")?,
+    })
```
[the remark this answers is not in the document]
aaomidi: So concise! Love it

facilitator/src/http.rs
aaomidi: Thank you for this detailed explanation! The first option felt more rusty :rabbit:, so I went with that.

facilitator/src/transport/s3.rs
```diff
@@ -136,23 +141,37 @@ impl S3Transport {
...
-                            // read.
-                            .timeout_connect(10_000) // ten seconds
-                            .timeout_read(10_000) // ten seconds
-                            .call();
-                        if response.error() {
-                            return Err(CredentialsError::new(format!(
+                            .finish();
+
+                        let mut request = prepare_request(
+                            None,
```
tgeoghegan, line 155: nit: because there's no named parameters in Rust, it's not obvious what is being omitted here. Instead of having a single `prepare_request(Option<Agent>, RequestParameters)`, you could provide `prepare_request(RequestParameters)` and `prepare_request_with_agent(Agent, RequestParameters)`.

facilitator/src/transport/gcs.rs
```diff
@@ -369,12 +422,18 @@ mod tests {
             .expect_at_most(1)
             .create();
 
+        println!(
+            "TEST {} {}",
+            mockito::server_url(),
+            Url::parse(&mockito::server_url()).expect("hello")
+        );
```
tgeoghegan, line 429: Looks like debug code that should be deleted

facilitator/src/transport/gcs.rs
```diff
@@ -192,7 +226,13 @@ impl StreamingTransferWriter {
             minimum_upload_chunk_size,
             buffer: Vec::with_capacity(minimum_upload_chunk_size * 2),
             object_upload_position: 0,
-            upload_session_uri: upload_session_uri.to_owned(),
+            upload_session_uri: Url::parse(&upload_session_uri).context(format!(
+                "failed to parse upload_session_uri url: {}",
+                &upload_session_uri
+            ))?,
+            agent: AgentBuilder::new()
+                .timeout(std::time::Duration::from_secs(10))
```
tgeoghegan, line 234:
```suggestion
                .timeout(Duration::from_secs(10))
```
...and `use std::time::Duration` at the top of the file

facilitator/src/gcp_oauth.rs
```diff
@@ -193,14 +205,22 @@ impl OauthTokenProvider {
...
     /// ureq::Response, whose body will be an OauthTokenResponse if the HTTP
-    /// call was successful, but may be an error.
-    fn account_token_from_gke_metadata_service() -> Response {
-        ureq::get(DEFAULT_OAUTH_TOKEN_URL)
-            .set("Metadata-Flavor", "Google")
-            // By default, ureq will wait forever to connect or read.
-            .timeout_connect(10_000) // ten seconds
-            .timeout_read(10_000) // ten seconds
+    /// call was successful.
+    fn account_token_from_gke_metadata_service(agent: Agent) -> Result<Response> {
```
tgeoghegan, line 209: I'm honestly not sure why this wasn't already a method on `OauthTokenProvider`, but now that it takes the `Agent`, it should just be `fn account_token_from_gke_metadata_service(&self) -> Result<Response>` and use `self.agent`, ideally without cloning.

facilitator/src/gcp_oauth.rs
```diff
@@ -262,29 +292,32 @@ impl OauthTokenProvider {
...
+
+        let request = prepare_request(
+            Some(self.agent.clone()),
+            RequestParameters {
+                url: Url::parse(&request_url).context(format!(
+                    "failed to parse request_url to generate access token: {}",
+                    &request_url
+                ))?,
+                method: Method::Post,
+                token_provider: Some(self),
```
tgeoghegan, line 304: I don't think this will work. `prepare_request` calls `RequestParameters::authenticated_request` which calls `token_provider.ensure_oauth_token` which will call back into this and you will have an infinite recursion. I think you just need to call `self.ensure_default_account_token()` and wrap that in a `StaticOauthTokenProvider`.

facilitator/src/http.rs
```diff
@@ -1,65 +1,127 @@
...
+    }
+}
+
+impl std::convert::From<String> for StaticOauthTokenProvider {
+    fn from(token: String) -> Self {
+        StaticOauthTokenProvider { token }
+    }
+}
 
 /// Struct containing parameters for send_json_request
```
tgeoghegan, line 61: This comment is no longer accurate since I don't pass `RequestParameters` to `send_json_request`.

tgeoghegan (changes requested):
Some more comments but this is on the right track! I think we're getting close.

facilitator/src/gcp_oauth.rs
```diff
@@ -262,29 +292,32 @@ impl OauthTokenProvider {
         // https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken
         let request_url = format!("https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{}:generateAccessToken",
```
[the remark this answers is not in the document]
aaomidi: After a bit of discussion with @jsha I think my decision was to just trust Url::parse for now, and then see if we need to change our URL generation per user input. Right now I think the risk is low and join() has a few gotchas.

facilitator/src/transport/gcs.rs
```diff
@@ -61,30 +72,35 @@ impl Transport for GCSTransport {
...
         let encoded_key = urlencoding::encode(&[&self.path.key, key].concat());
-        let url = format!(
-            "{}/storage/v1/b/{}/o/{}",
-            STORAGE_API_BASE_URL, self.path.bucket, encoded_key
+
+        let formatted_url = &format!(
+            "{}storage/v1/b/{}/o/{}",
+            storage_api_base_url().to_string(),
+            self.path.bucket,
+            encoded_key
```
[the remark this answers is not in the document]
aaomidi: Url::join() seems to have a few gotchas that I don't fully understand yet and don't want to potentially introduce a bug. If it's ok maybe let's leave this code as is for now and then come back to it when I've learned a bit more about the semantics of joins?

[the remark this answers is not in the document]
aaomidi: I also want to kinda just move all of our URL building to joins, so I feel like I can tackle them all in one go.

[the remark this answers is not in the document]
aaomidi: I will however  move the URL parsing here into functions

facilitator/src/transport/s3.rs
aaomidi: I didn't even notice this because vscode was highlighting the labels for me: 

image

I'll add that other function, but probably as `prepare_request_without_agent`, instead of `with`, considering majority of our usage of this function will come with the agent.

facilitator/src/gcp_oauth.rs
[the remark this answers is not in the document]
tgeoghegan: Sounds reasonable to me.

facilitator/src/transport/gcs.rs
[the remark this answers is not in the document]
tgeoghegan: OK, seems reasonable.

jsha:
For some reason I can't comment on some of these conversations but regarding whether to use `Url::join` to build URLs: I think what you probably want instead is [`path_segments_mut`](https://docs.rs/url/2.2.0/url/struct.Url.html#method.path_segments_mut)

tgeoghegan:
Is this ready for re-review?

facilitator/src/http.rs
```diff
@@ -1,65 +1,127 @@
...
-#[derive(Debug, Default)]
-pub(crate) struct JsonRequestParameters<'a> {
-    /// The ureq::Request to be sent
-    pub request: Request,
+#[derive(Debug)]
+pub(crate) struct RequestParameters<'a> {
+    /// The url to request
+    pub url: Url,
```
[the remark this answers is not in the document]
aaomidi: Since this change would make the default URL handling a little messy, I've avoided using a reference.

aaomidi:
Yes @tgeoghegan, thanks for reviewing!

tgeoghegan:
> Yes @tgeoghegan, thanks for reviewing!

Ack, I will review this afternoon!

facilitator/src/gcp_oauth.rs
```diff
@@ -117,42 +135,49 @@ impl fmt::Debug for OauthTokenProvider {
...
         let key_file: Option<ServiceAccountKeyFile> = match key_file_reader {
             Some(reader) => {
                 serde_json::from_reader(reader).context("failed to deserialize JSON key file")?
             }
             None => None,
         };
-        Ok(OauthTokenProvider {
+        let agent = AgentBuilder::new()
+            .timeout(std::time::Duration::from_secs(10))
+            .build();
```
tgeoghegan, line 169: nit: this could use `crate::http::create_agent()`

tgeoghegan (approved):
A couple more minor things to address but I think this is pretty much good to go. Can you share any testing details? If you were able to do intake and aggregations in a dev env, then that should exercise all the code paths involved here.

aaomidi:
> edit: also, please consider rebasing into fewer commits.

I was going to squash and merge, should I avoid doing that?

facilitator/Cargo.toml
```diff
@@ -1,17 +1,18 @@
 [package]
```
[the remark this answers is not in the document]
aaomidi: Not even sure why it removed the spaces, that was less valid TOML haha

tgeoghegan:
I think a single commit or a small number of commits, each of which explains clearly what the change is and why it was made, is more pleasant and more useful than 25 commits reading "work in progress" "oops" "address feedback" (I'm not picking on you, these are the kinds of intermediate commits I make all the time). But it's subjective and since I don't propose to write a commit message linter, it's up to you to decide what you think is good enough.

aaomidi:
Will try to consolidate the commits into manageable pieces.

tgeoghegan:
Heuristically, I think about it like this: what is going to be most useful to future authors and debuggers of this code when they  they want to know why we're setting so-and-so header in an HTTP request? They're going to `git blame` the line they are interested in and look up the commit, and we should do what we can today to give that future reader the quickest possible path to the context they need to understand the code.

PR #336, opened 2021-01-26
facilitator: use a new SqsClient for each request

tgeoghegan:
NCI observed a problem where acknowledging tasks in SQS was failing
because the connection pool underneath Rusoto was closing idle
connections, leading to a race condition when trying to reuse an
SqsClient. The simple fix is to create a new SqsClient for each request
to the SQS API.


tgeoghegan:
We confirmed that this change resolved the issue NCI was seeing by cherry-picking this change into 0.6.2 and deploying it to their production.

aaomidi (approved):

PR #339, opened 2021-01-29
Facilitator changes for integration-tester

aaomidi:
> So this is absolutely wrong because I'm using pkcs8, the pkcs10 package isn't ready yet. But it technically "works", although I think the public key it creates is wrong. I'll have to see what happens if I encrypt with this type of key. 
> 
> It might not be too hard to manually implement pkcs10 stuff ourselves... who knows though.

Well, what's up there is old news. The code can handle creating base64ed public keys now!


facilitator/src/bin/facilitator.rs
```diff
@@ -705,15 +731,47 @@ fn main() -> Result<(), anyhow::Error> {
     result
 }
 
+fn get_ecies_public_key(
```
aaomidi, line 734: I was thinking of potentially turning the input of this into a short lived struct and make this a method on it. Not sure if that would be cleaner than this.

tgeoghegan: This seems essentially consistent with other functions like `intake_task_queue_from_args` or `public_key_map_from_arg`.

facilitator/src/bin/facilitator.rs
```diff
@@ -466,7 +466,20 @@ fn main() -> Result<(), anyhow::Error> {
                             "Base64 encoded X9.62 uncompressed public key for the PHA \
                             server",
                         )
-                        .required(true),
+                )
+                .arg(
+                    Arg::with_name("pha-manifest-url")
```
tgeoghegan, line 471: From reading the remainder of the changes, it seems like this is the full URL to a specific manifest, e.g., "https://somebase.url/ta-ta-ingestor-1-manifest.json`. Could you instead take parameters for ingestor name and locality and then use the existing `SpecificManifest::from_https`?

facilitator/taplo.toml
```diff
@@ -0,0 +1,4 @@
+include = ["Cargo.toml"]
```
tgeoghegan, line 1: What does this do?

tgeoghegan (commented):
I scanned this quickly and it seems like you're going in the right direction here. Good find on the `pkix` crate! I think you'll need to add something like a `generate-ingestion-sample-worker` subcommand to `facilitator`, akin to `intake-batch-worker`, except it might wait on a timer instead of pulling tasks from a queue.

facilitator/taplo.toml
aaomidi: Was thinking we can potentially use https://taplo.tamasfe.dev/ to keep our `Cargo.toml` files in proper formatting. Was just experimenting with it and won't include it in this PR at the end. Will make a separate PR for it later.

aaomidi:
> I think you'll need to add something like a generate-ingestion-sample-worker subcommand to facilitator

Isn't the generate-ingestion-sample subcommand something I can use for this? I'm a little confused on why I would need to wait on a timer.

PR #340, opened 2021-01-29
terraform: basic Prometheus alerting rules

tgeoghegan:
Configures the Prometheus Helm chart to include alerting rules defined
in a standalone YAML file. Updates `prometheus-alertmanager`
configuration so that only alerts with `severity: page` are routed to
the Victorops receiver.

We now have the following alerts set up:

 - page if any `workflow-manager` instance has not run in the last 30
   minutes
 - page if more than 95% of intake tasks fail during the last hour
 - page if more than 95% of aggregate tasks fail during the last 8 hours
 - page if any PubSub queue has undelivered messages for an entire hour

### DIFF
--- a/docs/monitoring.md
+++ b/docs/monitoring.md
@@ -1,0 +1,90 @@
+# Metrics, graphs and alerting
+
+We deploy Prometheus and Grafana into Prio clusters. We gather metrics from a few sources:
+
+ - host level metrics via [`prometheus-node-exporter`](https://prometheus.io/docs/guides/node-exporter/)
+ - Kubernetes metrics via [`prometheus-kube-state-metrics`](https://github.com/kubernetes/kube-state-metrics)
+ - [GCP Cloud Monitoring](https://cloud.google.com/monitoring) via [`stackdriver-exporter`](https://github.com/prometheus-community/stackdriver_exporter)
+ - `workflow-manager` pushes metrics via [`prometheus-pushgateway`](https://prometheus.io/docs/instrumenting/pushing/)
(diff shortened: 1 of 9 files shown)


aaomidi (approved):

aaomidi:
Just to double check, do we have code that only deploys this on prod/staging?

tgeoghegan:
Good question! The alerts get deployed everywhere, but `prometheus-alertmanager` is not hooked up to Victorops in all environments. So outside of prod, alerts will be tracked in `alertmanager`, but only in prod will they generate pages. We will have to get more sophisticated here, though, because while I want to get alerts from staging, I don't want them to buzz someone's phone at night. I think the thing to do will be to introduce an environment label on alerts so we can define different alert routing rules which resolve to different receivers with different urgencies. But right now, I want to get any alerting in prod at all.

tgeoghegan:
I went and created a `prio-staging` routing key in VictorOps (we already had a `prio-prod-us` one). In this latest commit, I configure `prod-us.tfvars` and `staging-{facil,pha}.tfvars` to send alerts to their corresponding VO routing keys. `prio-prod-us` is hooked up to an escalation policy that pages; `prio-staging` is hooked up to one that emails and sends a message to a slack channel.

aaomidi (approved):

tgeoghegan:
While testing in dev, I found a problem with the `prometheus-alertmanager` config: I amended the alert routing so that only `severity: page` alerts go to VictorOps, but that means there's nowhere to send other alerts. I'm going to set up a default alert route that emails or slacks us. While I'm re-opening this PR, I'm also going to add an alert for dead letter queue subscription depth, so we can catch undelivered messages.

PR #342, opened 2021-02-01
terraform: deploy us-va in prod-us

tgeoghegan:
Adds an instance to support `us-va`. Our understanding is that we expect
a load similar to MD or CT, and so we allocate the same numbers of
workers.


aaomidi (approved):
:shipit:

Issue #41, opened 2020-10-06
Implement retries in S3 requests


tgeoghegan:
As a workaround for a mismatch between connection pool timeouts in Hyper and AWS S3, #36 carefully constructs HTTP clients with a carefully chosen timeout value. For more robustness, we should implement retries.

yuriks:
https://docs.aws.amazon.com/AmazonS3/latest/dev/ErrorBestPractices.html

tgeoghegan:
#260 is progress towards this, but I'm leaving this open because that PR only handles HTTP dispatch errors. We should have more comprehensive error handling as described in Amazon docs, and should also do exponential backoff.

tgeoghegan:
More good material on AWS retry best practices: https://github.com/rusoto/rusoto/issues/234#issuecomment-395534542

PR #348, opened 2021-02-04
facilitator: don't abort integration for bad batch

### LINKED ISSUE #343: facilitator: Don't let a single invalid batch abort the entire aggregation
Currently, the entire aggregation task is aborted and declared a failure if any individual ingestion, own validation or peer validation batch is found to be invalid (e.g., invalid signature or malformed messages). So for instance, if we are summing over 900 batches and one is mangled for some reason, we will never aggregate the other 899 which might be fine. We have a means to relay to the portal server exactly what set of batches was included in an aggregation, so we should just ignore bad batches and sum all the others.

### LINKED ISSUE #345: facilitator: metrics on invalid peer batches
Intake or aggregation jobs can fail for a bunch of reasons, but one common case is malformed data or invalid signatures -- i.e., bad data sent by a peer (whether ingestion server or peer data share processor). We should track metrics on these specific failure modes so that we can detect if a peer starts sending a lot of invalid data.

tgeoghegan:
We are losing whole aggregations because one peer validation batch out
of ~900 happens to have an invalid signature or an incorrect packet file
digest in its header. We should still be able to aggregate the other 899
batches in such a case. We now continue with the aggregation if an
individual call to `BatchAggregator::aggregate_share()` fails.

We also now gather metrics on how many invalid own and peer validation
batches are encountered during an aggregation. This will enable us to
get a better idea of how often the bug that caused invalid validation
batches occurs.

Resolves #343
Resolves #345

### DIFF
--- a/facilitator/src/aggregation.rs
+++ b/facilitator/src/aggregation.rs
@@ -91,6 +91,7 @@ impl<'a> BatchAggregator<'a> {
             self.aggregation_batch.path(),
         );
         let mut invalid_uuids = Vec::new();
+        let mut included_batch_uuids = Vec::new();
 
         let ingestion_header = self.ingestion_header(&batch_ids[0].0, &batch_ids[0].1)?;
 
@@ -107,7 +108,18 @@ impl<'a> BatchAggregator<'a> {
             .collect::<Vec<Server>>();
 
         for batch_id in batch_ids {
-            self.aggregate_share(&batch_id.0, &batch_id.1, &mut servers, &mut invalid_uuids)?;
+            if let Err(error) =
+                self.aggregate_share(&batch_id.0, &batch_id.1, &mut servers, &mut invalid_uuids)
+            {
+                log::info!(
+                    "ignoring batch {}/{} due to error {:?}",
+                    batch_id.0,
+                    batch_id.1,
+                    error
+                );
+            } else {
+                included_batch_uuids.push(batch_id.0);
+            }
             callback();
         }
 
@@ -145,7 +157,7 @@ impl<'a> BatchAggregator<'a> {
 
         let sum_signature = self.aggregation_batch.put_header(
             &SumPart {
-                batch_uuids: batch_ids.iter().map(|pair| pair.0).collect(),
+                batch_uuids: included_batch_uuids,
                 name: ingestion_header.name,
                 bins: ingestion_header.bins,
                 epsilon: ingestion_header.epsilon,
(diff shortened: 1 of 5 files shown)


tgeoghegan:
One problem worth mentioning in this PR is retries. Let's say we are aggregating some batches and while trying to download one of them, we encounter a 500 internal server error from GCS. Previously, we would fail the overall aggregation, and then retry at some point because PubSub would redeliver the task. Now, we will simply skip over that batch, even though it's possible we'd succeed if we retried. I opted not to deal with this for now because it requires us to go over the errors reported in facilitator with a fine toothed comb and accurately classify them as retryable, or not. I opted to punt on that in the interest of getting this fix out sooner, to improve the flow of data in production.

In any case, rather than surface retryability all the way up to say `aggregate.rs`, I think the thing to do would be to improve the various API clients we have and make them do retries as appropriate based on each service's documented best practices. Then a caller doing (for instance) `GCSTransport.get()` could assume that appropriate retries had already been done on their behalf.

aaomidi (approved):
Looks good to me!

facilitator/src/aggregation.rs
```diff
@@ -207,10 +219,37 @@ impl<'a> BatchAggregator<'a> {
...
             );
-        let peer_validation_header = peer_validation_batch
-            .header(&self.peer_validation_transport.batch_signing_public_keys)?;
-        let own_validation_header = own_validation_batch
-            .header(&self.own_validation_transport.batch_signing_public_keys)?;
+
+        let peer_validation_header = match peer_validation_batch
+            .header(&self.peer_validation_transport.batch_signing_public_keys)
+        {
+            Ok(header) => header,
```
aaomidi, line 226: nit: a bit of repeatedish code here between peer and own validation header, but I don't think it's really an issue.

tgeoghegan: I agree. The tricky thing is that here, I want to propagate the error as `?` would, but I want to intercept it so I can increment the metric, and doing that is a bit awkward as we see here. I think a better solution would be to either:
- plumb the metric collector down into the place where the header was rejected (`batch.rs` in this case) and increment the counter there
- plumb a better error (i.e. a variant on `crate::Error` like `InvalidHeaderError`) all the way up to some context that has the metrics collector in scope and can increment the metric appropriately after matching on the error.

Both of these seem like more complicated exercises than I want to tackle in this PR.

jsha (approved):
Looks good to me. The one thing I would add is a threshold for how many bad batches an aggregation has before it gives up and errors out. I'm imagining a case where >90% of batches are bad, but the aggregations are plugging away and we're not noticing.

tgeoghegan:
That's why I put in the metrics on invalid peer and own validation batches. Now we can monitor that and potentially alert on it. Additionally, we report to the portal server how many batches are included in an aggregation as well as how many individual contributions were summed. MITRE has already wired up an alert for mismatched sum part files (i.e., to catch cases where NCI sends a sum for a period but ISRG didn't), and they're going to extend it to look for significant discrepancies between how much is included in either sum, so we should get reports if e.g., NCI submits a sum part with 100 batches it it, but ISRG's corresponding sum only has 10.

tgeoghegan:
Filed some issues to track retry logic in each client: #351, #352, #353, #354 and we already had #41 for S3.

PR #350, opened 2021-02-04
ureq version bump to 2.0

aaomidi:
The change caused an issue like this:

```
gs://prio-timg-dev-facil-narnia-ingestor-1-ingestion/kittens-seen%2F2021%2F02%2F04%2F17%2F16%2Fba53ce56-f4a1-4e75-bee5-88ffa33d656b.batch
```

- [x] Tested in test env

After testing this in a dev environment the problem has gone away.

The issue was that the object name was getting urlencoded twice, which when Google decoded it it still left one round of encoding in it.

### DIFF
--- a/facilitator/Cargo.lock
+++ b/facilitator/Cargo.lock
@@ -75,9 +75,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.13"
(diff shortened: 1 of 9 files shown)


tgeoghegan (approved):
LGTM, good find!

Issue #266, opened 2020-12-04
manifest linter should reject unknown keys in JSON


tgeoghegan:
This ingestor manifest:

```
{
  "format": 1,
  "server-identity": {
    "gcp-service-account-email": "enpa-ingestion-batch-writer@enpa-ingestion-narnia.iam.gserviceaccount.com",
    "google-service-account": "112310747466759665351"
  },
  "batch-signing-public-keys": {
    "projects/enpa-ingestion-narnia/locations/us/keyRings/enpa-signature-key-ring/cryptoKeys/enpa-signature-key/cryptoKeyVersions/1": {
      "public-key": "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEtt7BOz0So/JY4Ml0CT2zxQAtXUDY\nJoiG9A82XI2lA56fIAE8ECkQX5NuRCrk6lx3Tg6biH0GBZMWSfwENa6DQg==\n-----END PUBLIC KEY-----\n",
      "expiration": "20210101T000000Z"
    }
  }
}
```

... is invalid because it has `google-service-account` instead of `gcp-service-account-id`. However, because JSON parsing ignores unexpected keys, and `gpp-service-account-id` is optional, it doesn't raise an error.

bmw:
We should continue to not error on unknown keys outside of linting right? If we're fine erroring all the time, this is trivial with [`#[serde(deny_unknown_fields)]`](https://serde.rs/container-attrs.html#deny_unknown_fields), but I'd have to do some more digging to figure out how to do this conditionally.

tgeoghegan:
I think we should always reject unexpected keys. The format for manifests is versioned so I don't think we need to be particularly permissive in parsing (i.e., we don't aim to be backward or forward compatible). Plus, the issue description contains an example of a manifest where the author _meant_ to use `gcp-service-account-id` instead of `google-service-account`, and the error would have been more obvious if we'd rejected the unexpected key.

PR #357, opened 2021-02-05
Reject unexpected manifest keys

bmw:
Fixes https://github.com/abetterinternet/prio-server/issues/266.

Now when running `facilitator lint-manifest` on the manifest from #266, the output is:
```
Error: failed to decode JSON manifest

Caused by:
    unknown field `google-service-account`, expected one of `aws-iam-entity`, `gcp-service-account-id`, `gcp-service-account-email` at line 5 column 28
```

The changes made to the other test cases in this PR were to get them passing for the documented reason. For instance, the comment above a manifest we expect to be invalid may say something like `// No format key`, however, the manifest was also invalid due to an incorrect field name, missing comma, etc.

### DIFF
--- a/facilitator/src/manifest.rs
+++ b/facilitator/src/manifest.rs
@@ -479,7 +479,7 @@ mod tests {
     }
 }
         "#,
-            // unexpected field
+            // unexpected top-level field
             r#"
 {
     "format": 0,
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):
LGTM, thank you!

PR #368, opened 2021-02-08
Removes Facilitator's incomplete WFM implementation

### LINKED ISSUE #366: facilitator: delete dead `workflow-manager` implementation
The `workflow-manager` we ship is written in Go and is under `workflow-manager`. There's another, partial Rust implementation of a `workflow-manager` under `facilitator`. We should delete that and any code elsewhere in the crate that only it uses.

aaomidi:
Fixes #366


tgeoghegan (changes requested):
I think there are also some things in `config.rs` that are only used by `workflow.rs`, like `DayDuration`, that can be deleted. Similarly we shouldn't need the dependency on the `structopt` crate anymore. Can you please check for any other dead code or dependencies we can drop?

tgeoghegan (approved):

PR #369, opened 2021-02-09
Add dependabot configuration for all of our currently used packages

aaomidi:
We're relying on a bunch of dependencies, let's add dependabot (https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/) configuration to our source so we can use the automated dependency updates offered by dependabot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,27 @@
+version: 2
+
+updates:
+  - package-ecosystem: "gomod"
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):
LGTM, assuming the dependabot config file is correct, but could you amend the commit with some more information on what dependabot is, why we want it, and how we have configured it?

PR #370, opened 2021-02-09
workflow-manager: control memory footprint

### LINKED ISSUE #162: workflow-manager: limit listing of files to recent ones
The S3 ListObjectsV2 API (https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html) has a limit of 1000 objects; presumably GS has a similar limit. When workflow-manager runs, it should construct a set of path prefixes that would match the given --max-age flag, in order to ignore stuff that is too old and get more relevant results in that 1000.

tgeoghegan:
Previously, `workflow-manager` would list the entirety of the ingestion,
own validation and peer validation buckets, and then process the
contents looking for batches to intake, or validations and ingestion
batches to aggregate, or task markers. This meant that the memory
footprint of `workflow-manager` was a function of how many objects were
in the bucket, which occasionally caused OOMs.

We ameliorate this situation by using GCS and S3 API features allowing
us to refine our queries on bucket contents. On the GCS side, we use
`startOffset` and `endOffset` parameters to only list batch files and
task markers whose timestamps fall within the interval of interest. On
the S3 side, we construct a series of requests with a `Prefix` parameter
(see comments in `storage.go` for more details). Now, instead of listing
seven days' (or whatever the bucket retention policies are set to) worth
of content, we should list a number of objects corresponding to the
`intake-max-age`, `aggregation-period` and `grace-period` parameters (in
production, these are 6h, 8h and 4h, respectively).

In either case, constructing the requests requires knowing the
aggregation IDs up front, because the aggregation ID comes before any
timestamp in all our object names. Therefore, we now perform a query
to discover the top-level directories in the ingestion bucket to
determine what aggregation IDs exist, and then enumerate batch files and
task markers for each aggregation in sequence. So beyond restricting the
scope of queries by time intervals, we now also will query objects
pertaining to each aggregation ID separately, further reducing how many
objects must be listed in memory at once.

In support of all this, I refactored some existing structs and functions
from various places in `workflow-manager` into the `time` module. I also
renamed `bucket` to `storage` (to avoid the stutter of bucket.Bucket)
and added some tests to both.

Finally, I amended the `kubernetes.tf` module so that when setting up a
test env, we now make an extra `sample-maker` cronjob in each namespace.
On top of the existing one which generated fake data for the
`kittens-seen` aggregation ID, we now also generate sample data with
aggregation ID `dogs-petted`, so that our dev and staging environments
will consistently exercise multiple aggregations traveling through the
system.

Fixes #162


aaomidi (approved):

workflow-manager/batchpath/batchpath.go
```diff
@@ -40,6 +55,22 @@ func (bpl List) Swap(i, j int) {
 	bpl[i], bpl[j] = bpl[j], bpl[i]
 }
 
+// WithinInterval returns the subset of the batches in the receiver that are
+// within the given Interval.
+func (bpl List) WithinInterval(interval wftime.Interval) []string {
```
aaomidi, line 60: I would probably move this to time.go module you made. It seems like it would make sense to have something like this on the Interval struct

tgeoghegan: So something like `func (i *Interval) BatchesWithin(list batchpath.List) []string`? Sure, sounds reasonable!

aaomidi: I think more so `func (i *Interval) TimeWithin(time time.Time)` but that works too

tgeoghegan: That's even better! I will incorporate that suggestion.

PR #380, opened 2021-02-11
Automatically build main-latest images when pushes are made to main

aaomidi:


tgeoghegan:
I like the idea of having non-tagged container images out there. This is going to overwrite the `:dev` versions of all four images on each merge to main, right? Could we instead version them with the short git SHA (i.e. `letsencrypt/prio-facilitator:feedface`) so that we then have a history of images built from `main`? Also, what are our storage limits for Docker Hub?

.github/workflows/push-docker-image-release.yml
```diff
@@ -56,53 +56,3 @@ jobs:
       run: docker push letsencrypt/prio-workflow-manager:${{ steps.get_version.outputs.VERSION }}
     - name: push-latest
       run: docker push letsencrypt/prio-workflow-manager:latest
-
-  push-deploy-operator:
```
aaomidi, line 60: We're not using these yet. When we do I'll bring them back.

tgeoghegan (approved):

Issue #298, opened 2021-01-06
Use log levels in GCP


tgeoghegan:
Right now, everything we log from either `workflow-manager` or `facilitator` appears in the GCP log explorer as error level messages. We'd be able to search logs much more effectively if log messages were appropriately sorted into error, info, debug, etc.

tgeoghegan:
Done in #413 and #424

PR #413, opened 2021-02-16
Improve go logging

aaomidi:
Partially addresses https://github.com/abetterinternet/prio-server/issues/298

Move to `zerolog`, no particular reason for this library, but it seemed to be pretty clean and doesn't lock us into using stackdriver.

I've confirmed log levels work in my dev environment.


workflow-manager/main.go
```diff
@@ -76,29 +77,42 @@ var (
...
 	workflowManagerLastFailure.SetToCurrentTime()
-	log.Fatalf(format, args...)
+	log.Fatal().Msgf(format, args...)
+}
+
+func prepareLogger() {
+	zerolog.LevelFieldName = "severity"
+	zerolog.TimestampFieldName = "timestamp"
+	zerolog.TimeFieldFormat = time.RFC3339Nano
+	zerolog.SetGlobalLevel(zerolog.TraceLevel)
```
tgeoghegan, line 87: Trace is too verbose to be the default, but it'd be nice to be able to turn up logging when we need to for debugging. Can you set the default to `Info`, and provide a command line flag for log level? That said if we never log anything above `Info` then we can just turn down the default and skip the command line flag for now.

workflow-manager/main.go
```diff
@@ -270,14 +284,17 @@ func main() {
 			aggregationPeriod:       aggregationPeriodParsed,
 			gracePeriod:             gracePeriodParsed,
 		})
+		if err != nil {
+			log.Err(err).Str("aggregation ID", aggregationID).Msgf("Failed to schedule aggregation tasks: %s", err)
+		}
```
tgeoghegan, line 289: omfg was I not checking this error?! Wow, I'm really glad you caught this! Besides logging the error, this branch should return, to avoid incorrectly setting `workflowManagerLastSuccess`.

workflow-manager/main.go
```diff
@@ -76,29 +77,42 @@ var (
...
 
 func main() {
+	prepareLogger()
 	startTime := time.Now()
-	log.Printf("starting %s version %s. Args: %s", os.Args[0], BuildInfo, os.Args[1:])
+	log.Info().
+		Str("app", os.Args[0]).
+		Str("version", BuildInfo).
+		Str("Args", strings.Join(os.Args[1:], ",")).
+		Msgf("starting %s version %s. Args: %s", os.Args[0], BuildInfo, os.Args[1:])
```
tgeoghegan, line 97: There are several places where values get logged in fields and then also in the message format string. Is that intentional?

tgeoghegan (changes requested):
Looks great! I left a couple comments, but the only really important one is that we should `return` if a call to `scheduleTasks` fails.

workflow-manager/main.go
aaomidi: I initially removed them but then brought it back so with a quick glance to the logs it's visible. I did not do this to more complicated structures that we might not necessarily care about seeing at a quick glance to the logs:

e.g. 

kept it for this case: https://github.com/abetterinternet/prio-server/pull/413/files#diff-58d7389d6ae70e2e9725b328431cd15b6b70b2e9908ce8f5d9b55c84b3f1e3d7R364

removed it for this case:  https://github.com/abetterinternet/prio-server/pull/413/files#diff-58d7389d6ae70e2e9725b328431cd15b6b70b2e9908ce8f5d9b55c84b3f1e3d7R519

It's a very subjective decision.

aaomidi: We can be consistent and only do structured if that makes more sense. Your call!

tgeoghegan: I think that over time, as we operate the service, we'll learn more about what's annoying and what's helpful and can always refine these log statements.

tgeoghegan (approved):

PR #416, opened 2021-02-17
facilitator: structured error handling

tgeoghegan:
This change is the start of more rigorous error handling in
`facilitator`, making better use of Rust enums to classify errors. We
introduce several new variants to the `Error` enum to specifically
capture errors around parsing or validating Avro batches. This enables
`BatchAggregator` to report when a failure is due to invalid data, in
which case we do not want to retry. We also can now tell when a failure
was due to bad data from a peer (report this but don't alert) or from
ourselves (do alert).
    
Over time, we should eliminate most if not all uses of `anyhow!` and the
`.context()` method from `Anyhow::Context` in favour of introducing
specific variants on `crate::Error`, and have all module interfaces
return `Result<T, crate::Error>` rather than the implicit `Result<T,
anyhow::Error>`.
    
See https://nick.groenen.me/posts/rust-error-handling/ for stimulating
discussion of library vs. application error handling in Rust and that
intersects with usage of `thiserror` vs. `anyhow`.

### DIFF
--- a/facilitator/src/lib.rs
+++ b/facilitator/src/lib.rs
@@ -21,23 +21,56 @@ pub const DATE_FORMAT: &str = "%Y/%m/%d/%H/%M";
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
-    /// Compatibility enum entry to ease migration.
-    #[error(transparent)]
-    // TODO(yuriks): The `From` impl should stay disabled while not updating errors, to avoid
-    //   accidentally wrapping an `EofError` inside `AnyhowError`, which would then get missed when
-    //   code tries to pattern-match against it. I don't trust our current test coverage enough to
-    //   detect all cases where that happens and leads to incorrect behavior.
-    //   Later on, `anyhow` downcasting with `Error::is` can be used to check for `EofError`.
-    AnyhowError(/*#[from]*/ anyhow::Error),
+    /// An error occurred while reading or writing an ingestion batch
+    #[error("error reading/writing ingestion batch")]
+    BadIngestionBatch(#[source] Box<Error>),
+    /// An error occurred while reading or writing a peer validation batch
+    #[error("error reading/writing peer validation batch")]
+    BadPeerValidationBatch(#[source] Box<Error>),
+    /// An error occurred while reading or writing an own validation batch
+    #[error("error reading/writing own validation batch")]
+    BadOwnValidationBatch(#[source] Box<Error>),
 
+    // Errors from batch module
+    #[error("{header_path}: key {key_name} not present in key map")]
+    NoSuchKey {
+        header_path: String,
+        key_name: String,
+    },
+    #[error("invalid signature on header {header_path} with key {key_name}")]
+    InvalidSignature {
+        header_path: String,
+        key_name: String,
+    },
+    #[error("packet file digest in header {header_path} {expected_digest} does not match actual packet file digest {actual_digest}")]
+    DigestMismatch {
+        header_path: String,
+        expected_digest: String,
+        actual_digest: String,
+    },
+
+    // Errors from aggregation module
+    #[error("values in batch headers do not match: {lhs:?}\n{rhs:?}")]
+    HeaderMismatch { lhs: String, rhs: String },
+
+    // Errors from idl module
     #[error("avro error: {0}")]
-    AvroError(String, #[source] avro_rs::Error),
+    Avro(String, #[source] avro_rs::Error),
     #[error("malformed header: {0}")]
-    MalformedHeaderError(String),
+    MalformedHeader(String),
     #[error("malformed data packet: {0}")]
-    MalformedDataPacketError(String),
+    MalformedDataPacket(String),
     #[error("end of file")]
-    EofError,
+    Eof,
+
+    #[error("invalid value for argument {0}")]
+    InvalidArgument(String),
+
+    /// Catchall case for errors not captured by specific variants of this enum.
+    /// As we encounter more errors that should be distinguished from others
+    /// programmatically, specific variants should be added to this enum.
+    #[error(transparent)]
+    Anyhow(#[from] anyhow::Error),
 }
 
 /// An implementation of transport::TransportWriter that computes a SHA256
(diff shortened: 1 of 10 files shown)


facilitator/src/lib.rs
```diff
@@ -21,23 +21,56 @@ pub const DATE_FORMAT: &str = "%Y/%m/%d/%H/%M";
...
+    // Errors from idl module
     #[error("avro error: {0}")]
-    AvroError(String, #[source] avro_rs::Error),
+    Avro(String, #[source] avro_rs::Error),
     #[error("malformed header: {0}")]
-    MalformedHeaderError(String),
+    MalformedHeader(String),
     #[error("malformed data packet: {0}")]
-    MalformedDataPacketError(String),
+    MalformedDataPacket(String),
```
jsha, line 62: It seems like these errors compete with Bad{Peer,Own}{Validation,Ingestion}Batch. Are the `Bad` Errors supposed to subsume these, or be a catch-all for when these don't apply? I would (at a glance) vote in favor of subsuming.

facilitator/src/lib.rs
```diff
@@ -21,23 +21,56 @@ pub const DATE_FORMAT: &str = "%Y/%m/%d/%H/%M";
...
+    #[error("packet file digest in header {header_path} {expected_digest} does not match actual packet file digest {actual_digest}")]
+    DigestMismatch {
+        header_path: String,
+        expected_digest: String,
+        actual_digest: String,
+    },
+
+    // Errors from aggregation module
+    #[error("values in batch headers do not match: {lhs:?}\n{rhs:?}")]
+    HeaderMismatch { lhs: String, rhs: String },
```
jsha, line 54: Same question: How does this interact with the `Bad` errors?

facilitator/tests/integration_tests.rs
```diff
@@ -311,9 +311,11 @@ fn aggregation_including_invalid_batch() {
...
     .generate_sum_part(&batch_uuids_and_dates, || {})
     .unwrap_err();
-    // Ideally we would be able to match on a variant in an error enum to check
-    // what the failure was but for now check the error description
-    assert!(err.to_string().contains("packet file digest in header"));
+
+    match err {
+        Error::BadPeerValidationBatch(_) => (),
+        _ => assert!(false, "unexpected error {:?}", err),
+    };
```
jsha, line 318: This works, but I'm also a fan of `assert!(matches!(err, Error::BadPeerValidationBatch(_)))`

facilitator/src/lib.rs
tgeoghegan: The idea is that `Bad{Own,Peer}{Validation,Ingestion}Batch` can wrap any other error defined in `crate::Error`. A module like `idl` knows only that it is reading or writing an Avro message. It doesn't have the context to know if that message was part of an own or peer or ingestion batch, so all it can say in an error case is that a message was unparseable and return an error like `Error::Avro`. Then a higher level module like `aggregation.rs` has the context to know what the batch with a bogus header was, so it can augment the `Error::Avro` with `Error::BadOwnValidationBatch`. 

Then if the caller of `aggregation` gets `Error::BadOwnValidationBatch(Box<Error::Avro(avro_rs::Error)>)`, it knows from the top level error that the problem was a bad validation batch emitted by self, and if it cares, it can traverse the chain of sources to get more specifics about what went wrong (or it can log the whole chain of source errors).

Perhaps the tl;dr is that the different modules in `facilitator` treat each other like libraries and wrap each other's errors.

aaomidi (approved):

facilitator/src/lib.rs
jsha: > Perhaps the tl;dr is that the different modules in facilitator treat each other like libraries and wrap each other's errors.

Given that, I think it makes sense for the `idl` module to define its own error type (or pass through `avro_rs::Error`), and have `facilitator`'s error type only include the wrapping types. I'm thinking something like:


facilitator:
```
enum Action {
  Validation,
  Ingestion,
}

enum Error {
  Batch(Entity, Action, BadBatch),
  InvalidArgument(String),
  Anyhow(#[from] anyhow::Error),
}

enum BadBatch {
  Avro(..),
  MalformedHeader(..),
  MalformedDataPacket(..),
  ..
}
```

tgeoghegan: I gave this a try and to be honest I really don't like it: if there are distinct concrete error types for each module, then you can't return the concrete `A::Error` type from module `A`'s methods unless `A::Error` defines variants that wrap the errors from each module that `A` uses, which leads to each error enum having an `Anyhow(#[from] anyhow::Error)` variant, and several error enums having `Avro(#[from] idl::Error)` variants.

The other thing I can think to do would be to define a trait implemented by all error types in the crate, and then everything can return `Result<T, Box<dyn FacilitatorError>>`, but I think that's pretty ugly, makes it difficult to match on errors, and in terms of the method signature giving callers a hint about which errors they can expect to get from any method, I think it's equivalent to having one big `Error` type covering the whole crate.

jsha: > unless A::Error defines variants that wrap the errors from each module that A uses
> which leads to each error enum having an Anyhow(#[from] anyhow::Error)

One way to address this would be to make the Error type a struct with a Kind enum, like `io::Error`. That way the struct can take care of setting an optional source, which can be anything, while the Kind enum can remain nice and clean for matching.

All that said: If you are happy with this approach, I'm happy. I just keep pushing on it because I know you like to explore alternate ways to do things, and this doesn't feel quite right to me yet.

tgeoghegan: >All that said: If you are happy with this approach, I'm happy. I just keep pushing on it because I know you like to explore alternate ways to do things, and this doesn't feel quite right to me yet.

I agree, and I enjoy trying different approaches as well. I think I'll ponder a bit more and experiment a bit more to see if something that feels right emerges.

facilitator/tests/integration_tests.rs
tgeoghegan: I've since learned about [`assert_matches`](https://docs.rs/assert_matches/1.5.0/assert_matches/), which is really slick because besides asserting on a pattern match, it also lets you bind parts of the pattern for further matches. e.g.
```
assert_matches!(err, Error::BadPeerValidationBatch(e) => {
    assert_eq!(e, "whatever expected error message");
});
```

tgeoghegan:
The primary motivation for this PR is gone, and the changes here are long since out of date w.r.t. `prio-server` main. If we ever did pursue this, we would open a new pull request. I captured what's interesting about this set of changes in #1001.

PR #424, opened 2021-02-18
Better logging for faciltiator

aaomidi:
This is step one in the progress for better logging. Let's get the severity values right and then we can look into adding more fields to search by to the payload.

### DIFF
--- a/facilitator/src/logging.rs
+++ b/facilitator/src/logging.rs
@@ -3,23 +3,25 @@ use log::{Level, Record};
 use serde::Serialize;
 use std::io::Write;
 
+/// Severity maps `log::Level` to Google Cloud Platform's notion of Severity.
+/// https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogSeverity
 #[derive(Debug, Serialize)]
-#[serde()]
+#[serde(rename_all = "UPPERCASE")]
 enum Severity {
-    DEBUG,
-    INFO,
-    WARNING,
-    ERROR,
+    Debug,
+    Info,
+    Warning,
+    Error,
 }
 
 impl Severity {
-    fn from_log_level(level: log::Level) -> Severity {
+    fn from_log_level(level: Level) -> Severity {
         match level {
-            Level::Error => Severity::ERROR,
-            Level::Warn => Severity::WARNING,
-            Level::Info => Severity::INFO,
-            Level::Debug => Severity::DEBUG,
-            Level::Trace => Severity::DEBUG,
+            Level::Error => Severity::Error,
+            Level::Warn => Severity::Warning,
+            Level::Info => Severity::Info,
+            Level::Debug => Severity::Debug,
+            Level::Trace => Severity::Debug,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


facilitator/src/logging.rs
```diff
@@ -0,0 +1,57 @@
+use env_logger::fmt::Formatter;
+use log::{Level, Record};
+use serde::Serialize;
+use std::io::Write;
+
+#[derive(Debug, Serialize)]
+#[serde()]
```
tgeoghegan, line 7: I think this should be `#[serde(rename_all = "UPPERCASE")]` and then the variatns should be named according to Rust convention. The enum should also have a doccomment explaining why it exists -- it's to map `log::Level` to GCP's notion of log severity, right?

facilitator/src/logging.rs
```diff
@@ -0,0 +1,57 @@
...
+enum Severity {
+    DEBUG,
+    INFO,
+    WARNING,
+    ERROR,
+}
+
+impl Severity {
+    fn from_log_level(level: log::Level) -> Severity {
```
tgeoghegan, line 16:
```suggestion
    fn from_log_level(level: Level) -> Severity {
```

tgeoghegan (changes requested):
Looks good to me, bar a couple minor issues. Can you confirm that when we log multi-line statements (e.g., a task pulled from a PubSub subscription), it yields a single log statement in the Google Log Explorer?

aaomidi:
```
dequeued task: ack ID: BCEhPjA-RVNEUAYWLF1GSFE3GQhoUQ5PXiM_NSAoRRQCCBQFfH1xXlp1VVUaB1ENGXJ8ZnI4CBoGChMHeVVbEQh6bVxXJ1AFH3Z_Y3doWRYDAEZVeltz_LyNr6Lm2GMXIkrt3YXbbDPRh7pdZho9XxJLLD5-MzVFQV5AEkw4AERJUytDCypYEU4 task: aggregation ID: kittens-seen batch ID: 80d939fe-8f86-4f75-832b-701ae333a6cf date: 2021/02/19/18/36
```

Single line

facilitator/src/logging.rs
```diff
@@ -0,0 +1,59 @@
+use env_logger::fmt::Formatter;
+use log::{Level, Record};
+use serde::Serialize;
+use std::io::Write;
+
+/// Severity maps `log::Level` to Google Cloud Platform's notion of Severity.
+/// https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogSeverity
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "UPPERCASE")]
```
tgeoghegan, line 9: The point of using `rename_all = "UPPERCASE"` is that you can then have the variants be named in the more conventional Rust fashion, which would be `Debug, Info, Warning, Error`, but still have them be serialized in uppercase.

https://serde.rs/container-attrs.html

aaomidi: Oh that's an embarrassing misunderstanding! Thanks will fix this.

tgeoghegan (approved):

PR #441, opened 2021-02-23
Allow us to define what kind of machines and node count for GKE

aaomidi:
We're spinning up a LOT of nodes and big machines for dev environments and staging, we don't need to do this. This PR surfaces the variables used when constructing the GKE environment.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -167,12 +167,13 @@ variable "victorops_routing_key" {
   description = "VictorOps/Splunk OnCall routing key for prometheus-alertmanager"
   default     = "bogus-routing-key"
 }
+
 variable "cluster_settings" {
   type = object({
-    initial_node_count : number
-    min_node_count : number
-    max_node_count : number
-    machine_type : string
+    initial_node_count = number
+    min_node_count     = number
+    max_node_count     = number
+    machine_type       = string
   })
 }
 
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -24,11 +24,6 @@ variable "aws_profile" {
   default = "leuswest2"
 }
 
-variable "machine_type" {
-  type    = string
-  default = "e2.small"
-}
-
 variable "localities" {
   type = list(string)
 }
@@ -172,6 +167,14 @@ variable "victorops_routing_key" {
   description = "VictorOps/Splunk OnCall routing key for prometheus-alertmanager"
   default     = "bogus-routing-key"
 }
+variable "cluster_settings" {
+  type = object({
+    initial_node_count : number
+    min_node_count : number
+    max_node_count : number
+    machine_type : string
+  })
+}
 
 terraform {
   backend "gcs" {}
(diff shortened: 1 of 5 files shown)


terraform/main.tf
```diff
@@ -172,6 +167,14 @@ variable "victorops_routing_key" {
   description = "VictorOps/Splunk OnCall routing key for prometheus-alertmanager"
   default     = "bogus-routing-key"
 }
+variable "cluster_settings" {
+  type = object({
+    initial_node_count : number
```
tgeoghegan, line 172: I didn't know this was also valid syntax for TF object definitions! Elsewhere we use `name = type`, and if possible I'd prefer to use that here for consistency.

terraform/variables/prod-us.tfvars
```diff
@@ -69,6 +68,12 @@ ingestors = {
     }
   }
 }
+cluster_settings = {
+  initial_node_count : 4
+  min_node_count : 4
+  max_node_count : 5
```
tgeoghegan, line 74: As elsewhere, please use `label = value` style for consistency.

tgeoghegan, line 74: The only reason that `initial_node_count` and `min_node_count` were 4 was to hack around a GKE-imposed resource limit on the number of Kubernetes jobs per namespace, which was necessary back when we spawned one k8s job per task. We don't need to do that anymore, so we can use lower values to allow the cluster to scale down if nodes are idle. Let's try `initial_node_count = 1` and `min_node_count = 3`.

terraform/variables/staging-facil.tfvars
```diff
@@ -45,6 +44,12 @@ ingestors = {
     }
   }
 }
+cluster_settings = {
+  initial_node_count : 2
+  min_node_count : 2
+  max_node_count : 3
+  machine_type : "e2-standard-4"
```
tgeoghegan, line 51: Let's try `e2-standard-2`.

terraform/variables/staging-facil.tfvars
```diff
@@ -45,6 +44,12 @@ ingestors = {
     }
   }
 }
+cluster_settings = {
+  initial_node_count : 2
+  min_node_count : 2
```
tgeoghegan, line 49: Our node pools are regional/multi-zonal, so all these values are per-AZ. I think one node per AZ will be enough for staging. Can we try `initial_node_count = 1` and `min_node_count = 1`?

tgeoghegan (changes requested):

terraform/main.tf
aaomidi: I actually got this syntax for where we define ingestors: https://github.com/abetterinternet/prio-server/blob/main/terraform/main.tf#L36-L45

aaomidi: Oh now I see what you meant, using `=` instead of `:`

I'm really curious if there is a difference between these two.

PR #443, opened 2021-02-24
Implements trace logging

aaomidi:
This PR implements trace logging. We want to be able to track how data flows through the various components of our system.

Lifecycle:

- Workflow manager discovers there is intake/aggregation to be done. It creates Intake jobs in PubSub and assigns a UUID `trace ID` to the PubSub message.
- Facilitator parses that `trace ID` and when logging the intake/aggregation process it will also print out the `trace ID`

Limitations:
- I didn't find a way to change how the trace ID flows between the intake -> aggregation tasks. Which means that there won't be any matching trace ID between the intake and aggregation tasks as they're generated in the workflow manager.
- I believe to have trace IDs that cross that boundary, we would need to change the spec and contain them in the batches.


facilitator/src/bin/facilitator.rs
```diff
@@ -795,6 +795,7 @@ fn generate_sample(sub_matches: &ArgMatches) -> Result<(), anyhow::Error> {
 }
 
 fn intake_batch<F: FnMut()>(
+    trace_id: Option<&String>,
```
tgeoghegan, line 798: `&String` seems odd. Why not `&str`, or just `String`?

facilitator/src/task.rs
```diff
@@ -66,6 +66,8 @@ pub trait Task: Debug + Display + Sized + serde::de::DeserializeOwned {}
 #[derive(Debug, Deserialize, PartialEq)]
 #[serde(rename_all = "kebab-case")]
 pub struct IntakeBatchTask {
+    /// The trace identifier for the intake, typically a UUID
+    pub trace_id: Option<String>,
```
tgeoghegan, line 70: I was trying to figure out why this has to be `Option<String>`, and the explanation I came up with is that when we deploy a `workflow-manager`/`facilitator` that support trace IDs, there will still be queued up tasks that don't have trace IDs, so `facilitator` needs to handle both. If that's why this is `Option`, could you please (1) explain that in a comment and (2) file an issue to remove support for tasks without trace IDs once we are sure we don't need to support them anymore?

tgeoghegan, line 70: If the trace IDs are UUIDs, let's enforce that by making this `Option<Uuid>` so that we can error out if we get malformed trace IDs.

workflow-manager/go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	cloud.google.com/go/pubsub v1.10.0
 	cloud.google.com/go/storage v1.13.0
 	github.com/aws/aws-sdk-go v1.37.20
+  github.com/google/uuid v1.2.0
```
tgeoghegan, line 9:
```suggestion
	github.com/google/uuid v1.2.0
```

workflow-manager/task/task.go
```diff
@@ -31,6 +32,8 @@ type Task interface {
 
 // Aggregation represents an aggregation task
 type Aggregation struct {
+	// TraceID is the tracing identifier for the aggregation. Typically a UUID
+	TraceID string `json:"trace-id"`
```
tgeoghegan, line 36: Similarly to the Rust side, let's formally make this a `uuid.UUID` so the type system can help us avoid mistakes.

workflow-manager/task/task.go
```diff
@@ -42,6 +45,13 @@ type Aggregation struct {
 	// this task
 	Batches []Batch `json:"batches"`
 }
+func (a Aggregation) LogEvent() *zerolog.Event {
```
tgeoghegan, line 48: Unused function should be deleted.

tgeoghegan (changes requested):
The goal is for every log line that pertains to a given intake or aggregation task to be annotated with the trace ID. What I want is to be able to do something like this:

1. Get an alert that something is wrong
1. Search for error messages and extract the trace ID from one
1. Search for logs that have that trace ID
1. Get all logs from the relevant intake or aggregation task.

It seems like this commit plumbs the trace ID into `intake_batch` in `facilitator.rs`, but then that function doesn't do anything with the trace ID. What's missing here:

1. Plumb the trace ID into `aggregate()` in `facilitator.rs`
1. Have both intake and aggregate pass the trace ID down such that *every* log statement related to a task is annotated with the trace ID.

tgeoghegan:
>     * I didn't find a way to change how the trace ID flows between the intake -> aggregation tasks. Which means that there won't be any matching trace ID between the intake and aggregation tasks as they're generated in the workflow manager.
> 
>     * I believe to have trace IDs that cross that boundary, we would need to change the spec and contain them in the batches.

That's fine -- an aggregation task is distinct from an intake task. It would be nice to be able to get the trace IDs for all the intake tasks that contributed to an aggregation task, but I don't know how to do that off the top of my head either.

facilitator/src/task.rs
aaomidi: Yep! That was the reason I made them optional. I will add the explanation and the issue!

aaomidi:
Comment to remove the Optional modifier https://github.com/abetterinternet/prio-server/issues/452

facilitator/src/aggregation.rs
```diff
@@ -246,14 +250,16 @@ impl<'a> BatchAggregator<'a> {
...
+                "trace id {} validation headers do not match. Peer: {:?}\nOwn: {:?}",
+                self.trace_id,
                 peer_validation_header,
                 own_validation_header
             ));
         }
         if !ingestion_header.check_parameters(&peer_validation_header) {
             return Err(anyhow!(
-                "ingestion header does not match peer validation header. Ingestion: {:?}\nPeer:{:?}",
+                "trace id {} ingestion header does not match peer validation header. Ingestion: {:?}\nPeer:{:?}",
```
tgeoghegan, line 261: This is cool, but it'd be even better if we could log the trace ID as a field in a structured JSON message the way we do on the Go side. If nothing else, it'd make it easier to construct log queries that span `workflow-manager` and `facilitator`. [slog](https://crates.io/crates/slog) seems popular. Maybe we should adopt that?

aaomidi: That's in my "TODO" for a future PR that just overhauls logging in facilitator.

tgeoghegan: [tracing](https://crates.io/crates/tracing) also seems popular and is by the Tokio authors, which is promising.

tgeoghegan: >"TODO" for a future PR

Sounds good!

tgeoghegan (approved):

PR #453, opened 2021-03-01
terraform: support multiple peer DSPs

### LINKED ISSUE #60: Terraform: refactor peer data share processor variables to account for multiple operators
#51 makes the assumption that all peer data share processors are operated by a single organization (i.e., ISRG operates all facilitators, NIH/NCI operates all PHA servers). In particular, the Terraform modules assume a single global manifest is associated with all peer data share processors. This won't hold forever, so we should refactor the representation of peer data share processors to allow for multiple operators and multiple global manifests. This would mean restructuring the `peer_share_processor_names` in the top-level `.tfvars` to either be a list of `pha-name, global manifest` pairs, or perhaps a map structure like

```
{
    operator-name-1 => {
        global-manifest-url => <url>
        pha-names => [pha-name1, pha-name2, ...]
    }
    operator-name-2 => {
        global-manifest-url => <url>
        pha-names => [...]
    }
}
```

...and then create and configure data share processors appropriately.

tgeoghegan:
We can now configure manifest base URLs for the peer data share
processor and portal server on a per-locality basis. Specifically, this
enables us to configure the `ta-ta` locality in production to exchange
data with the staging/test environments operated by our colleagues at
the NCI and MITRE.

Resolves #60

This isn't deployable right now because test-en-analytics.cancer.gov isn't available. NCI is currently rebuilding their test env. We also now repeat some amount of information in each locality. It'd be nice to be able to specify default peer and portal server manifest base URLs to avoid repetition, but I opted to do what would yield the simplest Terraform, since our modules are already kind of complicated.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -32,8 +32,10 @@ variable "ingestors" {
   type = map(object({
     manifest_base_url = string
     localities = map(object({
-      intake_worker_count    = number
-      aggregate_worker_count = number
+      intake_worker_count                    = number
+      aggregate_worker_count                 = number
+      peer_share_processor_manifest_base_url = string
+      portal_server_manifest_base_url        = string
     }))
   }))
   description = "Map of ingestor names to per-ingestor configuration."
@@ -46,14 +48,6 @@ variable "manifest_domain" {
 
 variable "managed_dns_zone" {
   type = map(string)
-}
-
-variable "peer_share_processor_manifest_base_url" {
-  type = string
-}
-
-variable "portal_server_manifest_base_url" {
-  type = string
 }
 
 variable "test_peer_environment" {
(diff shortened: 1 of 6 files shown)


tgeoghegan:
[NCI's test global manifest](https://test-en-analytics.cancer.gov/global-manifest.json) has become available since I opened this PR, so it should be mergeable an deployable.

aaomidi (approved):

aaomidi:
I haven't done the staging deploy yet, so I can include this change as well.

tgeoghegan:
> I haven't done the staging deploy yet, so I can include this change as well.

In any case, this change should only actually affect the eventual production deploy. If I did it right, there should be no change to staging. Another note: you will have to update the `.tfvars` files for your test envs so that each locality defines `peer_share_processor_manifest_base_url` and `portal_server_manifest_base_url`.

PR #458, opened 2021-03-02
Use slog in facilitator

aaomidi:
Using slog I'm able to capture more details from the log statements and push them into the JSON payload for LogEntry.

Created a slog drain: https://docs.rs/slog/2.7.0/slog/trait.Drain.html
Applied it as a global logger per slog-scope: https://docs.rs/slog-scope/4.4.0/slog_scope/

### DIFF
--- a/facilitator/src/aggregation.rs
+++ b/facilitator/src/aggregation.rs
@@ -11,8 +11,8 @@ use crate::{
 use anyhow::{anyhow, Context, Result};
 use avro_rs::Reader;
 use chrono::NaiveDateTime;
-use log::info;
 use prio::server::{Server, VerificationMessage};
+use slog_scope::{error, info};
 use std::{
     collections::{HashMap, HashSet},
     convert::TryFrom,
@@ -87,12 +87,11 @@ impl<'a> BatchAggregator<'a> {
         mut callback: F,
     ) -> Result<()> {
         info!(
-            "trace id {} processing intake from {}, own validity from {}, peer validity from {} and saving sum parts to {}",
-            self.trace_id,
+            "processing intake from {}, own validity from {}, peer validity from {} and saving sum parts to {}",
             self.ingestion_transport.transport.transport.path(),
             self.own_validation_transport.transport.path(),
             self.peer_validation_transport.transport.path(),
-            self.aggregation_batch.path(),
+            self.aggregation_batch.path(); "trace id" => self.trace_id
         );
         let mut invalid_uuids = Vec::new();
         let mut included_batch_uuids = Vec::new();
@@ -198,6 +197,8 @@ impl<'a> BatchAggregator<'a> {
         servers: &mut Vec<Server>,
         invalid_uuids: &mut Vec<Uuid>,
     ) -> Result<()> {
+        let trace_id = self.trace_id;
+
         let mut ingestion_batch: BatchReader<'_, IngestionHeader, IngestionDataSharePacket> =
             BatchReader::new(
                 Batch::new_ingestion(self.aggregation_name, batch_id, batch_date),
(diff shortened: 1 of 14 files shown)


facilitator/src/test_utils.rs
```diff
@@ -119,8 +118,8 @@ pub fn default_pha_signing_public_key() -> UnparsedPublicKey<Vec<u8>> {
 // the top of any test we want logs from.
 // https://docs.rs/env_logger/0.8.2/env_logger/#capturing-logs-in-tests
 pub fn log_init() {
-    let _ = env_logger::builder()
-        .filter_level(LevelFilter::Info)
-        .is_test(true)
-        .try_init();
+    // let _ = env_logger::builder()
```
aaomidi, line 121: Need to figure out if this is still applicable with this.

tgeoghegan: [This](https://github.com/slog-rs/slog/wiki/FAQ#my-logs-are-not-captured-in-unit-tests) might help?

facilitator/src/logging.rs
```diff
@@ -32,28 +42,84 @@ struct LogEntry {
...
 }
 
+impl convert::From<&Record<'_>> for LogEntry {
+    fn from(record: &Record) -> Self {
+        let severity: Severity = record.level().into();
+        let message: String = record.msg().to_string();
+        let time = chrono::Utc::now().to_rfc3339();
+
+        let mut entry = LogEntry::new(message, severity, time);
+        let _ = record.kv().serialize(record, &mut entry);
```
tgeoghegan, line 68: What's the returned value being discarded here?

facilitator/src/logging.rs
```diff
@@ -32,28 +42,84 @@ struct LogEntry {
...
+        self.extra.insert(k.to_string(), value);
+
+        Ok(())
+    }
+}
+
+/// JSONDrain is a custom slog drain implementation that implements the
+/// slog::Drain trait. It acts as a final output for the log messages.
+#[derive(Debug)]
+struct JSONDrain {}
```
tgeoghegan, line 91: Would [slog-json](https://docs.rs/slog-json/2.3.0/slog_json/) save us some code? Seems like a `slog-json::Json` hooked up to [`std::io::stdout`](https://doc.rust-lang.org/std/io/fn.stdout.html) would do the trick.

facilitator/src/aggregation.rs
```diff
@@ -387,7 +398,7 @@ impl<'a> BatchAggregator<'a> {
                     // server.aggregate
                     .context(format!(
                         "trace id {} failed to validate packets",
-                        self.trace_id
+                        &self.trace_id
```
tgeoghegan, line 401: I think maybe errors should not include the trace ID in the message, because then they won't show up when we search for log statements by the JSON field "trace id". I think the thing to do would be to have any log line that logs an error include a trace ID. So for instance, this error returned by `BatchAggregator` would eventually get logged by [`facilitator.rs::aggregate_worker()`](https://github.com/abetterinternet/prio-server/blob/d1b991ec8733105a8a1d89176dd7b84f74215c75/facilitator/src/bin/facilitator.rs#L1166):

```
            match result {
                Ok(_) => queue.acknowledge_task(task_handle)?,
                Err(err) => {
                    error!("error while processing task {}: {:?}", task_handle, err);
                    queue.nacknowledge_task(task_handle)?;
                }
            }
```
`trace_id` is in scope there, so the above would become:
```
            match result {
                Ok(_) => queue.acknowledge_task(task_handle)?,
                Err(err) => {
                    error!("error while processing task {}: {:?}", task_handle, err; "trace id" => trace_id);
                    queue.nacknowledge_task(task_handle)?;
                }
            }
```

tgeoghegan (changes requested):
I think this looks pretty good, given my limited understanding of `slog`. I like how it's a mostly seamless, drop in replacement for `log` since the macros are the same.

tgeoghegan:
@aaomidi I think this PR is obsolete in light of #566, though the research you did here was invaluable, so thank you! Do you mind if we close this?

PR #462, opened 2021-03-05
Sync main with bugfixes on release/0.6.2-nci

tgeoghegan:
NCI has been running a `facilitator` forked from tag 0.6.2 for some time now, and I have been making targeted hotfixes on `release/0.6.2-nci` to minimize risk to NCI's deployment. This PR brings `main` to a state where it can safely be deployed to NCI and eventually MITRE's AWS deployments. See the individual commit messages for details on what changed and the motivation.

### DIFF
--- a/facilitator/src/aws_credentials.rs
+++ b/facilitator/src/aws_credentials.rs
@@ -1,160 +1,208 @@
-use anyhow::Result;
-use rusoto_core::{
-    credential::EnvironmentProvider,
-    credential::{
-        AutoRefreshingProvider, AwsCredentials, ContainerProvider, CredentialsError,
-        InstanceMetadataProvider, ProfileProvider, ProvideAwsCredentials,
-    },
+use crate::{
+    config::Identity,
+    http::{prepare_request_without_agent, Method, RequestParameters},
 };
+use anyhow::{Context, Result};
+use async_trait::async_trait;
+use rusoto_core::credential::{
+    AutoRefreshingProvider, AwsCredentials, CredentialsError, DefaultCredentialsProvider,
+    ProvideAwsCredentials, Secret, Variable,
+};
+use rusoto_mock::MockCredentialsProvider;
 use rusoto_sts::WebIdentityProvider;
-use std::{boxed::Box, time::Duration};
+use std::{
+    boxed::Box,
+    convert::From,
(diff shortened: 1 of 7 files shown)


facilitator/src/aws_credentials.rs
```diff
@@ -1,160 +1,261 @@
...
-    webidp_provider: WebIdentityProvider,
-}
 
-impl ChainProvider {
-    /// Set the timeout on the provider to the specified duration.
-    #[allow(dead_code)]
-    pub fn set_timeout(&mut self, duration: Duration) {
-        self.instance_metadata_provider.set_timeout(duration);
-        self.container_provider.set_timeout(duration);
+    fn new_default() -> Result<Self> {
```
aaomidi, line 79: Might be more rusty to use: https://doc.rust-lang.org/std/default/trait.Default.html

aaomidi (approved):

facilitator/src/aws_credentials.rs
tgeoghegan: The semantics are messy here: the default for this enum should probably be `new_web_identity_from_kubernetes_environment` to preserve backward compatibility, but `default` here refers to Rusoto's `DefaultCredentialsProvider`. This is one of those problems that I would classify as inherently stupid (we are forced by backward compat into making an API design that is bad), and thus any solution to it will reflect that stupidity.

aaomidi: Love it! Sounds good to me. :shipit:

tgeoghegan: The other thing I would note is that implementing `Default` is meant to be a convenience for the user of this enum, to spare them the hassle of filling in fields they don't care about or selecting an enum variant when they have no way of knowing which one they want. In this case, there is no such caller: we provide the `Provider::new()` associated function which does the right thing based on its inputs, so implementing `Default` wouldn't save anyone any hassle.

tgeoghegan:
This change needs some more testing before I can merge it to main, to make sure that it behaves correctly in the presence of `TASK_QUEUE_USE_DEFAULT_AWS_CREDENTIALS_PROVIDER` (and the storage equivalents), so I will merge once I have enough confidence that it is safe.

PR #463, opened 2021-03-05
Terraform changes to allow for integration testing

aaomidi:
- Fixes the Makefile
- Adds a new target called `portal_server_resources`.
- Changes when and how `has_test_environment` is set to true. Effectively this doesn't change the behavior of `has_test_environment`. But makes it dependent on the `fake_server_resources` module, which is what we want to get the output to update when doing `apply bootstrap`.

# Portal Server Resources 

This is a new module that creates the portal server. When the "deployment" (full dual-testing peer environments) are being deployed, this will create a fake "portal server/collector" for the aggregations of facil and pha. This isn't the **best** solution because ideally these would live inside the one environment (usually facil) that has the integration-tester set.

However, trying to get to that ideal solution through terraform seems difficult as a deployment would end up having to have an extra step just to make that work. Or it would involve manually going in and adding permission for the PHA aggregators to be able to write to the portal server. These didn't seem like ideal solutions when writing this PR, but there will be an open issue on how to improve that.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -508,7 +508,6 @@ output "singleton_ingestor" {
     batch_signing_key_name      = module.fake_server_resources[0].batch_signing_key_name
   } : {}
 }
-
 output "own_manifest_base_url" {
   value = module.manifest.base_url
 }
@@ -518,5 +517,5 @@ output "use_test_pha_decryption_key" {
 }
 
 output "has_test_environment" {
-  value = lookup(var.test_peer_environment, "env_with_ingestor", "") == var.environment
+  value = length(module.fake_server_resources) != 0
 }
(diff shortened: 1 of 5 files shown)


terraform/main.tf
```diff
@@ -518,5 +517,5 @@ output "use_test_pha_decryption_key" {
 }
 
 output "has_test_environment" {
-  value = lookup(var.test_peer_environment, "env_with_ingestor", "") == var.environment
+  value = length(module.fake_server_resources) != 0
```
aaomidi, line 520: So it seems if you don't somehow use the output of a targeted module, these never get updated.

tgeoghegan:
Looks fine but both the commit and the PR should have a more descriptive title than "Fixes some things that were missed"

tgeoghegan (approved):

tgeoghegan (approved):

PR #507, opened 2021-03-22
terraform: autoscale k8s deployments

### LINKED ISSUE #484: Autoscale deployments based on PubSub queue depth

tgeoghegan:
To spare ourselves the hassle of manually resizing deployments, we
configure Kubernetes to automatically resize `intake-batch` and
`aggregate` worker pools. This requires deploying the Stackdriver custom
metrics adapter to make PubSub metrics visible to a Kubernetes
Horizontal Pod Autoscaler, then configuring an HPA for each deployment
that consults the PubSub `num_undelivered_messages` metric. See
`terraform/README.md` for more details and discussion of config
parameter choices.

This commit also modifies the `integration-tester` deployment so that it
emits more batches, forcing dev and staging clusters to exercise the
autoscaling feature. We also amend the alert for `intake-batch` task
queue size: we no longer expect that queue to periodically empty since
we have configured Kubernetes to keep it at a steady state of ~150
messages.

Resolves #484


terraform/README.md
```diff
@@ -74,3 +74,15 @@ When instantiating a new GKE cluster, you will want to merge its configuration i
...
+
+GCP Stackdriver exposes metrics for PubSub, but to make those metrics visible to Kubernetes components, we deploy Google's [Custom Metrics Stackdriver Adapter](https://github.com/GoogleCloudPlatform/k8s-stackdriver/tree/master/custom-metrics-stackdriver-adapter), which is an implementation of the [external.metrics.k8s.io API](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-metrics-apis) that exposes Stackdriver metrics (see the `custom_metrics_adapter` module). Now that Kubernetes can see GCP metrics, we can configure HPAs to watch those metrics.
+
+We create an HPA for each Kubernetes Deployment (see the `kubernetes` Terraform module for details). We set `min_replicas` to 3 and `max_replicas` to 50. A minimum of three ensures we have reasonable minimum capacity at all times, while a maximum of 50 ensures that that we can scale up in response to really extraordinary spikes of workload while limiting the risk of flooding the Kubernetes cluster with more pods than it can possibly schedule.
```
aaomidi, line 84: Isn't max_replicas of 50 high? If I recall correctly, we've never needed more than 10, and it seems like if we're going above 10 there might be a problem that we should look at.

aaomidi (approved):

terraform/README.md
tgeoghegan: We have more than 10 intake batch workers in several states already, and 25 in California. 50 was chosen to be higher than we'd ever be likely to need but low enough that in the case of a runaway autoscale we don't take down the cluster. For the time being I'm optimizing for simplicity, hence the one-size-fits-all solution with one set of parameters for all localities. If we find this to be inefficient or unreliable then we can figure out more complex solutions, but as ever I intend to do the stupidest thing I can get away with.

I do like your idea of at least notifying us if we scale past some threshold. #508

aaomidi: Ah I see! Thanks!

PR #551, opened 2021-04-07
facilitator: better error on decryption failure

### LINKED ISSUE #550: facilitator: better error message when packet decryption fails
@hostirosti pointed out that when intake or aggregation fails because of a decryption error, you get a pretty cryptic error message like `failed to construct validation message`. This is because [`prio::server::Server::generate_verification_message`](https://github.com/abetterinternet/libprio-rs/blob/d4f97df29aab58b7ce6f0cc9263acb227e9bf9fb/src/server.rs#L93) returns `Option<VerificationMessage>` rather than `Result<VerificationMessage>`, and so absent more specific information, [we just log that there was some kind if problem creating a verification message](https://github.com/abetterinternet/prio-server/blob/7f463d67ef50e4c95cef18562ad564711eb4e57d/facilitator/src/intake.rs#L176).

We should make a change to `libprio-rs` so that this (and several other methods) return `Result<>`, and then we could provide more useful errors.

This will require a PR on `libprio-rs`, a release of that crate and then a PR on `prio-server` to adopt.

tgeoghegan:
Adopts the updated `libprio-rs` API to better discriminate between
failures of `prio::server::Server::generate_verification_message` so
that a better error message can be logged if an input share is rejected
due to a packet encryption key mismatch.

Resolves #550

### DIFF
--- a/facilitator/Cargo.lock
+++ b/facilitator/Cargo.lock
@@ -1646,9 +1646,9 @@ checksum = "ac74c624d6b2d21f425f752262f42188365d7b8ff1aff74c82e45136510a4857"
 
 [[package]]
 name = "prio"
-version = "0.2.0"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "74c293e32d9ab6c9afadf14e14d76df204f17618f4b74b38fa2adb18b352fbf2"
+checksum = "3a2d8d68c9a3ad7a7b5a5c02bfb21f0b8c1e8ba42dc36179d80765df6cc18d3e"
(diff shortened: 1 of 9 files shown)


tgeoghegan:
This won't built until `Cargo.toml` is updated to use a `prio` that contains https://github.com/abetterinternet/libprio-rs/pull/25

- [ ] bump `prio` dependency to include https://github.com/abetterinternet/libprio-rs/pull/25

aaomidi (approved):
I remember making this change with the integration tester stuff on a private branch! Glad to see we're using it now.

Issue #546, opened 2021-04-05
facilitator: plumb trace ID into all log messages


tgeoghegan:
We generate a trace ID for each task scheduled by `workflow-manager` and transmit it to `facilitator` through a message queue. Some `facilitator` log messages then log the trace ID, but we should plumb it through the entire program so that all log messages emitted by `facilitator` are annotated with the trace ID to allow for more powerful log searches.

tgeoghegan:
I'm making some good progress on this. I've got `slog-rs`, `slog-json` and `slog-term` hooked up. Example JSON log output:

```
{
    "message": "done",
    "severity": "INFO",
    "time": "2021-04-13T14:26:22.196132671-07:00",
    "facilitator_output_path": "/tmp/facil-out",
    "pha_output_path": "/tmp/pha-out",
    "batch_date": "2021/04/13/21/26",
    "batch_id": "42c040d1-b1e8-42c4-bd0e-cc5971e4afa9",
    "trace_id": "0c64c8c3-69af-43b5-9326-abe696650c32",
    "aggregation_name": "test-aggregation",
    "thread_id": "ThreadId(2)",
    "module_path": "facilitator::sample",
    "line": 362,
    "file": "src/sample.rs",
    "version": "(BUILD_INFO unavailable)"
}
```
Or, if `stderr` is a tty, you get:

![Screenshot-20210413143705-1842x153](https://user-images.githubusercontent.com/19212995/114624547-bbb89c00-9c65-11eb-9fe0-dd0d03bfdfe3.png)

I still need to figure out how to handle errors (which at a minimum need the trace ID on them), and how to carve the 3,000+ line change into multiple PRs.

tgeoghegan:
This is going to require a few PRs, because I don't want to overwhelm my gracious and patient reviewers with a massive change. 

1. I needed to refactor sample generation to be oriented around a struct, so that we would eventually have somewhere to stick a `slog::Logger`: #559.
1. I learned that generating sample data at the command line got broken somewhere along the way, so I had to make some fixes to that: #564 
[...]

PR #565, opened 2021-04-14
facilitator: use slog for structured logging

tgeoghegan:
We adopt the `slog-rs` family of crates for structured logging. `slog`
allows us to create trees of `Loggers` and `Drains`. `Logger`s can be
annotated with key-value pairs and then every message logged through
that `Logger` will themselves be annotated with the k-v pairs. Further,
the `slog` macros `info`, `error`, etc. allow adding arbitrary
additional k-v pairs to individual events. This lets us emit structured
log messages which can then be searched in something like the GCP Log
Explorer. `Drain`s can be configured to format log records in different
manners. We use `slog_json` to emit structured JSON logs, `slog_term` to
pretty print messages to a tty or a much simpler decorator for use in
tests.

`slog`'s model is that you create a root `Logger`, and then create a
tree of child loggers from it, each of which can be specialized with
context specific to some module or task. For instance, in this commit,
we augment `aggregation::BatchAggregator` so that it creates a `Logger`
from a provided parent logger, and annotates it with a trace ID and
other interesting values.

To ease the transition from the existing logging model to `slog`, we use
`slog_scope` as a drop-in replacement for the `log`/`env_logger` combo
previously employed in `facilitator`. Subsequent PRs will plumb `Logger`
instances into each module of `facilitator` and eventually `slog_scope`
should be removed altogether.

This commit is a start on #546.

This PR is stacked on top of #559 and #564


tgeoghegan:
Actually, let's land the other two changes before I bombard my graceful, patient reviewers with this one.

PR #566, opened 2021-04-14
facilitator: use slog for structured logging

tgeoghegan:
We adopt the `slog-rs` family of crates for structured logging. `slog`
allows us to create trees of `Loggers` and `Drains`. `Logger`s can be
annotated with key-value pairs and then every message logged through
that `Logger` will themselves be annotated with the k-v pairs. Further,
the `slog` macros `info`, `error`, etc. allow adding arbitrary
additional k-v pairs to individual events. This lets us emit structured
log messages which can then be searched in something like the GCP Log
Explorer. `Drain`s can be configured to format log records in different
manners. We use `slog_json` to emit structured JSON logs, `slog_term` to
pretty print messages to a tty or a much simpler decorator for use in
tests.

`slog`'s model is that you create a root `Logger`, and then create a
tree of child loggers from it, each of which can be specialized with
context specific to some module or task. For instance, in this commit,
we augment `aggregation::BatchAggregator` so that it creates a `Logger`
from a provided parent logger, and annotates it with a trace ID and
other interesting values.

To ease the transition from the existing logging model to `slog`, we use
`slog_scope` as a drop-in replacement for the `log`/`env_logger` combo
previously employed in `facilitator`. Subsequent PRs will plumb `Logger`
instances into each module of `facilitator` and eventually `slog_scope`
should be removed altogether.

This commit is a start on #546

### DIFF
--- a/facilitator/src/aggregation.rs
+++ b/facilitator/src/aggregation.rs
@@ -4,6 +4,10 @@ use crate::{
         IngestionDataSharePacket, IngestionHeader, InvalidPacket, Packet, SumPart,
         ValidationHeader, ValidationPacket,
     },
+    logging::{
+        EVENT_KEY_AGGREGATION_NAME, EVENT_KEY_INGESTION_PATH, EVENT_KEY_OWN_VALIDATION_PATH,
+        EVENT_KEY_PACKET_UUID, EVENT_KEY_PEER_VALIDATION_PATH, EVENT_KEY_TRACE_ID,
+    },
     metrics::AggregateMetricsCollector,
     transport::{SignableTransport, VerifiableAndDecryptableTransport, VerifiableTransport},
     BatchSigningKey, Error,
@@ -11,11 +15,11 @@ use crate::{
 use anyhow::{anyhow, Context, Result};
 use avro_rs::Reader;
 use chrono::NaiveDateTime;
-use log::info;
 use prio::{
     field::Field32,
     server::{Server, VerificationMessage},
 };
+use slog::{info, o, Logger};
 use std::{
     collections::{HashMap, HashSet},
     convert::TryFrom,
@@ -37,6 +41,7 @@ pub struct BatchAggregator<'a> {
     share_processor_signing_key: &'a BatchSigningKey,
     total_individual_clients: i64,
     metrics_collector: Option<&'a AggregateMetricsCollector>,
+    logger: Logger,
 }
 
 impl<'a> BatchAggregator<'a> {
@@ -53,7 +58,15 @@ impl<'a> BatchAggregator<'a> {
         own_validation_transport: &'a mut VerifiableTransport,
         peer_validation_transport: &'a mut VerifiableTransport,
         aggregation_transport: &'a mut SignableTransport,
+        parent_logger: &Logger,
     ) -> Result<BatchAggregator<'a>> {
+        let logger = parent_logger.new(o!(
+            EVENT_KEY_TRACE_ID => trace_id.to_owned(),
+            EVENT_KEY_AGGREGATION_NAME => aggregation_name.to_owned(),
+            EVENT_KEY_INGESTION_PATH => ingestion_transport.transport.transport.path(),
+            EVENT_KEY_OWN_VALIDATION_PATH => own_validation_transport.transport.path(),
+            EVENT_KEY_PEER_VALIDATION_PATH => peer_validation_transport.transport.path(),
+        ));
         Ok(BatchAggregator {
             trace_id,
             is_first,
@@ -77,6 +90,7 @@ impl<'a> BatchAggregator<'a> {
             share_processor_signing_key: &aggregation_transport.batch_signing_key,
             total_individual_clients: 0,
             metrics_collector: None,
+            logger,
         })
     }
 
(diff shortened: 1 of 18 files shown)


facilitator/src/logging.rs
```diff
@@ -1,67 +1,169 @@
...
-            module_path: record.module_path(),
-            file: record.file(),
-            line: record.line(),
-            thread_id: format!("{:?}", thread::current().id()),
-        }
-    }
+/// Initialize logging resources. On success, returns a root [`slog::Logger`][1]
+/// from which child loggers may be built and a GlobalLoggerGuard which must be
+/// kept live by the caller to enable the `slog_scope` global logger to function
+/// in modules that haven't yet opted into managing their own `slog::Logger`.
```
jsha, line 93: This is a bear of a sentence tree! Can you break it up a little?

facilitator/src/logging.rs
```diff
@@ -1,67 +1,169 @@
...
+/// Returns an error if `LoggingConfiguration` is invalid.
+///
+/// [1]: https://docs.rs/slog/2.7.0/slog/struct.Logger.html
+pub fn setup_logging(config: &LoggingConfiguration) -> Result<(Logger, GlobalLoggerGuard)> {
+    // We have to box the Drain so that both branches return the same type
+    let drain: Box<dyn Drain<Ok = (), Err = std::io::Error> + Send> =
```
jsha, line 99: Maybe alias the Drain type?

```
type MyDrain = Drain<Ok = (), Err = std::io::Error>;
```

facilitator/src/logging.rs
```diff
@@ -1,67 +1,169 @@
...
+            let json_drain = Json::new(stderr())
+                .set_newlines(true)
+                // slog_json::JsonBuilder::add_default_keys adds keys for
+                // timestamp, level and the actual message, but we need the JSON
+                // keys to match what the Google Cloud Logging agent expects:
+                // https://cloud.google.com/logging/docs/agent/configuration#process-payload
+                .add_key_value(o!(
+                    "time" => PushFnValue(|_, serializer| {
+                        serializer.emit(chrono::Local::now().to_rfc3339())
+                    }),
```
jsha, line 112: https://docs.rs/slog/2.7.0/slog/struct.PushFnValue.html says:

> It's more ergonomic for closures used as lazy values to return type implementing Serialize, but sometimes that forces an allocation (eg. Strings)

Since this is called once, I don't think we're worried about allocations. Can you use `chrono::Local::now().to_rfc3339()` directly? Presumably its output is String and String implements Serialize?

jsha (changes requested):
Generally looks good!

facilitator/src/aggregation.rs
```diff
@@ -53,7 +58,15 @@ impl<'a> BatchAggregator<'a> {
...
         peer_validation_transport: &'a mut VerifiableTransport,
         aggregation_transport: &'a mut SignableTransport,
+        parent_logger: &Logger,
     ) -> Result<BatchAggregator<'a>> {
+        let logger = parent_logger.new(o!(
+            TRACE_ID_EVENT_KEY => trace_id.to_owned(),
+            AGGREGATION_NAME_EVENT_KEY => aggregation_name.to_owned(),
+            INGESTION_PATH_EVENT_KEY => ingestion_transport.transport.transport.path(),
+            OWN_VALIDATION_PATH_EVENT_KEY => own_validation_transport.transport.path(),
+            PEER_VALIDATION_PATH_EVENT_KEY => peer_validation_transport.transport.path(),
```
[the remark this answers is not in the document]
tgeoghegan: You're right, and it's better for autocomplete, too.

facilitator/src/logging.rs
```diff
@@ -1,67 +1,169 @@
...
+/// The name of the aggregation
+pub(crate) const AGGREGATION_NAME_EVENT_KEY: EventKey = "aggregation_name";
+/// The storage path from which ingestion batches are read/written
+pub(crate) const INGESTION_PATH_EVENT_KEY: EventKey = "ingestion_path";
+/// The storage path from which own validation batches are read/written
+pub(crate) const OWN_VALIDATION_PATH_EVENT_KEY: EventKey = "own_validation_path";
+/// The storage path from which peer validation batches are read/written
+pub(crate) const PEER_VALIDATION_PATH_EVENT_KEY: EventKey = "peer_validation_path";
+/// The UUID of a packet that something happened to
+pub(crate) const PACKET_UUID_EVENT_KEY: EventKey = "packet_uuid";
```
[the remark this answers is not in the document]
tgeoghegan: We have to make items `pub` to be visible to `src/bin/facilitator.rs`, but the others are used from other modules in the crate. My thinking is to make things as private as I possibly can. I need to double check whether `src/bin/` can see `pub(crate)` items though.

facilitator/src/logging.rs
tgeoghegan: [You cannot currently define an alias on traits](https://doc.rust-lang.org/nightly/unstable-book/language-features/trait-alias.html). The best you can do is define a supertrait that captures all the traits you want to alias, but then you have to opt all the types you want to describe with the supertrait in with empty impls. I gave this a try in 4dd69fab579a1665f93e940670d08730bb6fe15d (which I separated from the changes addressing your other comments) and it works, but I'm not sure I like it better given the verbosity.

facilitator/src/logging.rs
tgeoghegan: I rewrote the doccomment. Possibly this would be even better if `setup_logging` returned a struct wrapping a `slog::Logger` and a `GlobalLoggerGuard`, and then you could have nice doccomments on that, but I will be getting rid of the `slog_scope` logger in favor of opting everything into child `slog::Loggers` very soon.

jsha (approved):

facilitator/src/logging.rs
jsha: Good point about the problem of trait aliases. I kinda like the version you wound up with, but not enough to stand on principle about it; either way is fine.

facilitator/src/logging.rs
jsha: I see a 👍🏻 on this but not a corresponding change in the latest commit.

aaomidi (approved):

facilitator/src/logging.rs
tgeoghegan: Well, I had fun figuring it out, and it seems courteous to defer to the reviewer, so I'll the supertrait.

facilitator/src/logging.rs
tgeoghegan: We now capture the time without the serializer:
```
                "time" => FnValue(|_| {
                    chrono::Local::now().to_rfc3339()
                }),
```

PR #572, opened 2021-04-15
deploy-tool: key backups

### LINKED ISSUE #491: Backup Kubernetes secrets in cloud secrets storage
We use Kubernetes secrets to store batch signing keys and packet decryption keys. It's inconvenient if the batch signing keys are ever lost (you'd have to generate a new one and advertise it in a manifest alongside the old one, which our tools don't currently support), but losing the packet decryption keys means that some amount of data send by ingestors will be irrevocably lost.

Just to make it possible for us to recover from disasters, we should teach `deploy-tool` and `deploy-operator` to back up keys they create into some secure storage outside of the blast radius of the Kubernetes cluster, such as the cloud platform's secrets management service.

tgeoghegan:
For disaster recovery, `deploy-tool` can now back up batch signing and
packet decryption keys into GCP Secret Manager. See
`terraform/README.md` for details. For the time being , we only support
backing up secrets to GCP, but could add support for other cloud secret
stores.

Resolves #491


deploy-tool/key/key_backup.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+		if version.GetState() == secretmanagerpb.SecretVersion_ENABLED {
+			return secretAndVersionExist, nil
+		}
+	}
+}
+
+// BackupSecret backs up the provided Kubernetes secret into GCP Secret Manager,
+// if it is not already backed up.
+func (b *GCPSecretManagerBackup) BackupSecret(k8sSecret *k8sv1.Secret) error {
```
aaomidi, line 125: This should take in a context argument I believe and pass it around to the calls that use a context argument.

aaomidi (approved):
:shipit:

Issue #290, opened 2020-12-21
Tool for backfilling intakes and aggregations


tgeoghegan:
If we fall far enough behind on intake or aggregations for some reason, there's currently no way to backfill the data. It should be possible to write a tool that can take date or batch ID parameters and schedule `intake-batch` or `aggregate` jobs that were missed. This could be a subcommand on `workflow-manager`, since it already has logic for enumerating buckets looking for batches, scheduling jobs, etc. In my head this is a tool that an operator would run on their machine which schedules Kubernetes jobs, rather than something that runs as a cronjob in the cluster.

tgeoghegan:
Over the weekend, I wrote up a utility to replay JSON tasks into the system (#576). It doesn't completely address this problem because it still requires manually fetching failed tasks from dead letter queues and replaying them, so I will keep this open even after merging #576 to explore further improvements to backfills.

BranLwyd:
The above commit against `workflow-manager` will allow aggregation to be triggered against arbitrary aggregation windows. This is sufficient to allow operators to backfill aggregations.

PR #576, opened 2021-04-19
task-replayer: simple utility for replaying tasks

tgeoghegan:
We use message queues (GCP PubSub or AWS SNS/SQS) to manage tasks in
`prio-server`. After a task is retried five times, it gets delivered to
a dead-letter queue for triage. In some cases, the thing to do is just
to replay the failed task into the queue so it can be tried again,
perhaps because a change has been deployed that will now let the task
succeed.

This commit adds a Go utility `task-replayer` which permits operators to
replay a task from a JSON file. See `task-replayer/README.md` or
`task-replayer --help` for details.

Relates to #290


aaomidi (approved):

aaomidi:
Would be nice to get this automatically building as part of our CI pipeline too

tgeoghegan:
> Would be nice to get this automatically building as part of our CI pipeline too

Great idea, I'll tack on a commit to this PR to do that.

Issue #353, opened 2021-02-04
facilitator: retries in GCP token fetcher


tgeoghegan:
Code in `gcp_oauth.rs` that obtains Oauth tokens from various GCP APIs should have appropriate retry logic.

tgeoghegan:
Resolved by #578

Issue #352, opened 2021-02-04
facilitator: retries in GCS client


tgeoghegan:
We should research best practices for Google Cloud Storage clients and implement appropriate retry logic.

tgeoghegan:
[Google Cloud Storage retry strategy](https://cloud.google.com/storage/docs/retry-strategy)
[status and error codes for the JSON storage API](https://cloud.google.com/storage/docs/json_api/v1/status-codes)
[GCS Go SDK's `shouldRetry` function](https://github.com/googleapis/google-cloud-go/blob/master/storage/go110.go)

tgeoghegan:
Resolved by #578

Issue #351, opened 2021-02-04
facilitator: retries in GCP PubSub client


tgeoghegan:
We should research best practices for GCP PubSub clients and implement appropriate retry logic.

tgeoghegan:
[PubSub error codes reference](https://cloud.google.com/pubsub/docs/reference/error-codes)
[PubSub Go SDK retry logic](https://github.com/googleapis/google-cloud-go/blob/master/pubsub/apiv1/subscriber_client.go)

tgeoghegan:
Resolved by #578

PR #578, opened 2021-04-20
facilitator: introduce `http::RetryingAgent`

tgeoghegan:
In order to be more robust against transient failures in cloud APIs, we
introduce a `RetryingAgent` which uses the `backoff` crate to resend
failed HTTP requests with exponential backoff if the failure is
determined to be transient (e.g., an HTTP 503). The `RetryingAgent` is
adopted throughout the crate, meaning we now get retries when:

  - fetching peer JSON manifests
  - pulling messagesfrom GCP PubSub subscriptions
  - reading or writing to Google Cloud Storage
  - obtaining Oauth tokens from the GKE metadata service or IAM API

We should also record Prometheus metrics about retryable request
failures, which will happen in a future PR.

This change doesn't help with AWS API clients (S3, SQS and IAM) because
they do not use our `ureq` based HTTP client but rather Rusoto (and that
in turn uses Hyper).

Related to #351, #352, #353

### DIFF
--- a/facilitator/src/http.rs
+++ b/facilitator/src/http.rs
@@ -1,8 +1,9 @@
 use anyhow::{Context, Result};
-use std::fmt::Debug;
-use std::time::Duration;
+use std::{convert::From, default::Default, fmt::Debug, time::Duration};
 use ureq::{Agent, AgentBuilder, Request, Response, SerdeValue};
 use url::Url;
+
+use crate::retries::retry_request;
 
 /// Method contains the HTTP methods supported by this crate.
 #[derive(Debug)]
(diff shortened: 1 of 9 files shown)


facilitator/src/http.rs
```diff
@@ -25,23 +26,112 @@ impl Method {
...
+/// An HTTP agent that can be configured to manage "Authorization" headers and
+/// retries using exponential backoff.
+#[derive(Debug, Clone)]
+pub(crate) struct RetryingAgent {
+    /// Agent to use for constructing HTTP requests.
+    agent: Agent,
+    /// By default, requests which fail due to transport problems or which
+    /// return any 5xx HTTP code will be retried with exponential backoff. If
+    /// set, these HTTP codes will also cause retries.
+    additional_retryable_http_codes: Option<Vec<u16>>,
```
jsha, line 38: `Vec<u16>` seems enough, since the vec can be empty.

facilitator/src/http.rs
```diff
@@ -25,23 +26,112 @@ impl Method {
...
+    fn is_http_status_retryable(&self, http_status: u16) -> bool {
+        if http_status >= 500 {
+            return true;
+        }
+
+        if let Some(codes) = &self.additional_retryable_http_codes {
+            return codes.contains(&http_status);
+        }
+
+        false
```
jsha, line 86: If you make `additional...` a Vec, this can be simplified to:

```
http_status > 500 || self.additional_retryable_http_codes.contains(&http_status)
```

facilitator/src/http.rs
```diff
@@ -25,23 +26,112 @@ impl Method {
...
+
+    fn is_error_retryable(&self, error: &ureq::Error) -> bool {
+        match error {
+            ureq::Error::Status(http_status, _) => self.is_http_status_retryable(*http_status),
+            ureq::Error::Transport(_) => true,
+        }
+    }
+
+    /// Send the provided request with the provided JSON body.
+    pub(crate) fn send_json_request(&self, request: Request, body: SerdeValue) -> Result<Response> {
```
jsha, line 97: If you're cloning Request and SerdeValue anyhow, why take ownership? I think these should be `&`

jsha (approved):
Looks good to me!

facilitator/src/http.rs
tgeoghegan: Excellent suggestion, especially given how it simplifies `is_http_status_retryable`.

facilitator/src/http.rs
tgeoghegan: I was somewhat blindly replicating the API of the underlying `ureq::Request` methods, but you are right, these can be references.

PR #579, opened 2021-04-20
facilitator: structured logging in intake-batch

tgeoghegan:
Augments `intake::BatchIntaker` with a `slog::Logger` field, to which we
attach specialized labels.

Also provides a `slog::Value` implementation on `task::TaskHandle` so
that `TaskHandle<T>` values can easily be inserted into events.

Relates to #546


aaomidi (approved):

aaomidi:
This looks good to me, I didn't look into the log calls in every file, but I understand PRs like this just involve changing every file.

PR #595, opened 2021-04-23
facilitator: batch and transport use slog::Logger

tgeoghegan:
Augments `batch::{BatchReader, BatchWriter}` as well as all of the
`transport::Transport` implementations to use `slog::Logger`.


jsha:
You've got conflicts. :-)

tgeoghegan:
> You've got conflicts. :-)

So I did -- fixed.

facilitator/src/batch.rs
```diff
@@ -106,14 +107,22 @@ impl Batch {
     }
 }
 
+impl Display for Batch {
```
jsha, line 110: Seems like this might be better as `Debug`? I can imagine accidentally calling `to_string()` on a Batch thinking it's some other type, and not realizing that reduces to its header key. Or (looking at your call site) making `header_key` public and calling it directly?

jsha (approved):
Looks good! One thing I noticed, not particular to this PR: The `EVENT_KEY_` prefix leads to some ambiguous parses, like `EVENT_KEY_IDENTITY` -- is that "event key_identity", or "event_key identity". I know it's the latter, but could be confusing to a future ready.

One fix would be to put the event keys in their own module, and only ever import the module, not the items. So you'd refer to things like `event::IDENTITY` and `event::STORAGE_KEY`

facilitator/src/batch.rs
tgeoghegan: > One fix would be to put the event keys in their own module, and only ever import the module, not the items. So you'd refer to things like `event::IDENTITY` and `event::STORAGE_KEY`

This is a great idea. I made a module `logging::event` that defines the event key constants and it's much nicer. That's in a follow up PR which I will link here once I create it.

> Seems like this might be better as Debug? I can imagine accidentally calling to_string() on a Batch thinking it's some other type, and not realizing that reduces to its header key. Or (looking at your call site) making header_key public and calling it directly?

I think I did this as a way to get something that implements `slog::Value`, but you're right, I can just as easily call `header_key().to_owned()` rather than provide a flawed `Display` impl.

Issue #354, opened 2021-02-04
facilitator: retries in AWS SQS client


tgeoghegan:
We should research and implement appropriate retry logic in our AWS SQS client (if Rusoto makes it possible to see stuff like HTTP status codes).

tgeoghegan:
Resolved by #597

PR #597, opened 2021-04-26
facilitator: retries w/ backoff for AWS

### LINKED ISSUE #41: Implement retries in S3 requests

tgeoghegan:
Adapts `aws_credentials::retry_request` to use mod `retries` (and hence
`backoff::ExponentialBackoff` so that we get more graceful retry
behavior, in line with AWS developer guidelines. Additionally, our
assessment of which errors are retryable is more sophisticated: we now
retry on all credential fetch failures, on server errors (i.e., HTTP
5xx) and when throttled.

Resolves #41, #354

### DIFF
--- a/facilitator/Cargo.lock
+++ b/facilitator/Cargo.lock
@@ -645,6 +645,7 @@ dependencies = [
  "avro-rs",
  "backoff",
  "base64 0.13.0",
+ "bytes",
  "chrono",
  "clap",
  "derivative",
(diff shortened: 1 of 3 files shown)


jsha (approved):
Looks good!

PR #602, opened 2021-04-28
facilitator: move event keys to `logging::event`

tgeoghegan:
Per @jsha's suggestion in #595, we now define the various `EVENT_KEY_*`
constants in the `logging::event` module, allowing us to rename the
constants from `EVENT_KEY_THING` to just `THING`.


jsha (approved):
🎉

Issue #543, opened 2021-04-05
facilitator: improve sharing of cached credentials


tgeoghegan:
In #486, I tried to make it easy to share credentials across threads by making sure all structs that handle creds are `Clone`. However, we still don't share credentials particularly effectively: empirically, from examining logs, we see far more traffic to the GKE metadata endpoint and AWS STS than should be happening, since most credentials are valid for one hour after issuance. By source inspection, I think this is because we don't clone `GcpOauthTokenProvider` or `aws_credentials::Provider` as aggressively as we should.

- We create a distinct `Transport` for each of intake, peer validation and own validation in both `facilitator::intake_batch` and `facilitator::aggregate`, and each `Transport` has a different token provider.
- We don't re-use `Transport`s or token providers across iterations of the loops in `intake_batch_worker`, `aggregate_worker` or `generate_sample_worker`

PR #605, opened 2021-04-28
facilitator: implement `Clone` on creds providers

tgeoghegan:
As we work towards #543, we need efficient `std::clone::Clone`
implementations on our `Transport` and `TaskQueue` implementations so
they can be efficiently shared across tasks. This commit starts towards
that goal by making the GCP and AWS credential providers `Clone`.
`aws_credentials::Provider` was already `Clone`, mostly thanks to
Rusoto's underlying implementations, but
`gcp_oauth::GcpOauthTokenProvider` needed some rework to implement
threadsafe interior mutability. I also took the opportunity to add some
test coverage.

This commit consists mainly of changes cherry-picked from #486.

### DIFF
--- a/facilitator/src/gcp_oauth.rs
+++ b/facilitator/src/gcp_oauth.rs
@@ -1,33 +1,57 @@
 use anyhow::{anyhow, Context, Result};
 use chrono::{prelude::Utc, DateTime, Duration};
+use dyn_clone::DynClone;
 use jsonwebtoken::{encode, Algorithm, EncodingKey, Header};
 use rusoto_core::{credential::ProvideAwsCredentials, Region};
 use serde::{Deserialize, Serialize};
 use slog::{debug, o, Logger};
-use std::{fmt, io::Read, str};
+use std::{
+    fmt::{self, Debug},
+    io::Read,
+    str,
+    sync::{Arc, RwLock},
+};
 use ureq::Response;
 use url::Url;
 
 use crate::{
-    aws_credentials::{basic_runtime, get_caller_identity_token},
+    aws_credentials::{self, basic_runtime, get_caller_identity_token},
     config::WorkloadIdentityPoolParameters,
     http::{
         Method, OauthTokenProvider, RequestParameters, RetryingAgent, StaticOauthTokenProvider,
     },
 };
 
-fn default_oauth_token_url() -> Url {
-    Url::parse("http://metadata.google.internal:80/computeMetadata/v1/instance/service-accounts/default/token",)
-    .expect("unable to parse metadata.google.internal url")
+const DEFAULT_METADATA_BASE_URL: &str = "http://metadata.google.internal:80";
+const DEFAULT_TOKEN_PATH: &str = "/computeMetadata/v1/instance/service-accounts/default/token";
+const DEFAULT_IAM_BASE_URL: &str = "https://iamcredentials.googleapis.com";
+
+fn default_oauth_token_url(base: &str) -> Url {
+    let mut request_url = Url::parse(base).expect("unable to parse metadata.google.internal url");
+    request_url.set_path(DEFAULT_TOKEN_PATH);
+    request_url
 }
 
 // API reference:
 // https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken
-fn access_token_url_for_service_account(service_account_to_impersonate: &str) -> Result<Url> {
-    let request_url = format!("https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{}:generateAccessToken",
-            service_account_to_impersonate);
+fn access_token_url_for_service_account(
+    base: &str,
+    service_account_to_impersonate: &str,
+) -> Result<Url> {
+    let request_url = format!(
+        "{}{}",
+        base,
+        access_token_path_for_service_account(service_account_to_impersonate)
+    );
 
     Url::parse(&request_url).context(format!("failed to parse: {}", request_url))
+}
+
+fn access_token_path_for_service_account(service_account_to_impersonate: &str) -> String {
+    format!(
+        "/v1/projects/-/serviceAccounts/{}:generateAccessToken",
+        service_account_to_impersonate
+    )
 }
 
 /// Represents the claims encoded into JWTs when using a service account key
@@ -42,6 +66,7 @@ struct Claims {
 }
 
 /// A wrapper around an Oauth token and its expiration date.
+#[derive(Clone)]
 struct OauthToken {
     token: String,
     expiration: DateTime<Utc>,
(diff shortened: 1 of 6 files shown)


facilitator/src/gcp_oauth.rs
```diff
@@ -393,13 +452,14 @@ impl GcpOauthTokenProvider {
     /// The returned value is an owned reference because the token owned by this
     /// struct could change while the caller is still holding the returned token
     fn ensure_default_account_token(&mut self) -> Result<String> {
-        if let Some(token) = &self.default_account_token {
+        if let Some(token) = &*self.default_account_token.read().unwrap() {
```
aaomidi, line 455: nit: This took me a couple of seconds to understand, it might be nice if this was 

```
&(*self.default_account_token.read().unwrap())
```

or it might be worse, honestly no idea.

aaomidi (approved):

facilitator/src/gcp_oauth.rs
```diff
@@ -1,35 +1,59 @@
 use anyhow::{anyhow, Context, Result};
 use chrono::{prelude::Utc, DateTime, Duration};
+use dyn_clone::DynClone;
```
aaomidi, line 3: This is a really cool library

facilitator/src/gcp_oauth.rs
```diff
@@ -186,14 +225,15 @@ impl ProvideDefaultToken for ServiceAccountKeyFileDefaultTokenProvider {
 
 /// Uses a GCP Workload Identity Pool to federate GCP IAM with AWS IAM, allowing
 /// workloads in AWS Elastic Kubernetes Service to impersonate a GCP SA.
+#[derive(Clone)]
 struct AwsIamFederationViaWorkloadIdentityPoolDefaultTokenProvider {
-    aws_credentials_provider: Box<dyn ProvideAwsCredentials>,
+    aws_credentials_provider: aws_credentials::Provider,
```
aaomidi, line 230: I would personally feel more comfortable putting this in an Arc if we're going to clone this struct. I did read that internally they're using Arc, but if they make an implementation mistake in the future this is going to crash on its head.

Although, I have no idea what kind of performance impact making this Arc would have.

Either way, I think it's fine whatever decision you make.

facilitator/src/gcp_oauth.rs
```diff
@@ -393,13 +452,14 @@ impl GcpOauthTokenProvider {
...
-        if let Some(token) = &self.default_account_token {
+        if let Some(token) = &*self.default_account_token.read().unwrap() {
             if !token.expired() {
                 debug!(self.logger, "cached default account token is still valid");
                 return Ok(token.token.clone());
             }
         }
 
+        let mut default_account_token = self.default_account_token.write().unwrap();
         let http_response = self.default_token_provider.default_token()?;
```
jcjones, line 463: Now that we acquired the write lock, we need to double check that it's still expired. If we have contention, we could have multiple threads who all independently observed in the `read` lock that it was expired, but one already made it into the `write` and updated it.

At least, I'm pretty sure that's how the semantics of the above block work, since we don't keep the RwLockReadGuard in-scope.

jcjones (changes requested):
I am sorry I don't have a full review here yet, but I wanted to at least bring this to your attention before the weekend. I'll come back to this ASAP.

facilitator/src/gcp_oauth.rs
tgeoghegan: That's an interesting idea. Even if we did wrap every variant of `aws_credentials::Provider` (which is a type we define) in `std::sync::Arc`, all that would guarantee is that we can call `clone()` and have... something happen. It wouldn't protect us against a change in the implementation of say `rusoto_core::credentials::DefaultCredentialsProvider` that causes it to no longer use an `Arc<Mutex<>>` internally to efficiently share its state across instances. In fact, wrapping stuff in `Arc` would make it harder for us to detect that Rusoto had made a change such that any of `DefaultCredentialsProvider`, `AutoRefreshingProvider` or `WebIdentityProvider` was no longer `Clone`.

I'd argue that generally speaking, we can't always effectively defend ourselves today against arbitrary performance or correctness regressions introduced by our dependencies in the future, and we have to balance the risk of trusting dependencies to do their job correctly against the benefit of all the code they provide us. "Hell is other programmers", as Sartre would remind us.

facilitator/src/gcp_oauth.rs
tgeoghegan: This is a very good point. The tricky thing is that I don't think you can acquire the write lock while the read lock is held. I think that at the time that I wrote this, I convinced myself that it's OK for two threads to race to acquire the write lock and update the access token, because regardless of who wins (1) both threads leave with a valid access token and (2) self.default_account_token will contain a valid access token. However I think the more obvious and safer thing to do would be to use an `std::sync::Mutex` here instead of `RwLock` so that we don't have to juggle two locks, at the cost of more lock contention on the read path.

jcjones: All you have to do is double-check `token.expired()` after acquiring the write lock and before proceeding. It is a little more expensive but it follows the general threading model of keeping the common case fast, and assuming success but checking for failure.

Changing to a Mutex seems overkill if this is the only place we encounter issues.

tgeoghegan: I must be missing something. If `token` is in scope past the acquisition of `self.default_account_token.write()`, then `self.default_account_token().read()` will still be held and we'll deadlock because you can't acquire a write lock if there are any live readers.

jcjones: Insert after line 462 a check of expired and return early if the precondition no longer holds. It's a waste of time having obtained a write lock only to abandon ship, but it is consistent.

tgeoghegan: Oh, duh, I was missing something: for some reason my brain decided that only the read lock can let you read but of course you can check the value with the write lock.

facilitator/src/gcp_oauth.rs
tgeoghegan: Yes, I was really glad it exists! Makes me think that maybe there is a missing language feature here, if `DynClone` can be implemented for arbitrary traits.

facilitator/src/gcp_oauth.rs
tgeoghegan: I tried rewriting it as
```
if let Some(token) = self.default_account_token.read().unwrap().deref().as_ref() { ... }
```
...and I don't think it's anymore clear: if I don't know how `std::sync::RwLock` works, I still have to trawl through `std::sync` docs to understand what `as_ref` does. I also really don't like that I'm forced to insert an `unwrap()` call, which IMO is usually a sign that you're doing something sketchy.

I think Rust programmers just have to get used to using `*` to dereference `RwLock{Read|Write}Guard<T>` and `MutexGuard<T>` into `T`. I do with those types provided an explicit method like `deref_mutex()` that would make this more obvious for those of us who weigh readability over terseness but given the numerous examples in Rustdoc, this form seems to be the least surprising for Rust programmers.

jcjones (approved):
LGTM, thanks!

PR #608, opened 2021-04-28
facilitator: use `slog::Logger` in `http` module

tgeoghegan:
Plumbs structured logging into `http::RetryingAgent`, requiring the same
to be passed through other modules that depend on `RetryingAgent`.
`RetryingAgent::logger` isn't yet used; that will come in a subsequent
PR that passes the logger into `retries::retry_request`.

Part of #546


tgeoghegan:
One more PR after this to wrap up #546. Preview: https://github.com/abetterinternet/prio-server/pull/609

aaomidi (approved):

Issue #655, opened 2021-05-21
Refactor Terraform to support AWS EKS


tgeoghegan:
We want to be able to deploy `prio-server` clusters that run on AWS, using EKS, S3 and SNS/SQS. There should be a top-level parameter defined in an env's .tfvars file that governs whether the cluster is on GKE or EKS.

tgeoghegan:
Progress update: I've got a whole bunch of Terraform changes to let us conditionally spin up either an EKS or GKE cluster based on the top-level `use_aws` variable. The difficulty is that Terraform is kinda hostile to refactoring, which means that applying these changes to existing environments (like prod or staging) would require a whole lot of perilous `terraform import` work to account for the changes to the layout of objects in Terraform (i.e., some resources that have moved from one module to another, or whose TF identifier now includes a `[0]` because we use a `count = some_condition ? 1 : 0` to conditionally create them). I'm thinking of just keeping the old and new TF modules in git so we don't have to block deployment of international on gracefully migrating the prod-us deployment.

Beyond that I have also updated deploy-tool so it can wrangle manifests in AWS S3 (that's #694). All those changes are in #699.

At this point, all the jobs and deployments in the EKS cluster launch. `workflow-manager` works normally and can scan buckets for tasks and schedule tasks (i.e., publish messages to SNS). However the `intake-batch` works fail when trying to authenticate to GCP in order to write objects to the peer validation bucket. So next up, I need to implement #681.

tgeoghegan:
Resolved in #802

PR #703, opened 2021-06-07
deploy-tool: post manifests to S3 buckets

### LINKED ISSUE #694: Update deploy-tool to post manifests to S3
`deploy-tool` is only capable of uploading data share processor specific manifests to GCS. It needs to learn to detect if the manifest bucket is in S3 and upload there instead.

tgeoghegan:
Teaches `deploy-tool` to manage data share processor specific manifests
in AWS S3 buckets, as well as the GCS buckets it already knew how to
deal with. To support this, `deploy-tool` needs to be provided the AWS
region of the bucket and a profile to obtain credentials for AWS, so
this commit introduces just enough Terraform changes so that further
changes that will land for #655 can express those values.

Resolves #694


tgeoghegan:
I'm not allowed to formally request a review but @jcjones or @ezekiel please take a look (or feel free to pick my brain about WTF this change is for).

jcjones (approved):
At the core this is nice and simple, sorry it took me so long to review, I wanted to make sure I understood manifest-updater in general before trying to review this code.

tgeoghegan:
No sweat! In fact we should collectively work out a plan for what to do with manifest-updater since we will need to address key rotation soon.

PR #728, opened 2021-06-15
facilitator: client for GCP auth with workload identity pool

### LINKED ISSUE #681: facilitator: implement workload identity federation to use GCP from AWS
`prio-server` deployments running on either AWS or GCP need to access resources hosted on the other cloud platform. For workloads running in GCP accessing AWS, we use OIDC to federate identity between AWS IAM and accounts.google.com. For the other direction (AWS workloads accessing GCP resources), we currently distribute GCP service account key files to AWS EKS workloads and authenticate to GCP that way. To avoid dealing with managing those credentials, we should instead use [workload identity federation](https://cloud.google.com/iam/docs/access-resources-aws) so that we can configure AWS IAM roles to impersonate GCP service accounts. Besides setting up the appropriate resources in Terraform, we will need to teach the GCP API clients in `facilitator` how to do this.

tgeoghegan:
Adds support for GCP workload identity pools to allow workloads in
Amazon EKS to impersonate GCP service accounts and access GCS buckets,
which ISRG's international deployment will use to share validation
batches to peers.

`GcpOauthTokenProvider` now has a third way to obtain a default account
access token, which involves sending a `GetCallerIdentity` token to
`sts.googleapis.com`. The three modes of obtaining default tokens -- GKE
metadata service, service account key file and the new one -- are
factored out into a new `DefaultTokenProvider` enum to make the flow
through `GcpOauthTokenProvider` more clear.

`aws_credentials.rs` now has a function `get_caller_identity_token()`,
which constructs a `sts:GetCallerIdentity` request signed under the AWS
v4 request signing scheme. `rusoto_core::SignedRequest` is sadly
unsuitable for this because it cannot construct a request in the format
expected by `sts.googleapis.com`'s `token()` method (see
https://issuetracker.google.com/issues/190809963 for details). This
commit includes an implementation of request signing vendored from
`[rust-s3](https://crates.io/crates/rust-s3)`.

It's unfortunate to duplicate the implementation but (1) pulling the
`rust-s3` crate adds over 70 crates to our dependency tree and (2) the
AWS v4 signing scheme is unlikely to change, so I think this is the best
choice.

Resolves #681


facilitator/src/gcp_oauth.rs
```diff
@@ -84,17 +91,231 @@ struct ServiceAccountKeyFile {
...
+    AwsIamFederationViaWorkloadIdentityPool(WorkloadIdentityPoolParameters<P>),
+}
+
+impl<P: ProvideAwsCredentials> fmt::Debug for DefaultTokenProvider<P> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut debug_struct = f.debug_struct("DefaultTokenProvider");
+        match self {
+            Self::GkeMetadataService => debug_struct.field(
+                "token_source",
+                &"default service account from GKE metadata service".to_owned(),
```
jsha, line 118: This is an interesting construction. It looks like this param wants `&dyn Debug`. But `str` has a Debug impl: https://doc.rust-lang.org/nightly/std/primitive.str.html#impl-Debug. Any idea why a plain string literal wouldn't work here? Not a requirement for merging, just kibbitzing really.

facilitator/src/gcp_oauth.rs
```diff
@@ -84,17 +91,231 @@ struct ServiceAccountKeyFile {
...
+    /// Engine.
+    GkeMetadataService,
+    /// Use a GCP service account key file to authenticate to GCP IAM as some
+    /// service account.
+    ServiceAccountKeyFile(ServiceAccountKeyFile),
+    /// Use a GCP Workload Identity Pool to federate GCP IAM with AWS IAM,
+    /// allowing workloads in AWS Elastic Kubernetes Service to impersonate a
+    /// GCP SA.
+    AwsIamFederationViaWorkloadIdentityPool(WorkloadIdentityPoolParameters<P>),
+}
```
jsha, line 110: In another project I had an enum like this, but I would up being much happier making it a trait. Have you considered making this a trait instead? So then instead of `default_token` providing a match-based dispatch to `account_token_from_gke_metadata_service`, `account_token_with_key_file`, etc, you'd have a trait with one required method, `default_token`, that would be implemented by each of `GkeMetadataService` (a new unit type), `ServiceAccountKeyFile`, and `WorkloadIdentityPoolParameters<P>`.

Also I note that the dispatched-to methods don't actually seem to use `&self`; the trait approach would make more meaningful use of `&self`.

facilitator/src/gcp_oauth.rs
```diff
@@ -84,17 +91,231 @@ struct ServiceAccountKeyFile {
...
+        let request = agent.prepare_request(RequestParameters {
+            url: Url::parse(&key_file.token_uri).context(format!(
+                "failed to parse key_file.token_uri: {}",
+                &key_file.token_uri
+            ))?,
+            method: Method::Post,
+            ..Default::default()
+        })?;
+
+        let request = request.set("Content-Type", "application/x-www-form-urlencoded");
```
jsha, line 229: You might be interested in https://docs.rs/ureq/2.1.1/ureq/struct.Request.html#method.send_form (note: you have to click open the `impl Request` because of some horrible rustdoc bug that I probably introduced)

jsha (commented):
Generally looks good; 1 suggested refactoring, which you can take or leave.

facilitator/src/gcp_oauth.rs
```diff
@@ -50,8 +55,10 @@ impl OauthToken {
 }
 
 /// Represents the response from a GET request to the GKE metadata service's
-/// service account token endpoint, or to oauth2.googleapis.com.token.
-/// https://developers.google.com/identity/protocols/oauth2/service-account
+/// service account token endpoint, or to oauth2.googleapis.com.token[1], or to
+/// sts.googleapis.com's token() endpoint[2]
```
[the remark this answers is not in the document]
tgeoghegan: GCP's documentation calls it [Method: token](https://cloud.google.com/iam/docs/reference/sts/rest/v1/TopLevel/token) and I guess my brain decided methods have parameter lists, hence ().

facilitator/src/gcp_oauth.rs
tgeoghegan: If I do

facilitator/src/gcp_oauth.rs
tgeoghegan: I think I did this because I was replicating what I did in `aws_credentials::Provider`, where I provided an explanation for this pattern: https://github.com/abetterinternet/prio-server/blob/c0813359f01937d8861bdff70a6c3caea2ad379d/facilitator/src/aws_credentials.rs#L39

Those constraints don't really apply in this case. I'll give it a try with a trait object member on `GcpOauthTokenProvider` and see if I like it better.

facilitator/src/gcp_oauth.rs
tgeoghegan: Good tip, this would save me the trouble of constructing the form request body myself.

facilitator/src/gcp_oauth.rs
```diff
@@ -114,17 +335,11 @@ pub(crate) struct GcpOauthTokenProvider {
     logger: Logger,
 }
 
-impl fmt::Debug for GcpOauthTokenProvider {
+impl<P: ProvideAwsCredentials> fmt::Debug for GcpOauthTokenProvider<P> {
```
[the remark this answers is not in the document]
tgeoghegan: The compiler won't have it:
```
error[E0277]: the trait bound `T: ProvideAwsCredentials` is not satisfied
   --> src/gcp_oauth.rs:338:24
    |
313 | pub(crate) struct GcpOauthTokenProvider<P: ProvideAwsCredentials> {
    |                                            --------------------- required by this bound in `GcpOauthTokenProvider`
...
338 | impl<T> fmt::Debug for GcpOauthTokenProvider<T> {
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^ the trait `ProvideAwsCredentials` is not implemented for `T`
    |
help: consider restricting type parameter `T`
    |
338 | impl<T: ProvideAwsCredentials> fmt::Debug for GcpOauthTokenProvider<T> {
    |       ^^^^^^^^^^^^^^^^^^^^^^^
```

facilitator/src/gcp_oauth.rs
jsha: This is probably not the strict way to think about it, but maybe:

`&str` is already a fat pointer. `&dyn Debug` needs to be a fat pointer. There's not enough "fat" left in `&str` to stick the vtable in, so it needs another `&` where it can stick in some fat. :-D

facilitator/src/gcp_oauth.rs
tgeoghegan: Ultimately I like it better your way. Besides the advantages you describe, I also used `Box<dyn ProvideAwsCredentials>`, which means that several structs no longer need to be generic over `P: ProvideAwsCredentials`, at the cost of some dynamic dispatch. I think that's better than introducing AWS-related type parameters to numerous structs that have nothing (obvious) to do with AWS, though.

facilitator/src/gcp_oauth.rs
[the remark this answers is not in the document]
tgeoghegan: However this is moot, because as of the latest commit, `GcpOauthTokenProvider` is no longer generic over P.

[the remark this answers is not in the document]
jsha: Ohh I suspect this is is because ProvideAwsCredentials is not bounded by Debug. And the `#[derive(Debug)]` logic probably depends on basically "T can be Debug if all fields of T are Debug."

jsha (approved):
The switch to a trait for DefaultTokenProvider looks great!

tgeoghegan:
Squashed and rebased to:
- resolve conflict with `facilitator/Cargo.toml` on main
- remove unneeded `std::fmt::Debug` impl on `GcsTransport`
- improve an example in `docs/authentication-and-iam.md`

jsha (approved):

PR #737, opened 2021-06-16
facilitator: remove OpenSSL dependency

### LINKED ISSUE #451: Remove OpenSSL dependency in facilitator
The `facilitator` Dockerfile currently has to play some games to install OpenSSL and statically link it. If I remember right, this was necessary to enable Prometheus push support. We no longer push metrics from `facilitator` (we instead have a more conventional scrape endpoint, and it doesn't use TLS because it is only exposed within the cluster), and have disabled the `push` feature on the `prometheus` create, so we should check whether it still needs OpenSSL at all. If not, we can simplify our Dockerfile and reduce image size a good bit. If we still do need OpenSSL, we should look into upstreaming a change to `prometheus` so it can use `rustls` instead of OpenSSL.

tgeoghegan:
We now configure `kube` to use `rustls`, removing our dependency on any
native TLS implementation which ends up pulling in OpenSSL. This means
we no longer need to install `openssl-dev` when building the
`facilitator` Docker image, nor do we need to play special games to
statically link it on Alpine.

We still end up depending on
[`openssl-probe`](https://crates.io/crates/openssl-probe), but that
crate doesn't actually link OpenSSL and so is harmless.

This commit also separates the `prio-facilitator` Dockerfile so that we
copy just the compiled binary from the builder container into the image
we run, which cuts down image size by ~500 MB.

Closes #451

### DIFF
--- a/facilitator/Cargo.lock
+++ b/facilitator/Cargo.lock
@@ -690,21 +690,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
-name = "foreign-types"
-version = "0.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f6f339eb8adc052cd2ca78910fda869aefa38d22d5cb648e6485e4d3fc06f3b1"
-dependencies = [
- "foreign-types-shared",
-]
-
-[[package]]
-name = "foreign-types-shared"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
-
-[[package]]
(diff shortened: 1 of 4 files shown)


tgeoghegan:
With these changes, a `prio-facilitator` image tarball (`podman save --format oci-archive letsencrypt/prio-facilitator`) is ~240 MB, down from 791 MB. The vast majority of that is from not including intermediate build products (`facilitator/target`) in the final image.

I still need to test this in both GKE and EKS to make sure that the changes to the `prio-facilitator` image content don't somehow break DNS or something.

aaomidi:
This is such an exciting change!!

aaomidi (approved):

jsha:
Woo this is great! Incidentally should we maybe switch to release builds of facilitator now?

tgeoghegan:
I gave this a whirl in my dev cluster and unfortunately it doesn't work: https://github.com/abetterinternet/prio-server/issues/451#issuecomment-865365589

Hopefully someday upstream will integrate the changes we need to do this. In the meantime, I'm going to close this, since I highly doubt it'll merge cleanly by the time that happens.

PR #762, opened 2021-06-24
 `deploy-tool`: don't generate packet encryption key twice #756 

### LINKED ISSUE #756: `deploy-tool`: don't generate packet encryption key twice
#572 introduced a regression into `deploy-tool`: when generating keys and manifests for a newly deployed locality, `deploy-tool` generates the packet encryption key (which is shared across both ingestors in a locality, as opposed to the batch signing key which is unique for each (ingestor, locality) pair) twice, and updates the Kubernetes secret twice. As a consequence, whichever manifest gets generated first will contain a packet encryption key CSR that does not correspond to the private key used by either ingestor's `intake-batch` or `aggregate` worker deployment. 

The MITRE component that fetches packet encryption key CSRs happens to consult the locality's `g-enpa` manifest, meaning this bug is benign if that manifest is generated last (which is what happened in `us-dc`), but if the `apple` one gets generated last, then clients will be configured with the wrong key (which is what happened in `us-la`). Go iterates over maps in a random order, which is why this bug can sometimes be benign.

tgeoghegan:
In an effort to make this a bit easier to review, commit 171bbdd contains just the bugfix, and then 3d283eb has some refactoring to make this testable.

Resolves #756

### DIFF
--- a/deploy-tool/main.go
+++ b/deploy-tool/main.go
@@ -132,17 +132,59 @@ func generateAndDeployKeyPair(
 	return p256Key, nil
 }
 
-func manifestExists(fqdn, dsp string) bool {
-	// Remove the locality name from the FQDN
-	path := fmt.Sprintf("https://%s/%s-manifest.json", fqdn, dsp)
+// globalManifestExists checks if a global data share processor manifest exists
+// relative to the manifest base URL provided, returning (true, nil) if so,
+// (false, nil) if not, and (false, error) if it could be determined
+func globalManifestExists(manifestBaseURL string) (bool, error) {
+	url := fmt.Sprintf("https://%s/global-manifest.json", manifestBaseURL)
 
-	resp, err := http.Get(path)
+	resp, err := http.Get(url)
 	if err != nil {
-		log.Fatalf("error when getting manifest %s: %s", path, err)
+		return false, fmt.Errorf("error fetching manifest %s: %w", url, err)
 	}
 	defer resp.Body.Close()
 
-	return resp.StatusCode == 200
+	if resp.StatusCode == 403 || resp.StatusCode == 404 {
+		return false, nil
+	}
+
+	if resp.StatusCode != 200 {
+		return false, fmt.Errorf("unexpected HTTP status %d", resp.StatusCode)
+	}
+
+	return true, nil
+}
+
+// fetchSpecificManifest fetches the manifest for the specified data share
+// processor relative to the provided manifest base URL and returns it, if it
+// exists and is well-formed. Returns (nil, nil) if the manifest does not exist.
+// Returns nil and some error if something went wrong while trying to fetch or
+// parse the manifest.
+func fetchSpecificManifest(manifestBaseURL, dataShareProcessorName string) (*manifest.DataShareProcessorSpecificManifest, error) {
+	path := fmt.Sprintf("https://%s/%s-manifest.json", manifestBaseURL, dataShareProcessorName)
+
+	resp, err := http.Get(path)
+	if err != nil {
+		return nil, fmt.Errorf("error when getting manifest %s: %s", path, err)
+
+	}
+	defer resp.Body.Close()
+
+	// GETs on non-existent objects in storage.googleapis.com return HTTP 403
+	if resp.StatusCode == 403 || resp.StatusCode == 404 {
+		return nil, nil
+	}
+
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("unexpected HTTP status fetching manifest: %d", resp.StatusCode)
+	}
+
+	var manifest manifest.DataShareProcessorSpecificManifest
+	if err := json.NewDecoder(resp.Body).Decode(&manifest); err != nil {
+		return nil, fmt.Errorf("error parsing manifest: %w", err)
+	}
+
+	return &manifest, nil
 }
 
 func setupTestEnvironment(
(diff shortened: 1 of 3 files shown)


deploy-tool/main.go
```diff
@@ -132,26 +132,85 @@ func generateAndDeployKeyPair(
 	return p256Key, nil
 }
 
-func manifestExists(fqdn, dsp string) bool {
-	// Remove the locality name from the FQDN
-	path := fmt.Sprintf("https://%s/%s-manifest.json", fqdn, dsp)
+// globalManifestExists checks if a global data share processor manifest exists
+// relative to the manifest base URL provided, returning (true, nil) if so,
+// (false, nil) if not, and (false, error) if it could be determined
+func globalManifestExists(manifestBaseURL string) (bool, error) {
```
jcjones, line 138: Optimally this would be a thin wrapper for a method `fetchManifest` that could be called here and down in Fetcher so the status code and Sprintf-work could be consolidated by returning the response object in some way or other, but basically all this is a bit miserable so I think I agree with duplicating the code for this poor tool.

deploy-tool/main.go
```diff
@@ -235,27 +289,22 @@ func createManifest(
...
-	certificatesByNamespace := manifest.PacketEncryptionKeyCSRs{}
-	newCertificates := manifest.PacketEncryptionKeyCSRs{}
-	for name, packetEncryptionCertificate := range manifestWrapper.SpecificManifest.PacketEncryptionKeyCSRs {
-		if packetEncryptionCertificate.CertificateSigningRequest != "" {
-			newCertificates[name] = packetEncryptionCertificate
-			continue
+	for name, packetEncryptionCertificateSigningRequest := range manifestWrapper.SpecificManifest.PacketEncryptionKeyCSRs {
+		if packetEncryptionCertificateSigningRequest.CertificateSigningRequest != "" {
+			// We never create CSRs in Terraform, so this should never happen
+			return fmt.Errorf("unexpected packet encryption key CSR in Terraform output")
```
jcjones, line 298: might be worthwhile to indicate `name` in this Errorf since it is already formatted.
```suggestion
			return fmt.Errorf("unexpected packet encryption key CSR in Terraform output for name %s", name)
```

deploy-tool/main.go
```diff
@@ -207,25 +266,20 @@ func createBatchSigningPublicKey(
...
-	k8sSecretsClient := k8sClient.CoreV1().Secrets(manifestWrapper.KubernetesNamespace)
+	k8sSecretsClient := k8sSecretsClientGetter.Secrets(manifestWrapper.KubernetesNamespace)
 
-	newBatchSigningPublicKeys := manifest.BatchSigningPublicKeys{}
 	for name, batchSigningPublicKey := range manifestWrapper.SpecificManifest.BatchSigningPublicKeys {
 		if batchSigningPublicKey.PublicKey != "" {
-			newBatchSigningPublicKeys[name] = batchSigningPublicKey
-			continue
+			// We never create keys in Terraform, so this should never happen
+			return fmt.Errorf("unexpected batch signing key in Terraform output")
```
jcjones, line 280: might be worthwhile to indicate `name` in this Errorf since it is already formatted.
```suggestion
			return fmt.Errorf("unexpected batch signing key in Terraform output for name %s", name)
```

deploy-tool/main.go
```diff
@@ -338,6 +387,61 @@ func backupKeys(
...
+			return err
+		}
+
+		if manifest == nil {
+			continue
+		}
+
+		existingManifests[dataShareProcessorName] = struct{}{}
+
+		for keyName, certificateSigningRequest := range manifest.PacketEncryptionKeyCSRs {
```
jcjones, line 415: I understand the preconditions expected here, but ISTM we should verify that if `packetEncryptionKeyCSRs[keyName]` is not empty here, that it has the same CSR rather than clobbering it indiscriminately. 

This obviously can't happen when there are only two processors in the first place, but it'd be a painful bug if there were >2. Maybe it could just confirm that the key is empty for right now, and when this gets modified to handle additional processors as it might well be, it'll fail at that point and we can address it properly then.

jcjones (dismissed):

deploy-tool/main.go
tgeoghegan: I agree. Where I got hung up here is that in one case, all I care about is whether it exists, and in the other, I need the actual manifest. In any case this gets refactored away in #763 which is also why I was OK with letting this be ugly for that brief time between when this PR is merged and when that one lands.

deploy-tool/main.go
```diff
@@ -132,26 +132,85 @@ func generateAndDeployKeyPair(
...
+func globalManifestExists(manifestBaseURL string) (bool, error) {
+	url := fmt.Sprintf("https://%s/global-manifest.json", manifestBaseURL)
+
+	resp, err := http.Get(url)
+	if err != nil {
+		return false, fmt.Errorf("error fetching manifest %s: %w", url, err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == 403 || resp.StatusCode == 404 {
```
aaomidi, line 147: nit: We could probably abstract away the http call and error handling into its own function. I believe we're doing these generic if checks in multiple places now.

aaomidi (approved):

deploy-tool/main.go
tgeoghegan: J.C. made a similar observation, and my answer is that all of this gets refactored away in #763, which is context I should have provided in this PR.

PR #763, opened 2021-06-25
deploy-tool: get rid of own_manifest_base_url

### LINKED ISSUE #751: terraform/deploy-tool: get rid of `own_manifest_base_url`
For ENPA international, @hostirosti and I are both planning to skip using custom subdomains to serve manifests, and instead just serve them from `storage.googleapis.com/<bucket name>` or `<bucket name>.s3.amazonaws.com`, on the premise that the name of a cloud storage bucket is about as trustworthy as a DNS name in an X.509 cert. Currently, `terraform/main.tf` emits an output `own_manifest_base_url` which is used by `deploy-tool` to check if a manifest already exists. It should be able to do this with the `manifest_bucket` input, so we can get rid of the `own_manifest_base_url` parameter.

tgeoghegan:
Terraform output included both `own_manifest_base_url` and
`manifest_bucket`, both of which referred to the same storage. The
former was used to check for the existence of manifests via HTTP GET;
the latter to upload new manifests to a cloud storage bucket. This
commit eliminates `own_manifest_base_url`. To that end,
`manifest/writer.go` is renamed to `manifest/storage.go`, and we
introduce interfaces that capture both uploading and fetching manifests.

Closes #751


jcjones:
This is pretty substantial set of changes compared to the prior two, and some of what was reviewed before is replaced here. Anyway, I generally like this refactor.

manifest-updater/manifest/storage.go
```diff
@@ -0,0 +1,260 @@
...
+	AWSRegion string `json:"aws_region,omitempty"`
+	// AWSProfile is the AWS CLI config profile that should be used to
+	// authenticate to AWS, if the bucket is an S3 bucket
+	AWSProfile string `json:"aws_profile,omitempty"`
+}
+
+// NewStorage creates an instance of the appropriate implementation of Writer for
+// the provided bucket
+func NewStorage(bucket *Bucket) (Storage, error) {
+	if bucket.AWSRegion != "" {
```
jcjones, line 63: I'd recommend you make a specific Bucket `isAWS` and `isGCS` methods rather than this implicit "does AWS Region have data?" - you should let the helper method check AWSProfile as well to avoid mistakes in the future.

manifest-updater/manifest/storage.go
```diff
@@ -0,0 +1,260 @@
...
+
+func (s *GCSStorage) writeManifest(manifest interface{}, path string) error {
+	log.Info().
+		Str("path", path).
+		Msg("writing a manifest file")
+
+	ioWriter := s.getWriter(path)
+
+	if err := json.NewEncoder(ioWriter).Encode(manifest); err != nil {
+		_ = ioWriter.Close()
```
jcjones, line 100: consider not throwing away this error and instead always appending it beyond the first `err` even if it's `nil`. I don't know about this particular writer, but some do give different and yet still useful errors on Close for network failures that caused the earlier Write to fail, particularly if over UDP. But again, feel free to ignore if I'm offbase here.

manifest-updater/manifest/storage.go
```diff
@@ -0,0 +1,260 @@
...
+
+	var manifest IngestorGlobalManifest
+	if err := json.NewDecoder(reader).Decode(&manifest); err != nil {
+		return false, fmt.Errorf("error parsing manifest: %w", err)
+	}
+
+	return true, nil
+}
+
+func (s *S3Storage) getReader(path string) (io.ReadCloser, error) {
```
jcjones, line 248: Add a comment to this method that the returned Body, being an io.ReadCloser, needs to be closed to avoid leaking.

Also, we then need to Close it everywhere this gets called.  json.Decoder doesn't, as it only needs a Reader, so it definitely doesn't know to Close on completion.

This probably should have been caught via a closer lint; we might need to check our CI linting.

jcjones (changes requested):
A few minor things. This is much cleaner than it started with, and I don't know that the earlier stacks really need revision if this is where we're going.

manifest-updater/manifest/storage.go
tgeoghegan: Good catch, I'm spoiled by Rust's automagic closing of readers on drop. The same applies to the `cloud.google.com/go/storage.Reader`s returned by `GCSStorage.getReader` which implements `io.ReadCloser`, too.

manifest-updater/manifest/storage.go
tgeoghegan: No, I think you raise a good point. We have to be more careful than usual with `cloud.google.com/go/storage.Writer` because it doesn't actually write to the network until `Writer.Close()` is called, so if you ignore the error from `Close()` you might not learn about failed writes. In this case we've already failed but it's still good to be meticulous about error handling all the time.

manifest-updater/manifest/storage.go
tgeoghegan: I'll do you one better: I changed the definition of `Bucket` and the Terraform output so that instead of bucket name, we get a URL-like string in the form `gs://bucket-name` or `s3://bucket-name` so that `NewStorage` can more reliably determine which cloud storage service we're using. All other things being equal, I might have added an explicit field to `Bucket` to indicate which cloud platform we're using, but we use this bucket URL convention in several other places throughout `prio-server` so I'm going to err on the side of consistency.

jcjones (approved):
Looks good now!

tgeoghegan:
Thank you for the review. It turns out we'd never run CI on this PR because it wasn't targeting main, and on top of that some of the TF changes will collide with #802, so I'm going to rebase this on top of that change.

PR #795, opened 2021-07-09
facilitator: beef up config::Identity type

tgeoghegan:
This commit enhances the existing `config::Identity` type to make it
into a struct wrapping an `Option<String>`. The goal is to abstract away
the special handling of the value `Some("")` as being equivalent to
`None` and make it easier to use `Identity` throughout the crate instead
of passing AWS role ARNs and GCP service account emails as
`Option<&str>` or `Option<String>` in various places. This will make
things more clear when a subsequent commit introduces support for AWS
authentication flows that first require authenticating to GCP, and so
`aws_credentials::Provider` will have to manage more than one identity.

We provide `std::str::FromStr` and `serde::Deserialize` implementations
on `config::Identity` to allow for graceful instantiation of `Identity`
from program arguments or JSON manifests. Along the way, we make some
improvements to the defintions in `manifest.rs` to use `Identity` and
`StoragePath` instead of `Option<String>` and `String`.

### DIFF
--- a/facilitator/src/aws_credentials.rs
+++ b/facilitator/src/aws_credentials.rs
@@ -101,7 +101,7 @@ impl Provider {
         purpose: &str,
         logger: &Logger,
     ) -> Result<Self> {
-        match (use_default_provider, identity) {
+        match (use_default_provider, identity.as_str()) {
             (true, _) => Self::new_default(),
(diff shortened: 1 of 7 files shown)


jcjones (approved):
I see no issues here. LGTM.

PR #796, opened 2021-07-09
facilitator: fix GCP workload identity pool args

tgeoghegan:
PR #728 added support for GCP Workload Identity Pool to federate
identity from AWS to GCP. It added a
`gcp-workload-identity-pool-provider` command line argument, but failed
to wire it up to the various `facilitator` subcommands that need it.
Besides that, we need allow Terraform to pass
`GCP_WORKLOAD_IDENTITY_POOL_PROVIDER = ""` when running `facilitator`
(it's easier to handle this in Rust than to conditionally add values to
Terraform maps), so we add a `WorkloadIdentityPoolParameters::new`
function that looks for the empty string and treats it as equivalent to
the argument not having been specified.

### DIFF
--- a/facilitator/src/bin/facilitator.rs
+++ b/facilitator/src/bin/facilitator.rs
@@ -725,6 +725,7 @@ fn main() -> Result<(), anyhow::Error> {
                 .add_storage_arguments(Entity::Own, InOut::Output)
                 .add_use_bogus_packet_file_digest_argument()
                 .add_permit_malformed_batch_argument()
+                .add_gcp_workload_identity_pool_provider_argument()
         )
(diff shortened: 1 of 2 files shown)


jcjones (approved):
Generally speaking, I'm all onboard with replacing all these String types with concrete ones, for type safety if nothing else.

tgeoghegan:
> Generally speaking, I'm all onboard with replacing all these String types with concrete ones, for type safety if nothing else.

In the future, we'll parse JSON or YAML config documents into Rust structs with `serde`, and everything will be beautiful and nothing will hurt.

Issue #747, opened 2021-06-21
Enable data share processor to be purely GCP


tgeoghegan:
@hostirosti and I were talking and agreed it would be easier if the data share processor/aggregator that runs on GCP doesn't have to manage any AWS resources at all. This means:

- GCP DSP won't create AWS IAM roles to use when writing to AWS DSP's peer validation buckets
- GCP DSP won't advertise an AWS account ID from its global manifest
- AWS DSP will create an AWS IAM role that the GCP DSP's facilitator intake workers should assume to write peer validations (probably via the GCP SA used to write sum parts)
- DSP global manifest will bump to v2, which removes the AWS account ID (on the GCP DSP side) and adds an AWS IAM role (on the AWS DSP side)

The above list is my current guess as to what's needed. I need to review the design doc to make sure I haven't missed anything.

tgeoghegan:
The nuts and bolts of how this will work for a GKE workload writing to S3 buckets are that the EKS-resident DSP will create an IAM role, configured to allow the service account in the GKE-resident DSP's global manifest to assume it. So an intake-worker will need to obtain an Oauth token for global manifest's GCP SA using [`generateIdToken`](https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateIdToken) and then assume a role using WebIdentity. `facilitator` already knows how to get an identity token for the _default_ GCP SA from the GKE metadata service, and it already knows how to authenticate to `generateAccessToken` to allow GCP SA impersonation, so we just need to wire up an extra path or two through `gcp_oauth.rs` and `aws_credentials.rs`.

tgeoghegan:
This is resolved by #802, #799, #798

PR #802, opened 2021-07-13
Terraform support for international deployment

tgeoghegan:
This series of commits adds support for the international deployment. There are two important thrusts to this:

## Support deploying onto AWS Elastic Kubernetes Service (resolving #655)
We add a number of new modules and augment existing ones so that we can manage task queues, storage, Kubernetes cluster and more IAM resources than you can shake a policy document at.

## Enable GCP data share processor to avoid managing AWS resources
See #747 for motivation and some details. The upshot here is that instead of the GCP side creating AWS IAM roles that it will assume when writing to the peer's S3 buckets, we now have the AWS side create those roles and advertise them from its specific manifests. The GCP side must now advertise its sum part writer SA numeric ID from its global manifest so that the AWS side can permit that SA to assume the per-locality roles.

The total change is quite large, so I've tried to carve it up into smaller commits. I don't expect each intermediate commit to be deployable but I hope they make it easier to reason about the changes here.

What remains unfinished here is logging and metrics. I omitted that from this PR because these changes were enough for me to bring up a working environment and I wanted to get code review before moving forward.

The metrics story shouldn't be too hard to sort out: we need to flesh out `terraform/monitoring/monitoring.tf` so that it creates an EBS store (or whatever AWS storage primitive is appropriate) for Prometheus persistent metrics storage and we need to work out how to plumb SNS and SQS metrics into Prometheus so we can alert on them. 

Logging will be trickier: I think the move is to deploy a `fluentbit` DaemonSet to the cluster that is configured to forward pod logs into Stackdriver/Google Cloud Logging so we can store logs on the Google side and use GCP's excellent log searching console to monitor the EKS clusers.


tgeoghegan:
Since @hostirosti and I tested these changes in the international staging env we set up last week, and since this set of changes has very little risk to the US production deploy, I'm going to go ahead and merge it in order to unblock other ENPA international work.

PR #838, opened 2021-07-22
facilitator: bug fixes for international deployment

### LINKED ISSUE #833: `facilitator`: don't require `task-queue-identity` argument
`facilitator` may be passed `task-queue-identity` to specify a service account or IAM role to be used to access a PubSub subscription or SQS queue, respectively. It should be possible to not specify this parameter in order to use the default identity, but [we are parsing arguments incorrectly](https://github.com/abetterinternet/prio-server/blob/3719efd88740be55b399eb8a784c8a7630f1edfb/facilitator/src/bin/facilitator.rs#L2026).

### LINKED ISSUE #834: `facilitator`: skip crypto self check if manifests don't exist yet
In a previous PR, we introduced a crypto self-check at `facilitator` startup, which verifies public keys in manifests against private keys in Kubernetes secrets and noisily dies if they don't match up. I failed to account for the deploy-time case, where neither manifests nor secrets exist yet. We should teach `facilitator` to gracefully skip the crypto self check in that situation, because `terraform apply` can't succeed unless the `facilitator intake-batch-worker` and `aggregate-worker` deployments become healthy, which they can't if they crypto self-check fails!

tgeoghegan:
- Fix handling of `task-queue-identity` and a couple other `-identity` parameters
- The crypto self check will now gracefully be skipped if the instance's specific manifest does not exist, on the assumption that keys have not yet been generated and hence cannot be verified.
- Consult peer specific manifest to get identity to assume when writing to peer validation bucket as well as `peer-identity`

Resolves #833, #834

### DIFF
--- a/facilitator/src/bin/facilitator.rs
+++ b/facilitator/src/bin/facilitator.rs
@@ -1300,15 +1300,26 @@ where
     // We need the bucket to which we will write validations for the
     // peer data share processor, which can either be fetched from the
     // peer manifest or provided directly via command line argument.
+    let mut peer_validation_identity = value_t!(
+        sub_matches.value_of(Entity::Peer.suffix("-identity")),
+        Identity
+    )?;
     let peer_validation_bucket =
         if let Some(base_url) = sub_matches.value_of("peer-manifest-base-url") {
-            DataShareProcessorSpecificManifest::from_https(
+            let peer_manifest = DataShareProcessorSpecificManifest::from_https(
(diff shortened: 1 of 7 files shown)


hostirosti:
LGTM

tgeoghegan:
These changes have been tested in the international staging environment that Robert and I set up last week. in the interest of unblocking international ENPA work, I'm going to merge this.

PR #869, opened 2021-08-02
facilitator: fix `RUST_LOG` handling

tgeoghegan:
We configure log level via the `RUST_LOG` environment variable. We were
using the `option_env!` macro to get its value, but that macro gets
environment variables at _compile_ time, not _run_ time, so it was
effectively impossible to set the log level to anything but `INFO`,
except by recompiling `facilitator`. We now use `std::env::var`, and
also apply `str::to_uppercase` to the value so that either of `debug` or
`DEBUG` will enable debug level logs.


aaomidi (approved):

jcjones (approved):
LGTM

PR #889, opened 2021-08-09
terraform: avoid TF type coercions

tgeoghegan:
Terraform insists that both sides of a conditional have the same type,
which was causing TF to coerce the numeric `aws-account-id` value into a
string, leading to malformed global manifests. We lower the `jsonencode`
function call into either side of the ternary operator to avoid this
coercion.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -399,22 +399,20 @@ locals {
   # Does this specific environment home the ingestor?
   is_env_with_ingestor = local.deployment_has_ingestor && lookup(var.test_peer_environment, "env_with_ingestor", "") == var.environment ? true : false
 
-  global_manifest = jsonencode(
-    # If pure GCP, we use the newer global manifest format
-    var.pure_gcp ? {
-      format = 1
-      server-identity = {
-        gcp-service-account-id    = var.use_aws ? null : tostring(google_service_account.sum_part_bucket_writer.unique_id)
-        gcp-service-account-email = google_service_account.sum_part_bucket_writer.email
-      }
-      } : {
-      format = 0
-      server-identity = {
-        aws-account-id            = tonumber(data.aws_caller_identity.current.account_id)
-        gcp-service-account-email = google_service_account.sum_part_bucket_writer.email
-      }
+  # If pure GCP, we use the newer global manifest format
+  global_manifest = var.pure_gcp ? jsonencode({
+    format = 1
+    server-identity = {
+      gcp-service-account-id    = var.use_aws ? null : tostring(google_service_account.sum_part_bucket_writer.unique_id)
+      gcp-service-account-email = google_service_account.sum_part_bucket_writer.email
     }
-  )
+    }) : jsonencode({
+    format = 0
+    server-identity = {
+      aws-account-id            = tonumber(data.aws_caller_identity.current.account_id)
+      gcp-service-account-email = google_service_account.sum_part_bucket_writer.email
+    }
+  })
 
   manifest = var.use_aws ? {
     bucket      = module.manifest_aws[0].bucket


tgeoghegan:
Transcribed from Slack, way too many words about the TF/HCL behavior this change avoids:

Here's a weird Terraform behavior that illustrates why type coercions are terrible. [`jsonencode`](https://www.terraform.io/docs/language/functions/jsonencode.html) takes values, such as Hashicorp Configuration Language (HCL) maps, and returns a string of their JSON representation. Very useful for constructing config files from TF.

```
> jsonencode(10)
"10"
> jsonencode("string")
"\"string\""
> jsonencode([10, "string"])
"[10,\"string\"]"
> jsonencode({key = "value", key2 = 10})
"{\"key\":\"value\",\"key2\":10}"
> 
```

OK, so far so good: we can pass it a variety of HCL objects and get a JSON encoding back, even with `"` escaped. Let's say we're being fancy and want to emit a different object based on some condition. We break out the tried and true ternary operator (omitting this is Rust's big error, the ternary operator rules):

```
> jsonencode(true ? { string_key = "string" } : { number_key = 10 })
"{\"string_key\":\"string\"}"
```

Checks out. What does the other side of `:` look like?

```
> jsonencode(false ? { string_key = "string" } : { number_key = 10 })
"{\"number_key\":\"10\"}"
>
 ```

Wait, who ordered those quotes around `10`? It's clearly a bare number in the HCL map, so it should be encoded in JSON as a number without quotes. Is this a quirk of `jsonencode`? Let's try to rule that out...

```
> true ? {string_key = "string" } : {number_key = 10}
tomap({
  "string_key" = "string"
})
> false ? {string_key = "string" } : {number_key = 10}
tomap({
  "number_key" = "10"
})
> 
```

What?

```
> false ? {string_key = "string" } : {number_key = 10, other_key = 11}
tomap({
  "number_key" = "10"
  "other_key" = "11"
})
>
```

_What?_ Why on earth are these numerical values getting coerced into strings?

```
> { string_key = "string", number_key = 9 }
{
  "number_key" = 9
  "string_key" = "string"
}
> true ? { string_key = "string", number_key = 9 } : { number_key = 9 }
tomap({
  "number_key" = "9"
  "string_key" = "string"
})
```

_what?!_ Why does being in the context of a ternary expression change how `{ string_key = "string", number_key = 9 }` is evaluated? At this point I have a hard time even articulating what rule or rules Terraform is applying. We get a clue if we try something even weirder:

```
> false ? {string_key = "string" } : {number_key = [10]}

>  
Error: Inconsistent conditional result types

  on <console-input> line 1:
  (source code not available)

The true and false result expressions must have consistent types. The given
expressions are object and object, respectively.

>
```

Ah, so HCL insists that the types of both results line up. And we might infer that it's not sufficient that both be of type `map` -- internally, Terraform must be tracking that it's something like `map<value_type=string>`. What's chilling (to me) is that Terraform fails noisily if we try to treat `map<value_type=string>` equivalently to `map<value_type=list>` -- but if we provide `map<value_type=number>` where `map<value_type=string>` is expected, it silently coerces the number values into strings.

I welcome HCL being picky about types (though its generic `map` representation appears to assume that all values have the same type), but the silent coercion made this harder to figure out than it should have been, and it's hard to imagine that anybody really wants this behavior. In my case, the JSON blob I'm emitting has a schema that stipulates explicitly what type values must have so the implicit conversion to string is potentially disastrous.

Happily there's a straightforward fix: just lower the `jsonencode` function call into each side of the conditional, and then both will be of type `string` and the values passed to both `jsonencode` calls do not need to be of the same type or have a similar schema:

```
> true ? jsonencode({string_key = "string"}) : jsonencode({number_key = 10})
"{\"string_key\":\"string\"}"
> false ? jsonencode({string_key = "string"}) : jsonencode({number_key = 10})
"{\"number_key\":10}"
>
```

The lesson is: never try, just fail as hard and as noisily and as early as you can.

ezekiel (approved):

PR #926, opened 2021-08-24
Alert based on minimum observed queue sizes over time:

ezekiel:
Applying min_over_time() to each important queue alert allows us to
smooth momentary spikes when the queue quickly returns to zero. We care
when the queue has no empty periods for an extended duration.

### DIFF
--- a/terraform/modules/monitoring/prometheus_alerting_rules.yml
+++ b/terraform/modules/monitoring/prometheus_alerting_rules.yml
@@ -41,7 +41,7 @@ groups:
         {{ $labels.kubernetes_name }} in environment ${environment} is failing
         more than 95% of the time in the last 8 hours"
   - alert: intake_task_queue_growth
-    expr: stackdriver_pubsub_subscription_pubsub_googleapis_com_subscription_num_undelivered_messages{subscription_id!~".*-dead-letter",subscription_id=~".*-intake"} > 0
+    expr: min_over_time(stackdriver_pubsub_subscription_pubsub_googleapis_com_subscription_num_undelivered_messages{subscription_id!~".*-dead-letter",subscription_id=~".*-intake"}[15m]) > 0
     for: 3h
     labels:
       severity: page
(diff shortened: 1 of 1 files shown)


ezekiel:
Initial rule modifications were based on metrics from the `prio-staging` environment. For the production queues, `15m` is insufficient.

ezekiel:
Based on metrics for current production queues, `30m` should be a pretty appropriate balance.

terraform/modules/monitoring/prometheus_alerting_rules.yml
```diff
@@ -61,7 +61,7 @@ groups:
       description: "PubSub subscription {{ $labels.subscription_id }} in
       environment ${environment} has had undelivered messages for 10 hours"
   - alert: dead_letter_queue
-    expr: stackdriver_pubsub_subscription_pubsub_googleapis_com_subscription_num_undelivered_messages{subscription_id=~".*-dead-letter"} > 0
+    expr: min_over_time(stackdriver_pubsub_subscription_pubsub_googleapis_com_subscription_num_undelivered_messages{subscription_id=~".*-dead-letter"}[5m]) > 0
```
tgeoghegan, line 64: Any reason not to smooth this one out over 30 m? Same goes for `aggregate_task_queue_growth_sqs`, below.

tgeoghegan (approved):
LGTM with one question.

terraform/modules/monitoring/prometheus_alerting_rules.yml
ezekiel: The interval affects time-to-alert and time-to-auto-resolve - so I kept these two low because their `for:` configuration was already lower than the others, which seemed intentional.

If a ( pretty slight ) extension of the timeline for dead letter queues is appropriate, then I can normalize the values for all the definitions.

tgeoghegan: The only reason I did the 5m thing was per some advice on writing alerts I found... somewhere, I can't find the source anymore. Anyway, I'll turn the question around on you: if you say that taking the min over 5m is sufficient to debounce these signals, then I'm good with it. If you think that increasing both the `min_over_time` and `for` values to 30m will be more reliable, then we should do that. I'm not really concerned about having to wait an extra 25m to be alerted.

PR #931, opened 2021-08-26
terraform: add plan-refresh and apply-refresh to Makefile

ezekiel:
These supply the -refresh-only option to their respective (plan|apply)
terraform commands. Use these when the infrastructure state matches the
terraform state, but terraform suggests a refresh to pickup minor
changes which may have been applied outside of terraform.


branlwyd:
(closing this PR as it is very stale at this point)

PR #989, opened 2021-09-24
Make `trace_id` mandatory in tasks.

### LINKED ISSUE #452: Remove Option<> around trace_id in facilitator
Currently #443 implements the trace logging with an optional UUID parameter. The reason this is optional is that the already-scheduled intake and aggregation jobs won't have a UUID parameter, and requiring them to have one would cause those intakes and aggregations to fail.

Once a day or two after merging and deploying #443 has passed then we can go and remove the optional modifiers on those parameters.

Code with the optional UUID parameters have a comment referring to this issue. Just search `https://github.com/abetterinternet/prio-server/issues/452` in the codebase

BranLwyd:
`IntakeBatchTask` & `AggregationTask` both had `trace_id` fields that
were of type `Option<Uuid>`; the core change of this commit is to change
the type of these fields to `Uuid`. This is safe since all scheduled
jobs now have the `trace_id` parameter populated, so we can safely
deserialize them with a non-optional `trace_id`.

The remainder of the changes are straightforward from these changes.
Most test cases were using either "trace-id" or "None" for the trace
IDs; I replce these with an arbitrary UUID value. There were a couple of
commands (`intake-batch`, `aggregate`) which were hardwired to use
"None" as their trace ID--I changed these commands to instead generate a
new V4 (random) UUID, matching behavior at other places a trace ID is
generated.

Fixes #452.


facilitator/src/intake.rs
```diff
@@ -284,6 +284,8 @@ mod tests {
     use assert_matches::assert_matches;
     use prio::{encrypt::PublicKey, server::ServerError, util::SerializeError};
 
+    const TRACE_ID: Uuid = Uuid::from_bytes([97; 16]);
```
tgeoghegan, line 287: nit: this could be defined in `test_utils.rs` instead of being duplicated across test modules.

tgeoghegan (approved):
LGTM! I appreciate the use of `&'a Uuid` instead of `&'a str` in various function signatures.

PR #991, opened 2021-09-27
facilitator: move `rusoto-mock` to dev-dependency

tgeoghegan:
With some targeted use of `#[cfg(test)]`, we can omit
`aws_credentials::Provider::Mock` from the binary target of
`facilitator`, since it is only actually used in test targets. This
reduces binary size in the `release` configuration by 58,600 bytes
(`facilitator` is 27233040 bytes before, 27174440 after).


tgeoghegan:
n.b. This is stacked on #984.

BranLwyd (approved):
`#[cfg(test)]` is very nice for creating public test-only functions that would otherwise be a public part of the namespace.

PR #997, opened 2021-09-29
facilitator: stop writing validation packets to our own storage.

BranLwyd:
Instead, we recompute them as needed during aggregation--this works
because our validation packet is a function of the ingestion packet,
which we already read during aggregation. We still write validation
packet to our peer's storage, since knowing your peer's validation
packets is a requirement to perform validation.

### DIFF
--- a/facilitator/src/batch.rs
+++ b/facilitator/src/batch.rs
@@ -3,7 +3,7 @@ use crate::{
     idl::{BatchSignature, Header, Packet},
     metrics::BatchReaderMetricsCollector,
     transport::{Transport, TransportWriter},
-    DigestWriter, SidecarWriter, DATE_FORMAT,
+    DigestWriter, DATE_FORMAT,
 };
 use anyhow::{anyhow, Context, Result};
 use avro_rs::{Reader, Schema, Writer};
@@ -226,17 +226,14 @@ impl<'a, H: Header, P: Packet> BatchReader<'a, H, P> {
         let mut packet_file_reader = self
             .transport
             .get(self.batch.packet_file_key(), self.trace_id)?;
-        // SidecarWriter takes a Vec of std::io::write so we wrap the Vec we
-        // want to read the file into in a Vec.
-        let entire_packet_file = vec![Vec::new()];
-        let digest_writer = DigestWriter::new();
-        let mut sidecar_writer = SidecarWriter::new(entire_packet_file, digest_writer);
+        let mut entire_packet_file = Vec::new();
+        let mut digest_writer = DigestWriter::new(&mut entire_packet_file);
 
-        std::io::copy(&mut packet_file_reader, &mut sidecar_writer)
+        std::io::copy(&mut packet_file_reader, &mut digest_writer)
             .context("failed to load packet file")?;
 
         // ... then verify the digest over it ...
-        let packet_file_digest = sidecar_writer.sidecar.finish();
+        let packet_file_digest = digest_writer.finish();
         if header.packet_file_digest().as_slice() != packet_file_digest.as_ref() {
             let message = format!(
                 "packet file digest in header {} does not match actual packet file digest {}",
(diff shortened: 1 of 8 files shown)


workflow-manager/main.go
```diff
@@ -451,19 +418,17 @@ func scheduleTasks(config scheduleTasksConfig) error {
...
 	// to get the list of batches we can aggregate.
-	// Go doesn't have sets, so we have to use a map[string]bool. We use the
-	// batch ID as the key to the set, because batchPath is not a valid map key
-	// type, and using a *batchPath wouldn't give us the lookup semantics we
-	// want.
-	ownValidationsSet := map[string]bool{}
-	for _, ownValidationBatch := range ownValidationBatches.Batches {
-		ownValidationsSet[ownValidationBatch.ID] = true
+	// Go doesn't have sets, so we have to use a map[string]struct{}, with the batch ID being used
+	// as the map key.
```
tgeoghegan, line 424: nit: The Go set idiom is a pet peeve of mine so I have a habit of calling it out, but I think it's familiar enough to Go programmers that we can omit the explanatory comment.

facilitator/src/bin/facilitator.rs
```diff
@@ -1343,22 +1343,6 @@ where
         batch_signing_key: batch_signing_key_from_arg(sub_matches)?,
     };
 
-    // We created the bucket to which we write copies of our validation
-    // shares, so it is simply provided by argument.
-    let mut own_validation_transport = SignableTransport {
```
tgeoghegan, line 1348: Besides deleting the handling of these arguments, we should also be able to remove them from the `intake-batch`, `aggregate`, `intake-batch-worker` and `aggregate-worker` `SubCommand`s defined in `main()`. Additionally, we should be able to remove some entries from the `intake_batch_config_map` and `aggregate_config_map` resources in `terraform/modules/kubernetes/kubernetes.tf`, as well as the `--own-validation-input` argument passed to `workflow-manager`.

facilitator/src/lib.rs
```diff
@@ -39,15 +45,57 @@ pub enum Error {
...
+    // ingestion packets, do we abort handling of the entire
+    // batch (as implemented currently) or should we record it
+    // as an invalid UUID and emit a validation batch for the
+    // other packets?
+    for server in servers.iter_mut() {
+        let validation_message = match server
+            .generate_verification_message(r_pit, &packet.encrypted_payload)
+        {
+            Ok(m) => m,
+            Err(ServerError::Encrypt(_)) => {
```
tgeoghegan, line 67: It's a bit of a shame that we just drop the error here, but this was a problem before your PR. Filed #1033.

facilitator/src/lib.rs
```diff
@@ -58,46 +106,17 @@ impl DigestWriter {
...
-    fn new(writers: Vec<T>, sidecar: W) -> SidecarWriter<T, W> {
-        SidecarWriter { writers, sidecar }
-    }
-}
-
-impl<T: Write, W: Write> Write for SidecarWriter<T, W> {
-    fn write(&mut self, buf: &[u8]) -> Result<usize, std::io::Error> {
-        for writer in &mut self.writers {
-            writer.write_all(buf)?;
+        let rslt = self.writer.write(buf);
```
tgeoghegan, line 111: nit
```suggestion
        let result = self.writer.write(buf);
```

facilitator/src/batch.rs
```diff
@@ -301,80 +291,53 @@ impl<'a, H: Header, P: Packet> BatchWriter<'a, H, P> {
...
+        transport_writer
             .complete_upload()
             .context("failed to complete batch header upload")?;
 
-        let header_signature = key
-            .sign(&SystemRandom::new(), &sidecar_writer.sidecar)
-            .context("failed to sign header file")?;
-        Ok(header_signature)
+        // TODO(brandon): per the advice of ring's doc comments, create a single SystemRandom &
+        // reuse it throughout the application rather than creating a new SystemRandom each time.
```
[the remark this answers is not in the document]
BranLwyd: Filed #1034.

facilitator/src/lib.rs
BranLwyd: Done.

PR #1006, opened 2021-09-29
deploy-tool: reallow creation of singleton-ingestor manifests.

BranLwyd:
This was broken at some point in the past, and likely wasn't caught
before now only because no one has tried to create a new environment.

### DIFF
--- a/manifest-updater/manifest/storage.go
+++ b/manifest-updater/manifest/storage.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"path"
 	"strings"
 
 	"cloud.google.com/go/storage"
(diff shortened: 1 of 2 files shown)


manifest-updater/manifest/storage.go
```diff
@@ -119,9 +123,9 @@ func (s *GCSStorage) writeManifest(manifest interface{}, path string) error {
 	return nil
 }
 
-func (s *GCSStorage) getWriter(path string) *storage.Writer {
-	ioWriter := s.client.Bucket(s.manifestBucketLocation).
-		Object(path).
+func (s *GCSStorage) getWriter(key string) *storage.Writer {
+	ioWriter := s.client.Bucket(s.manifestBucket).
+		Object(path.Join(s.keyPrefix, key)).
```
tgeoghegan, line 128: nit: `path.Join` will use `\` to join path components on Windows, but I think GCS and/or S3 might expect `/`. We don't support running `deploy-tool` or any part of this application on Windows, though, so I think it's OK to avoid dealing with this.

tgeoghegan (approved):

manifest-updater/manifest/storage.go
BranLwyd: [`path`](https://pkg.go.dev/path) will always use (forward) slashes, even on platforms that use backslashes to separate filesystem paths:
> The path package should only be used for paths separated by forward slashes, such as the paths in URLs. This package does not deal with Windows paths with drive letters or backslashes; to manipulate operating system paths, use the path/filepath package.

([`path/filepath`](https://pkg.go.dev/path/filepath) is for filesystem paths, and is aware of backlash vs forward slash.)

tgeoghegan: Oh, cool, I must be thinking of some other language's `path` type.

PR #1022, opened 2021-10-07
Remove default values for `aws_profile`, `victorops_routing_key`.

branlwyd:
...and specify `victorops_routing_key` in each environment variables file that didn't already specify it.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -29,7 +29,7 @@ variable "aws_region" {
 }
 
 variable "aws_profile" {
-  type    = string
+  type = string
 }
 
 variable "localities" {
@@ -176,7 +176,7 @@ variable "packet_encryption_rotation" {
 
 variable "pushgateway" {
   type        = string
-  default     = ""
+  default     = "prometheus-pushgateway.monitoring:9091"
   description = "The location of a pushgateway in host:port form. Set to prometheus-pushgateway.default:9091 to enable metrics"
 }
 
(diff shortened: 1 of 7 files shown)


terraform/main.tf
```diff
@@ -215,7 +214,6 @@ variable "prometheus_server_persistent_disk_size_gb" {
 variable "victorops_routing_key" {
   type        = string
   description = "VictorOps/Splunk OnCall routing key for prometheus-alertmanager"
-  default     = "bogus-routing-key"
```
tgeoghegan, line 218: We need _some_ value here because otherwise, the first `make apply` of an env (i.e. the first one that instantiates the `monitoring` module) will fail, because it won't have a value to interpolate into `kubernetes_secret.prometheus_alertmanager_config`.

We should document in either/both of `docs/monitoring.md` or `terraform/README.md` what value is needed here. For any prod or staging environment that is going to be hooked up to Splunk OnCall, the operator needs to identify a routing key (and possibly create one). For dev envs that don't need to emit alerts anywhere past the local `prometheus-alertmanager`, `bogus-routing-key` suffices.

BranLwyd: OK -- I've restored it. I think it would be acceptable (& maybe safer) to fail if this isn't specified, and use the failure as a reminder to either set an appropriate value (for real envs), or a purposefully-bogus one (for dev/test envs). That is, default to failure, not to no-alert-routing.

OTOH, we purposefully copy around env files to create new environments already, so I think in practice the bigger problem would be forgetting to update an already-specified routing key & sending alerts to the wrong routing key. Given that, I think the above consideration would only really matter if we were to introduce an "empty" template vars file, intended to be copied & filled in for new environments.

PR #1027, opened 2021-10-08
workflow-manager: accept aggregation-window override.

BranLwyd:
This is implemented by the new `--aggregation-override-timestamp` flag.
Normally (without this flag), the most recently completed aggregation
window (shifted by a grace period) will be aggregated. With this flag,
the aggregation window containing the override timestamp will be
aggregated instead.

Note that for an aggregation to actually be triggered, it still must
pass the other requirements. Specifically, the relevant validations must
have been written, and no task marker must exist for the window.

### DIFF
--- a/workflow-manager/main.go
+++ b/workflow-manager/main.go
@@ -31,36 +31,44 @@ import (
 // BuildInfo is generated at build time - see the Dockerfile.
 var BuildInfo string
 
-var k8sNS = flag.String("k8s-namespace", "", "Kubernetes namespace")
-var ingestorLabel = flag.String("ingestor-label", "", "Label of ingestion server")
-var isFirst = flag.Bool("is-first", false, "Whether this set of servers is \"first\", aka PHA servers")
-var maxAge = flag.String("intake-max-age", "1h", "Max age (in Go duration format) for intake batches to be worth processing.")
-var ingestorInput = flag.String("ingestor-input", "", "Bucket for input from ingestor (s3:// or gs://) (Required)")
-var ingestorIdentity = flag.String("ingestor-identity", "", "Identity to use with ingestor bucket (Required for S3)")
-var ownValidationInput = flag.String("own-validation-input", "", "Bucket for input of validation batches from self (s3:// or gs://) (required)")
-var ownValidationIdentity = flag.String("own-validation-identity", "", "Identity to use with own validation bucket (Required for S3)")
-var peerValidationInput = flag.String("peer-validation-input", "", "Bucket for input of validation batches from peer (s3:// or gs://) (required)")
-var peerValidationIdentity = flag.String("peer-validation-identity", "", "Identity to use with peer validation bucket (Required for S3)")
-var aggregationPeriod = flag.String("aggregation-period", "3h", "How much time each aggregation covers")
-var gracePeriod = flag.String("grace-period", "1h", "Wait this amount of time after the end of an aggregation timeslice to run the aggregation")
-var pushGateway = flag.String("push-gateway", "", "Set this to the gateway to use with prometheus. If left empty, workflow-manager will not use prometheus.")
-var dryRun = flag.Bool("dry-run", false, "If set, no operations with side effects will be done.")
-var taskQueueKind = flag.String("task-queue-kind", "", "Which task queue kind to use.")
-var intakeTasksTopic = flag.String("intake-tasks-topic", "", "Name of the topic to which intake-batch tasks should be published")
-var aggregateTasksTopic = flag.String("aggregate-tasks-topic", "", "Name of the topic to which aggregate tasks should be published")
-var maxEnqueueWorkers = flag.Int("max-enqueue-workers", 100, "Max number of workers that can be used to enqueue jobs")
+// Flags.
+var (
+	k8sNS                  = flag.String("k8s-namespace", "", "Kubernetes namespace")
+	ingestorLabel          = flag.String("ingestor-label", "", "Label of ingestion server")
+	isFirst                = flag.Bool("is-first", false, "Whether this set of servers is \"first\", aka PHA servers")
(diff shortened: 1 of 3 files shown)


BranLwyd:
I tested this against my dev environment and was able to successfully re-run a previous aggregation. My steps were:

1. Delete the existing task marker.
2. Delete the existing sum part. (not technically necessary, but made it easier to validate that the re-aggregation actually occurred)
3. Run the workflow manager on my workstation with a set of flags retrieved from a run of the job on GKE, plus the new `--aggregation-override-timestamp` flag [1]. (I dropped `--push-gateway` since it refers to a non-public name, too.)
4. Read the logs & verify that the sum part got written.

[1] Full invocation: `./workflow-manager --aggregation-period 30m --grace-period 1h --intake-max-age 6h --is-first=false --k8s-namespace asgard --ingestor-label ingestor-1 --ingestor-input gs://prio-bran-facil-asgard-ingestor-1-ingestion --own-validation-input gs://prio-bran-facil-asgard-ingestor-1-own-validation --peer-validation-input gs://prio-bran-facil-asgard-ingestor-1-peer-validation --task-queue-kind gcp-pubsub --intake-tasks-topic bran-facil-asgard-ingestor-1-intake --aggregate-tasks-topic bran-facil-asgard-ingestor-1-aggregate --gcp-project-id prio-bran-dev --aws-sns-region "" --aggregation-override-timestamp 202109292330`

workflow-manager/main.go
```diff
@@ -31,36 +31,44 @@ import (
...
+	pushGateway            = flag.String("push-gateway", "", "Set this to the gateway to use with prometheus. If left empty, workflow-manager will not use prometheus.")
+	dryRun                 = flag.Bool("dry-run", false, "If set, no operations with side effects will be done.")
+	taskQueueKind          = flag.String("task-queue-kind", "", "Which task queue kind to use.")
+	intakeTasksTopic       = flag.String("intake-tasks-topic", "", "Name of the topic to which intake-batch tasks should be published")
+	aggregateTasksTopic    = flag.String("aggregate-tasks-topic", "", "Name of the topic to which aggregate tasks should be published")
+	maxEnqueueWorkers      = flag.Int("max-enqueue-workers", 100, "Max number of workers that can be used to enqueue jobs")
+
+	// Aggregation window flags, which determine which aggregation window will be aggregated (if not already aggregated).
+	// Normally, aggregation occurs for the window including the point in time of (now - grace-period).
+	// If aggregation-override-timestamp is specified, the aggregation window containing the override point will be aggregated instead.
```
tgeoghegan, line 55: Lines in this section are super long anyway due to the flag descriptions, but let's wrap the comments to 80 chars anyway.

tgeoghegan, line 55:
```suggestion
	// Aggregation window flags, which determine which aggregation window will be aggregated (if not already aggregated).
	// Normally, aggregation occurs for the window including the point in time of (now - grace-period).
	// If aggregation-override-timestamp is specified, the aggregation window containing the override point will be aggregated instead of the most recent aggregation window.
```

workflow-manager/main.go
```diff
@@ -619,3 +621,17 @@ func enqueueIntakeTasks(
...
+	return func(now time.Time) wftime.Interval {
+		return wftime.AggregationInterval(now, aggregationPeriod, gracePeriod)
+	}
+}
+
+func overrideAggregationWindow(when time.Time, aggregationPeriod time.Duration) aggregationIntervalFunc {
+	return func(time.Time) wftime.Interval {
+		return wftime.AggregationIntervalIncluding(when, aggregationPeriod)
+	}
+}
```
tgeoghegan, line 637: nit: this type and these functions feel like they belong in package `time`.

workflow-manager/main.go
```diff
@@ -31,36 +31,44 @@ import (
...
-
-// Define flags and arguments for other task queue implementations here.
-// Argument names should be prefixed with the corresponding value of
-// task-queue-kind to avoid conflicts.
+// Flags.
+var (
+	k8sNS                  = flag.String("k8s-namespace", "", "Kubernetes namespace")
+	ingestorLabel          = flag.String("ingestor-label", "", "Label of ingestion server")
+	isFirst                = flag.Bool("is-first", false, "Whether this set of servers is \"first\", aka PHA servers")
+	maxAge                 = flag.Duration("intake-max-age", time.Hour, "Max age (in Go duration format) for intake batches to be worth processing.")
```
tgeoghegan, line 39: TIL `flag.Duration`. Thanks!

tgeoghegan (changes requested):
Nice work! I left some nits but don't think they're blockers. I'm especially glad to learn that this can be run from the workstation of a sufficiently privileged operator. I had guessed we'd have to craft some `kubectl apply` invocations to throw job instances into clusters.

I have one change I'd like, which is to augment the logging in `enqueueAggregationTask`. Right now, we have

```
	aggregationTask.PrepareLog(log.Info()).
		Str("aggregation window", aggregationWindow.String()).
		Msg("Scheduling aggregation task")
```
I think we will want to know how many batches are included in the aggregation when we schedule these backfills, so that we can confirm that the two sides are producing sum parts that can be summed. Logging the UUID of each batch would be too much noise, but logging `batchCount` alongside the aggregation window should suffice. Then, we could have both sides run `workflow-manager --aggregation-override-timestamp <whatever> --dry-run`, compare the proposed aggregations and pull the trigger iff it all lines up.

workflow-manager/main.go
BranLwyd: Done.

BranLwyd:
It turns out the batch count is already logged -- check out `PrepareLog` in `task.go`:
```
func (a Aggregation) PrepareLog(event *zerolog.Event) *zerolog.Event {
	return event.
		Str("trace ID", a.TraceID.String()).
		Str("aggregation ID", a.AggregationID).
		Int("batch count", len(a.Batches))
}
```

(I confirmed this shows up in practice in the log line you called out, and the number that gets logged matches the logging I would have added there.)

PR #1042, opened 2021-10-18
Add autoscaling for intake/aggregate tasks.

### LINKED ISSUE #484: Autoscale deployments based on PubSub queue depth

branlwyd:
Autoscaling is based on the queue depth: if there is a queue, we'll
scale up. The current policy is to stabilize over 5 minutes, then
add/remove one replica per minute.

Fixes #484.

### DIFF
--- a/terraform/modules/custom_metrics/custom_metrics.tf
+++ b/terraform/modules/custom_metrics/custom_metrics.tf
@@ -354,10 +354,6 @@ resource "kubernetes_deployment" "aws_custom_metrics_adapter" {
             container_port = 6443
             name           = "https"
           }
-          port {
-            container_port = 8080
-            name           = "http"
-          }
           volume_mount {
             name       = "temp-vol"
             mount_path = "/tmp"
@@ -384,11 +380,6 @@ resource "kubernetes_service" "aws_custom_metrics_adapter" {
       name        = "https"
       port        = 443
       target_port = 6443
-    }
-    port {
-      name        = "http"
-      port        = 80
-      target_port = 8080
     }
     selector = { app = "custom-metrics-adapter" }
   }
--- a/terraform/modules/kubernetes/kubernetes.tf
+++ b/terraform/modules/kubernetes/kubernetes.tf
@@ -387,7 +387,6 @@ resource "kubernetes_deployment" "intake_batch" {
     namespace = var.kubernetes_namespace
   }
   spec {
-    replicas = var.min_intake_worker_count
     selector {
       match_labels = {
         app      = "intake-batch-worker"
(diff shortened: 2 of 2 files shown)


terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -558,6 +657,96 @@ resource "kubernetes_deployment" "aggregate" {
     }
   }
 }
+
+resource "kubernetes_manifest" "aggregate_queue_depth_metric" {
```
tgeoghegan, line 661: There's no GCP equivalent to this. Does the GCP metrics adapter automagically create the manifest?

tgeoghegan (changes requested):
A few nits but I'm really happy with how simple the target rule is. I'll be curious to see how this contends with a production locality like `us-ca-apple`.

terraform/modules/kubernetes/kubernetes.tf
tgeoghegan: This reminds me of network configuration, which mostlu Just Works in GKE, but EKS requires you to jump through a ton of hoops to construct subnets and security groups and so on. Anyway, I agree with your approach here.

terraform/modules/custom_metrics/custom_metrics.tf
```diff
@@ -0,0 +1,395 @@
...
+    port {
+      name        = "https"
+      port        = 443
+      target_port = 6443
+    }
+    port {
+      name        = "http"
+      port        = 80
+      target_port = 8080
+    }
```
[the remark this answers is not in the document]
BranLwyd: I believe [this page](https://kubernetes.io/docs/concepts/security/controlling-access/#api-server-ports-and-ips) gives a description of the two ports. tl;dr is: the secure port (443) is used for "normal" serving, uses TLS, has all the needed authn/authz functionality, etc; the insecure port (80) is used for turnup, bypasses a lot of security, but only binds to localhost (ie loopback) by default.

Given that last bit, I suspect that we don't actually need to expose port 80, because the metrics server won't be listening over the network on that port anyway. I tried removing it from my dev-env deployment and things seem to work OK (at least, the metric adapter logs look fine, and the horizontal-pod-autoscalers can still get their metrics). So I dropped port 80 entirely.

[and to close the loop on why it was there in the first place: I copied it blindly from [Amazon's suggested YAML config](https://raw.githubusercontent.com/amazon-archives/k8s-cloudwatch-adapter/7b73805a92a4c8dbf45b73125e8d535c867e3029/deploy/adapter.yaml)]

tgeoghegan (approved):
I appreciate the thorough research here! I agree with you on the question of `insecure_skip_tls_verify`. This seems safe to me since our application's deployment is on a purely private network.

PR #1044, opened 2021-10-21
terraform: force 3 nodes in staging-intl

tgeoghegan:
`prometheus-server` and `prometheus-alertmanager` use
`PersistentVolumeClaim`s bound to `PersistentVolume`s backed by EBS
volumes in `us-west-2a`. We force the EKS cluster node pool to be at
least 3 VMs to guarantee that at least one node will be scheduled into
`us-west-2a`, without which we cannot run the metrics stack.

### DIFF
--- a/terraform/variables/staging-intl.tfvars
+++ b/terraform/variables/staging-intl.tfvars
@@ -25,8 +25,8 @@ ingestors = {
   }
 }
 cluster_settings = {
-  initial_node_count = 2
-  min_node_count     = 1
+  initial_node_count = 3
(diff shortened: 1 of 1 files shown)


terraform/variables/staging-intl.tfvars
```diff
@@ -26,7 +26,7 @@ ingestors = {
 }
 cluster_settings = {
   initial_node_count = 2
```
BranLwyd, line 28: Not sure if this is necessary, but change `initial_node_count` as well?

PR #1061, opened 2021-10-29
Avoid a common race, fix gosec G601 errors

jmhodges:
We pull these variables out here to avoid a common race described at
https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable
. This was found with `gosec` and triggered its G601 error.


tgeoghegan (approved):
Thank you very much for this PR! The change looks sound and the broken Rust coverage check is due to an unrelated problem: https://github.com/abetterinternet/prio-server/pull/1078

PR #1064, opened 2021-10-29
use not-deprecated zap APIs

jmhodges:
This fixes staticcheck SA1019 because zap has deprecated the other
method call.


tgeoghegan (approved):
Merging in spite of `faciitator-coverage` failure, which is being addressed in #1078

Issue #294, opened 2021-01-05
Implement golang-ci lint


jsha:
This can help catch a variety of errors, including (via the gosec lint) the "captured loop variable" from https://github.com/abetterinternet/prio-server/pull/292/commits/7f09dbac1440fdf31d28d8f6654ca35c2d3f6728.

https://github.com/letsencrypt/boulder/blob/main/.golangci.yml
https://github.com/letsencrypt/boulder/blob/a214fb7757c56fb92262b6a265f49d5fcef34e05/test.sh#L234-L236
https://github.com/letsencrypt/boulder/blob/a214fb7757c56fb92262b6a265f49d5fcef34e05/test/boulder-tools/build.sh#L32-L33

PR #1065, opened 2021-10-29
tell gosec to ignore G101 on okay variables

jmhodges:
gosec tries very hard to look for hardcoded credentials, and it can be a
bit broad.

This is perhaps the most controversial PR towards #294. We might prefer
to silence this error globallly in the golangci-lint config, but it's
only in a couple of places.

### DIFF
--- a/deploy-tool/key/key_backup.go
+++ b/deploy-tool/key/key_backup.go
@@ -14,6 +14,7 @@ import (
 )
 
 const (
+	// #nosec G101
 	secretUIDLabel = "kubernetes_uid"
 )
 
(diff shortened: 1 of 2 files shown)


tgeoghegan:
Given [how this lint is implemented](https://securego.io/docs/rules/g101.html), I'd prefer to disable gosec G101. If the lint were scanning values for PEM encoded private keys or Base64 encoded SSH keys, then it would seem more valuable, but IMO  `password` and `secret` are perfectly reasonable variable names. On top of that, [GitHub already does language-agnostic secret scanning in source code](https://docs.github.com/en/code-security/secret-scanning/about-secret-scanning), so I don't think we need to rely on gosec.

I made a note to disable the lint in #1066, so I'm inclined to close this PR. OK with you @jmhodges?

jmhodges:
SGTM! I'll add it to the other PR

PR #1063, opened 2021-10-29
correct a typo in Enqueuer comment

jmhodges:

PR #1062, opened 2021-10-29
run gofmt -s on the codebase

jmhodges:
Only two files needed changes.

PR #1066, opened 2021-10-29
add Go linters using golangci-lint

### LINKED ISSUE #294: Implement golang-ci lint

jmhodges:
This was based off, initially, the letsencrypt/boulder golangci-lint
config, but made simpler by allowing more checks and ignoring fewer
APIs. This has led to a number of other PRs to correct lint errors.
Those are catalogued in the comments of #294.

There's a job entry for each Go directory in the Github Action CI config
because golangci-lint needs a `go.mod` in the same directory to work
correctly, cannot operate on multiple Go modules at once, and, when set
as different steps in one job, tosses out spurious "File exists" errors
from tar if each run tries to use the same FS.

This also updates deploy-tool-build.yml to use a new golangci-lint
version for the new config format. That job scan be removed in another PR
after this repo isn't configured to require those checks during PRs.

Fixes #294

(Updated issue body to match the current version of the PR)


tgeoghegan:
We will need to rebase this change on top of main after merging #1061, #1062, #1063, #1064, #1065, so that we can show that the new golanci-lint checks run successfully. Additionally, we need to delete the `deploy-tool-lint` and `task-replayer-lint` jobs from `deploy-tool-build.yml` as they are replaced by your new job definitions (actually, we should move everything in that file into `ci-build.yml` and then delete `deploy-tool-build.yml`). Additionally, it seems like [the new check fails](https://github.com/abetterinternet/prio-server/runs/4070516001?check_suite_focus=true), possibly because it's not being pointed at directories correctly?

@jmhodges you've already done 99% of the work here (thank you!) so I am happy to deal with the remaining tweaks and debugging by tacking on another commit to this PR. Please let me know if you'd rather do it yourself.

jmhodges:
I've left these as separate commits, but I'm happy to squash them all and update the original commit text!

jmhodges:
Okay, things are better and after some grappling with GitHub Actions, they're running clean. Still happy to squash any of these commits.

Two notes: 

I couldn't remove those now duplicated lints because they're still being checked for in PRs by the GitHub repo config. 

Also, I decided the deploy-tool-build.yml move was too much for this PR and would likely run into the same PR check config problem.

tgeoghegan:
Those are both good points. I filed #1083 to track migrating `deploy-tool-build.yml` and fixing the required checks. Thanks a ton for all this great work! We don't have a formal policy on merge commits, but my personal preferred practice is to squash into fewer commits, to reduce the noise in `main`'s commit history and improve the odds that any commit on `main` builds and passes tests. However I don't like to impose style rules on anyone unless I can produce a linter or some kind of automation to unambiguously enforce them, so @jmhodges please rebase and/or squash this however you see fit, let me know when you're happy with the commits, and I'll merge.

tgeoghegan (approved):

jmhodges:
So new lint failures snuck in, so I'm going to fix this up. This is close, though.

tgeoghegan:
I merged your #1086, and judging by the failures in this PR, this should be good to go if rebased on that change.

jmhodges:
Yeah, there's a UX change I'd like to make that I found from this error. I'm going to do that before saying "go".

Each golangci-lint step expects to own the build and pkg cache, even from each other. This turns into a lot of verbose, and unhelpful "File exists" errors from tar. Their suggested solution is to split the steps into jobs. The `skip-build-cache` and `skip-pkg-cache` booleans would cause the later steps to be much slower.

So, once I'm not out and about, I'll just skip that and make them individual jobs.

tgeoghegan:
@jmhodges thank you for your persistence on this! If you can confirm you're happy with the state of this change, I will merge it before we foist more merge conflicts on you.

jmhodges:
Okay, this is ready to go!

BranLwyd (approved):

tgeoghegan:
Thanks a ton for all this work @jmhodges, sincerely appreciated!

jmhodges:
Oh, btw, you may know this but: you could configure this repo in GitHub to make the merge button do "Squash merge" by default. I dunno if that satisfies your "automation" requirement for multiple vs single commit styling, but I think it's p close, at least!

tgeoghegan:
> Oh, btw, you may know this but: you could configure this repo in GitHub to make the merge button do "Squash merge" by default. I dunno if that satisfies your "automation" requirement for multiple vs single commit styling, but I think it's p close, at least!

I'm picky: I think the commits that GitHub constructs when you squash merge are ugly, because the resulting commit message just concatenates all the original commits' messages. Basically want everyone to go manually rebase and squash their changes and write artful and thoughtful commit messages, and that's what I struggle to enforce automagically.

PR #1081, opened 2021-11-01
key-rotator: add key.Raw type, serialize/deserialize key.Key including a key.Raw.

BranLwyd:
key.Raw represents a single "raw" (i.e. unversioned) key of a particular type. The only non-test type currently implemented is P256.

key.Raw is now included in key versions instead of an untyped string `keyMaterial`; we can directly serialize/deserialize from JSON thanks to key.Raw implementing encoding.TextMarshaller & encoding.TextUnmarshaller.

Stacked on https://github.com/abetterinternet/prio-server/pull/1067.


key-rotator/key/raw.go
```diff
@@ -0,0 +1,233 @@
...
+
+// Raw represents a raw (i.e. unversioned) asymmetric cryptographic key,
+// including both the private & public portions. It has functionality related
+// to serialization of the key.
+type Raw struct{ k raw }
+
+var _ encoding.BinaryMarshaler = Raw{}
+var _ encoding.BinaryUnmarshaler = &Raw{}
+var _ encoding.TextMarshaler = Raw{}
+var _ encoding.TextUnmarshaler = &Raw{}
```
tgeoghegan, line 62: Are these just here so that the compiler can enforce that `Raw` implements all four interfaces? If so, could you put in a comment explaining that?

key-rotator/key/raw.go
```diff
@@ -0,0 +1,233 @@
...
+	// keyType returns the type of the raw key.
+	keyType() Type
+
+	// equal determines if this raw is equal to the given raw, which can be
+	// assumed to be of the same key type.
+	equal(o raw) bool
+
+	// publicAsCSR returns a PEM-encoding of the ASN.1 DER-encoding of a
+	// PKCS#10 (RFC 2986) CSR over the public portion of the key, using the
+	// provided FQDN as the common name for the request.
```
tgeoghegan, line 150: nit: mention that the CSR will be signed with the private portion of the key (right?)

key-rotator/key/raw.go
```diff
@@ -0,0 +1,233 @@
...
+	// encoding of the public portion of the key, concatenated with the secret
+	// "D" scalar.
+	asX962Uncompressed() (string, error)
+
+	// asPKCS8 returns a base64 encoding of the ASN.1 DER-encoding of the key
+	// in PKCS#8 (RFC 5208) format.
+	asPKCS8() (string, error)
+}
+
+type p256 struct{ pk *ecdsa.PrivateKey }
```
tgeoghegan, line 167: nit: `pk` is a little ambiguous because it could be either private or public key. 
```suggestion
type p256 struct{ privKey *ecdsa.PrivateKey }
```

tgeoghegan (changes requested):

key-rotator/key/raw.go
BranLwyd: Yup, that's why. Comment added.

key-rotator/key/raw.go
BranLwyd: Yep, it's signed (at least per the documentation of `x509.CreateCertificateRequest`). Updated comments.

key-rotator/key/raw.go
BranLwyd: Done.

tgeoghegan (approved):

PR #1082, opened 2021-11-02
Move manifest data definitions from manifest-updater to key-rotator, add UpdateKeys functionality.

BranLwyd:
I plan to move functionality out of manifest-updater & into key-rotator
as I start to use it. The manifest-updater directory is eventually going
away, and this will help me ensure that I review code as I import it.

A few notes on UpdateKeys' functionality:
 * We keep existing batch signing/packet encryption manifest data if it
   already exists. This isn't for performance/safety (though those
   features are nice too, I suppose), but instead to keep these
   manifests easily comparable without needing to parse public-key
   information & to reduce the number of manifest writes we do.
   (otherwise, we would not be able to compare manifests with a simple
   `==`, since CSRs are nondeterministic)
 * The prefix logic is intended to allow us to avoid re-writing the
   currently-existing, non-timestamped key. The idea is that this key
   version will have a zero creation time; if we set the key prefix to
   an appropriate value (e.g.
   "staging-facil-asgard-ingestor-1-batch-signing-key") then this
   pre-existing key will correctly match based on key ID. Newly-created
   keys will have a similar key ID, but suffixed by the Unix timestamp
   they were created at.

Also, bump all Go projects to Go 1.17 (from 1.15).


.github/workflows/ci-build.yml
```diff
@@ -114,7 +114,7 @@ jobs:
     - name: Setup go
       uses: actions/setup-go@v1
       with:
-        go-version: 1.15
+        go-version: 1.17
```
tgeoghegan, line 117: Oof, good catch! It's a shame that dependabot doesn't update this (but, well, it can only do so much). I filed #1085 so we can document dependency updated like this that are outside of depdendabot's remit.

manifest-updater/build.sh
```diff
@@ -7,4 +7,6 @@ BUILD_ID="$(git symbolic-ref --short HEAD 2>/dev/null || true)+${COMMIT_ID}"
 BUILD_TIME="$(date)"
 BUILD_INFO="${BUILD_ID} - ${BUILD_TIME}"
 
-docker build --tag letsencrypt/prio-manifest-updater --build-arg BUILD_INFO="${BUILD_INFO}" .
+cd ..
+
+docker build -f manifest-updater/Dockerfile --tag letsencrypt/prio-manifest-updater --build-arg BUILD_INFO="${BUILD_INFO}" .
```
tgeoghegan, line 12: nit: does this work?
```suggestion
docker build -f manifest-updater/Dockerfile --tag letsencrypt/prio-manifest-updater --build-arg BUILD_INFO="${BUILD_INFO}" ..
```
Seems a bit less stateful than `cd`.

key-rotator/manifest/manifest.go
```diff
@@ -30,6 +37,68 @@ type DataShareProcessorSpecificManifest struct {
...
+
+func (m DataShareProcessorSpecificManifest) UpdateKeys(cfg UpdateKeysConfig) (DataShareProcessorSpecificManifest, error) {
+	// Copy the current manifest, clearing any existing batch signing/packet encryption keys.
+	newM := m
+	newM.BatchSigningPublicKeys, newM.PacketEncryptionKeyCSRs = BatchSigningPublicKeys{}, PacketEncryptionKeyCSRs{}
+
+	// Update batch signing key.
+	for _, v := range cfg.BatchSigningKey {
+		kid := cfg.BatchSigningKeyIDPrefix
+		if !v.CreationTime.IsZero() {
```
tgeoghegan, line 58: In what case would a `key.Key` version not have a `CreationTime`? My guess is it's in the case where we are migrating from a non-versioned key discovered in Kubernetes secrets? If I'm right, then there should only ever be one key version in a key with a zero `CreationTime`. This function already seems to make that assumption, and if I'm reading it right, the first zero `CreationTime` encountered when iterating the slice will "win" and be present in the manifest; others are silently dropped. We should have an explicit check for multiple versions with no `CreationTime` and return an error in that case.

key-rotator/manifest/manifest.go
```diff
@@ -63,7 +132,7 @@ type BatchSigningPublicKey struct {
 	// PKIX SubjectPublicKeyInfo structure. It must be an ECDSA P256 key.
 	PublicKey string `json:"public-key"`
 	// Expiration is the ISO 8601 encoded UTC date at which this key expires.
-	Expiration string `json:"expiration"`
+	Expiration string `json:"expiration,omitempty"`
```
tgeoghegan, line 135: Careful with this -- it's possible that the MITRE portal server and the Apple/Google ingestion servers' manifest parsers require `expiration` to be present even if they ignore the value.

key-rotator/manifest/manifest.go
```diff
@@ -30,6 +37,68 @@ type DataShareProcessorSpecificManifest struct {
...
+			newBSPK = BatchSigningPublicKey{PublicKey: pkix}
+		}
+		newM.BatchSigningPublicKeys[kid] = newBSPK
+	}
+
+	// Update packet encryption key.
+	for _, v := range cfg.PacketEncryptionKey {
+		if !v.Primary {
+			continue
+		}
```
tgeoghegan, line 79: Should we error out if more than one key version is marked primary?

tgeoghegan (changes requested):

manifest-updater/build.sh
BranLwyd: It does! Updated accordingly.

key-rotator/manifest/manifest.go
BranLwyd: Hmm, that's a good point.

Even if the other parties are reading this field, I strongly suspect no one is respecting it -- spot-checking a few prod-us manifests, many of these expiration dates have already passed. And IMO it's better for internal keys to not have a built-in expiration date -- when possible, rotation/versioning is a better solution than expiration -- so I think the ideal stable state would be for this field to not exist at all.

Do you have any suggestions here? I could write a date long in the future, but that feels hackish. Maybe we should either (a) ask [and risk no one having a clue what this field even is] or (b) skip writing it on a first pass and see who breaks.

key-rotator/manifest/manifest.go
BranLwyd: You're correct about when `key.Key` would have a zero `CreationTime` -- if and only if it is a "migrated" key version from the currently-deployed serialization strategy.

There are a number of sharp edges/risky cases that "shouldn't" be possible, but I strongly agree that it would be good to validate. OTOH, I am still up in the air about the exact validations & where they fit into the larger program logic.

At the least, I want to implement the following validations:
* Keys must either have no versions, or exactly one primary version.
* Post-rotation, keys must have at least one version.
* When being used to update keys in a manifest, keys must have at least one version.
* When being used to update keys in a manifest, keys must have versions with distinct creation times.
* When being used to update keys in a manifest, existing key versions' public portions should match with the public keys written to the manifest, matching by creation time/key ID (both of which would provide the same matching).
* Post-update, manifests should have exactly one packet encryption key version.
* Post-update, manifests should have at least one batch signing key version.

Any more that you can think of?

If you are OK with it, I would like to wait to add this kind of validation until I have the first version of the "main" functionality written tying everything together -- I want to think through where these validations best fit in a whole-program sense, which should be tied directly to data types vs included in program logic, which should be bypassable by passing the right --unsafe_no_really_force_rotation flag(s) vs should really always be applied, etc.

key-rotator/manifest/manifest.go
BranLwyd: Yes, but see my response to the other comment regarding validations.

key-rotator/manifest/manifest.go
tgeoghegan: I'm primarily worried with avoiding any silent "failures", scenarios where malformed or unexpected keys cause us to use whatever value an iterator coughs up first without errors being raised or anomalies being logged. I'm OK with deferring that to later since a good place to centralize that sort of validation might come along as more pieces of the overall system fall into place.

key-rotator/manifest/manifest.go
tgeoghegan: The field is absolutely being ignored, but I'm worried that someone's `json.Unmarshal` (or equivalent) call is going to fail because their parser marked the field as non-optional. I lean towards putting a far-future date in here--it's dumb, but we can be confident it won't break anyone and also it's arguably correct in the sense that maybe we will expire that key in the year 2525.

manifest-updater/build.sh
tgeoghegan: Do you perhaps need to push your latest changes for them to show up here?

BranLwyd: oops, yup. sorry about that.

key-rotator/manifest/manifest.go
BranLwyd: I added the list above to https://github.com/abetterinternet/prio-server/issues/24 to help ensure it doesn't get lost.

tgeoghegan (approved):
LGTM but let's wait for #1066 to land (I think you will need to rebase or at least make a further change to turn on `golangci-lint` for `key-rotator`).

PR #1084, opened 2021-11-03
Key rotator: introduce storage.Manifest, stolen & adapted from manifest-updater's storage.

BranLwyd:
This is a fairly large change, but I broke it up into small individual commits that should hopefully explain why I made the changes I made. (Feel free to ask, too.)

I updated deploy-tool to use this new package. Other than its unit tests, I also tested briefly against my dev environment -- deploying a data processor-specific manifest still works, I did not test global manifest.

Stacked on https://github.com/abetterinternet/prio-server/pull/1082.


deploy-tool/main.go
```diff
@@ -74,6 +81,32 @@ type SingletonIngestor struct {
 	BatchSigningKeyName       string `json:"batch_signing_key_name"`
 }
 
+type ManifestStorage interface {
```
tgeoghegan, line 84: It seems like we only have this and interface `ManifestWriter` anymore so that `deploy-tool/main_test.go` can provide its own `FakeManifestStorage` that implements the interface, but it seems like we could delete this if `deploy-tool/main_test.go` could use `memDS` from `key-rotator/storage/manifest_test.go`. Then again, the objective is to delete `deploy-tool` so it's probably not worth the hassle of sharing test code.

key-rotator/storage/manifest.go
```diff
@@ -0,0 +1,265 @@
...
+
+func (ds gcsDatastore) get(ctx context.Context, key string) ([]byte, error) {
+	r, err := ds.gcs.Bucket(ds.bucket).Object(key).NewReader(ctx)
+	if err != nil {
+		if err == storage.ErrObjectNotExist {
+			err = ErrObjectNotExist
+		}
+		return nil, fmt.Errorf("couldn't retrieve gs://%s/%s: %w", ds.bucket, key, err)
+	}
+	defer r.Close() // errchecked Close is below
```
tgeoghegan, line 190: Ooh, nice, this is much better than checking whether `Close()` failed on every error path. I'm a little concerned about the behavior of calling `Close()` twice, though:

> [The behavior of Close after the first call is undefined. Specific implementations may document their own behavior.](https://pkg.go.dev/io#Closer)

[The docs on `storage.Reader.Close()`](https://pkg.go.dev/cloud.google.com/go/storage#Reader.Close) don't say anything about whether it's safe. WDYT?

Orthogonally: I was pedantic about this in the previous `manifest-updater/manifest/storage.go`, but I think we only need to be super careful about checking `Close()` on the write path. On the read path, we will know that our reads failed because we can't parse a JSON object.

key-rotator/storage/manifest.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("couldn't read gs://%s/%s: %w", ds.bucket, key, err)
+	}
+	if err := r.Close(); err != nil {
+		return nil, fmt.Errorf("couldn't close gs://%s/%s: %w", ds.bucket, key, err)
+	}
+	return objBytes, nil
+}
+
+func (ds gcsDatastore) put(ctx context.Context, key string, data []byte) error {
```
tgeoghegan, line 201: We lost the structured log statement from `GCSStorage.writeManifest`

key-rotator/storage/manifest.go
```diff
@@ -0,0 +1,265 @@
...
+		Key:    aws.String(key),
+	})
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == s3.ErrCodeNoSuchKey {
+			err = ErrObjectNotExist
+		}
+		return nil, fmt.Errorf("couldn't retrieve s3://%s/%s: %w", ds.bucket, key, err)
+	}
+	r := objOut.Body
+	defer r.Close() // errchecked Close is below
```
tgeoghegan, line 241: As in the GCS case, I'm not certain if it's safe to `Close()` twice, and also I don't think we need to be as fussy about checking on `Close()` on the read path.

tgeoghegan (changes requested):
I really like the pattern of a concrete struct type wrapping a private interface type. I'm going to remember that idiom.

key-rotator/storage/manifest.go
BranLwyd: You're right -- I'm pretty sure this is safe in practice, but it's not promised to be safe by any of the APIs.

I switched to a defer, which will always call Close exactly once.

deploy-tool/main.go
BranLwyd: I made this change based on the principles of "define interfaces at point-of-use, not point-of-implementation" &  "take interfaces, return concrete types". But I had the same testing reservations that you do -- it'd be nice if we could centralize a single "for-tests" Manifest implementation. And then I noticed [Effective Go](https://golang.org/doc/effective_go#generality) suggests that sometimes it _is_ OK to return an interface (and the stdlib certainly does in some cases).

OTOH, out-of-package tests probably want a fake that lets them see what manifests are read & written, rather than what bytes are read & written. But storage.Manifest w/ memDS works at the level of bytes (because I really do want that level of control/testing when testing Manifest in-package).

I settled on making the struct Manifest private, introducing a new Manifest interface with the expected methods, and then adding a storage/test package with a Manifest fake inspired by deploy-tool's FakeManifestStorage to show that this layout works for a centralized testing implementation. I think this is worthwhile because I'll reuse it for the key-rotator main logic.

key-rotator/storage/manifest.go
BranLwyd: Restored.

key-rotator/manifest/manifest.go
```diff
@@ -132,7 +132,7 @@ type BatchSigningPublicKey struct {
 	// PKIX SubjectPublicKeyInfo structure. It must be an ECDSA P256 key.
 	PublicKey string `json:"public-key"`
 	// Expiration is the ISO 8601 encoded UTC date at which this key expires.
-	Expiration string `json:"expiration"`
+	Expiration string `json:"expiration,omitempty"`
```
tgeoghegan, line 135: Did we discuss this in this or another PR? I'm not sure if it's a good idea to emit manifests without an expiration value.

tgeoghegan (approved):

key-rotator/manifest/manifest.go
BranLwyd: discussed in #1082, fixed there, merge from main did not pick up this change & I missed it :(

PR #1090, opened 2021-11-04
Delete manifest updater.

BranLwyd:
With https://github.com/abetterinternet/prio-server/pull/1084, I have taken everything I want from manifest-updater. I don't think the rest will be needed in the future, and it's a minor maintenance burden, so `git rm` it. (It'll always be available in git history if we do need it.)

Stacked on https://github.com/abetterinternet/prio-server/pull/1084.


tgeoghegan (commented):
image
🎉 
However I think you will need to rebase on #1066 because of how it changes `.github/workflows/`

tgeoghegan (approved):

PR #1098, opened 2021-11-08
Implement storage.Key, based on Kubernetes secret store.

BranLwyd:
Stacked on #1096.


key-rotator/storage/key.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func (k k8sKey) PutPacketEncryptionKey(ctx context.Context, locality string, key key.Key) error {
+	return k.putKey(ctx, k.packetEncryptionKeyName(locality), key, serializePacketEncryptionSecretKey)
+}
+
+func (k k8sKey) putKey(ctx context.Context, secretName string, key key.Key, serializeLiveVersions func(key.Key) ([]byte, error)) error {
+	log.Info().
+		Str("storage", "kubernetes").
+		Str("secret", secretName).
```
tgeoghegan, line 71: It'll be obvious from the secret name, but we could also log a field for the secret "kind", i.e. batch signing vs packet encryption.

tgeoghegan (approved):

PR #1114, opened 2021-11-11
Key rotator binary

BranLwyd:
Other than unit tests, I ran this against my dev environment in dry run mode & the results look reasonable.

There are a number of tasks before we are ready to deploy this in production & start rotating keys:
  * A number of validations need to be added (tracked in https://github.com/abetterinternet/prio-server/issues/24 -- some of these are implemented in tests already, but they need to be added to the main program logic)
  * Metrics/alerting
  * Automated key backup

### DIFF
--- a/key-rotator/main.go
+++ b/key-rotator/main.go
@@ -1,0 +1,414 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
+	"golang.org/x/sync/errgroup"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/abetterinternet/prio-server/key-rotator/key"
+	"github.com/abetterinternet/prio-server/key-rotator/manifest"
+	"github.com/abetterinternet/prio-server/key-rotator/storage"
+
+	_ "k8s.io/client-go/plugin/pkg/client/auth" // included for k8s client auth plugins
+)
+
+var (
+	// Required configuration.
+	prioEnv           = flag.String("prio-environment", "", "Required. The prio `environment`, e.g. 'prod-us' or 'prod-intl'")
+	namespace         = flag.String("kubernetes-namespace", "", "Required. The Kubernetes `namespace`, e.g. 'us-ca' or 'ta-ta'")
+	manifestBucketURL = flag.String("manifest-bucket-url", "", "Required. The URL of the manifest `bucket`, e.g. 's3://bucket-name' or 'gs://bucket-name'")
+	locality          = flag.String("locality", "", "Required. The Prio `locality`, e.g. 'us-ca' or 'ta-ta'")
+	ingestors         = flag.String("ingestors", "", "Required. Comma-separated list of `ingestors`, e.g. 'apple' or 'g-enpa'")
+	csrFQDN           = flag.String("csr-fqdn", "", "Required. FQDN to use as common name in generated CSRs")
+
+	// Rotation configuration.
+	batchSigningKeyCreateMinAge   = flag.Duration("batch-signing-key-create-min-age", 9*30*24*time.Hour, "How frequently to create a new batch signing key version")               // default: 9 months
+	batchSigningKeyPrimaryMinAge  = flag.Duration("batch-signing-key-primary-min-age", 7*24*time.Hour, "How old a batch signing key version must be before it can become primary") // default: 1 week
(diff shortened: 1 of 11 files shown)


key-rotator/main.go
```diff
@@ -0,0 +1,414 @@
...
+	prioEnvironment   string
+	csrFQDN           string
+	batchRotationCFG  key.RotationConfig
+	packetRotationCFG key.RotationConfig
+}
+
+func rotateKeys(ctx context.Context, cfg rotateKeysConfig) error {
+	// Retrieve keys & manifests.
+	log.Info().Msgf("Reading keys & manifests")
+	oldPacketEncryptionKey, oldBatchSigningKeyByIngestor, oldManifestByIngestor, err := readKeysAndManifests(ctx, cfg.keyStore, cfg.manifestStore, cfg.locality, cfg.ingestors)
```
tgeoghegan, line 209: nit: wrap long lines

key-rotator/main.go
```diff
@@ -0,0 +1,414 @@
...
+			return fmt.Errorf("couldn't update manifest for (%q, %q): %w", cfg.locality, ingestor, err)
+		}
+		newManifestByIngestor[ingestor] = newManifest
+	}
+
+	// Write keys, then write manifests.
+	// We write keys first so that on failure, we avoid the situation of having
+	// written the public portion of a key to some manifest, while not having
+	// written the associated private key to a secret (which would then be
+	// lost).
```
tgeoghegan, line 250: Suppose a key rotation fails in this manner. How do we recover from it or reconcile the state? Will the next run of key-rotator make a new key, or will it find the key that was generated last time and publish it in a manifest?

tgeoghegan (approved):
Looks cromulent to me, with the caveat that I am reviewing this after having gotten up at 3 AM for today's IETF meeting so I haven't been incredibly thorough.

key-rotator/main.go
BranLwyd: Done. I went for a "soft" limit of 100 columns for code (allowing things like human-readable strings etc to scroll off to the right pretty far in some cases), going over slightly if I deemed things more readable that way.

key-rotator/main.go
BranLwyd: The model I am using is that the "atomic" write operations are at the level of a single key or manifest. Given that a run of the `key-rotator` may have to write multiple keys or manifests, this means that we must deal with partial failures (i.e. some subset of the writes that we wish to do fail), potentially leaving the keys/manifests in an "inconsistent" state from what was intended.

Given that the `key-rotator` will be a periodically-running cron job, I choose to allow retries to occur with successive runs of the `key-rotator`. This implies that the `key-rotator` must be able to handle reading the "inconsistent" state of a previous run that failed partway through, and reconcile things appropriately.

All that said, here's how transient failures at various points will be handled by following runs:
* If the failure occurs before we write anything (e.g. in a read, or as part of the in-memory rotate-key/update-keys-in-manifest operations): the written state has not changed in this case. From the perspective of later runs of `key-rotator` it is as if this run of `key-rotator` did not happen at all; there is no inconsistent state to be concerned with.
* If the failure occurs partway through writing keys: some of the keys will have been rotated (new primary, new version, deleted old version); other keys will not have been. No manifests will have been modified; manifests which include the rotated keys will therefore be "stale." The next run will re-read the keys; re-attempt rotation of the keys that failed to write in the previous run; then re-attempt writing updated manifests to match the rotated state of all keys. Since manifests are evaluated for update & write whether or not the current run rotated any of the keys contained in the manifest, we will write updated manifests for keys rotated in previous runs.
* If the failure occurs partway through writing manifests: all of the rotated keys will have been written. On the next run, we will re-evaluate all manifests for update/write, which causes us to attempt to re-write the updated manifests based on the keys rotated in previous runs.

While transiently leaving keys/manifests in an inconsistent state must be OK (even on a successful run, there will be a small interval of time where we could observe inconsistency between keys & manifests), we can still get into trouble if we let manifests get _too_ stale. Specifically, if a packet encryption key got so far ahead of its related manifests that the key did not include the version advertised as primary in the manifest, we would be asking clients to encrypt with a key we no longer consider valid for decryption. And if a batch signing key got so far ahead of its related manifests that the primary key was not included in the manifest, we would be trying to sign data with a key that we do not recognize as valid for verification. For these reasons, alerting on failure will be useful -- even something as simple as "alert on N consecutive failures" would be fine as a first pass IMO. And we need to select our rotation parameters such that consistent partial failures don't cause us to fall into the "actually bad" scenarios faster than we can react operationally.

... and all of that is a very long-winded description of something that should probably go into the code. I've added comments & test cases to make this clearer/verify that things are actually as I say.

BranLwyd: (also, those conditions on how "far ahead" the PEK/BSK can get before things go wrong sound like things we can detect, and therefore might be included in validations. updated https://github.com/abetterinternet/prio-server/issues/24 so I don't forget to evaluate adding these validations -- it's better to stop rotating until an operator attends to things, than to automatically rotate ourselves into a non-working state)

Issue #1112, opened 2021-11-10
Make `keyRotatorImage` in deploy-controller configurable


hostirosti:
Can we make this configurable to allow for independently build images to be used here?

https://github.com/abetterinternet/prio-server/blob/154024874f84b9fc526b5c5352b60a492dbe6917/deploy-operator/controllers/locality_controller.go#L49

BranLwyd:
The in-progress work to enable key rotation will not use the `deploy-operator`, but I will ensure that the docker image is configurable similarly to the `workflow-manager` & `facilitator`:

https://github.com/abetterinternet/prio-server/blob/154024874f84b9fc526b5c5352b60a492dbe6917/terraform/main.tf#L194-L212

BranLwyd:
This is implemented (but not yet deployed to production environments) as https://github.com/abetterinternet/prio-server/blob/f8e9707bb94471e6d9d7abcb9afda3216e652adc/terraform/main.tf#L190-L198.

PR #1115, opened 2021-11-11
Remove deploy-operator.

BranLwyd:
Our current plan is to use a cron job rather than an operator to do key-rotation work. While we may revisit the plan to use an operator in the future, for now it is dead code and it is causing at least a little confusion among other readers of this repo (https://github.com/abetterinternet/prio-server/issues/1112), so I'd like to remove it. We can retrieve if from Git history if it turns out to be necessary in the future.


tgeoghegan (approved):

tgeoghegan (changes requested):
Besides deleting the code, we should delete the Terraform stuff that creates the CRD, or any other TF stuff that deploy-operator depended on.

BranLwyd:
OK, I think I got it all. (I didn't realize it was there, but locality_kubernetes serves as a useful example -- I'll be reintroducing per-locality kubernetes config soon since that's the intended "domain" of the cron job I'm setting up.)

tgeoghegan (approved):

PR #1128, opened 2021-11-12
Add validations to key.Key type.

BranLwyd:
As part of this, I change the internal format (but not the serialization format) of key.Key, as well as switching the public key.Key API to be less error-prone.

This PR implements the following validations identified in https://github.com/abetterinternet/prio-server/issues/24:
* Keys must either have no versions, or exactly one primary version.
* Post-rotation, keys must have at least one version.
* When being used to update keys in a manifest, keys must have at least one version.[1]
* When being used to update keys in a manifest, keys must have versions with distinct creation times.

I suspect the most controversial change in this PR will be the change of CreationTime (time.Time) to CreationTimestamp (int64). While I strongly feel it is better to use an int64 for the internal representation of the creation timestamp, I could change the public API of Version to translate this back to a time.Time as needed if folks are wary of using integer timestamps instead of time.Time values.

The individual commit messages in this PR may be useful -- I tried to break out the individual changes into separate commits with useful messages.

[1] Only guaranteed because keys used as inputs to manifest updates are always the result of a rotation operation currently. I plan to add additional validation to the manifest update logic to catch this, too, in case a program logic error makes this untrue in the future.

### DIFF
--- a/key-rotator/key/key.go
+++ b/key-rotator/key/key.go
@@ -144,41 +144,46 @@ func (cfg RotationConfig) Validate() error {
 //    * If there is a key version not younger than `primary_min_age`, select
 //      the youngest such key version as primary.
 //    * Otherwise, select the oldest key version as primary.
+//
+// The returned key is guaranteed to include at least one version. If the
+// RotationConfig is such that all key versions would be deleted, an error is
+// returned.
 func (k Key) Rotate(now time.Time, cfg RotationConfig) (Key, error) {
 	// Validate parameters.
 	if err := cfg.Validate(); err != nil {
 		return Key{}, fmt.Errorf("invalid rotation config: %w", err)
 	}
 
-	// Copy the existing list of key versions, sorting by creation time (oldest
-	// to youngest). Also, validate that we aren't trying to rotate a key
-	// containing a version from the "future" to simplify later logic.
+	// Copy the existing list of key versions, sorting by creation time
+	// ascending (oldest to youngest). Also, validate that we aren't trying to
+	// rotate a key containing a version from the "future" to simplify later
+	// logic.
 	nowTS := now.Unix()
 	age := func(v Version) time.Duration { return time.Second * time.Duration(nowTS-v.CreationTimestamp) }
-	kvs := make([]Version, 0, 1+len(k.v))
+	vs := make([]Version, 0, 1+len(k.v))
 	for _, v := range k.v {
 		if age(v) < 0 {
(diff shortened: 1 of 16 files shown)


aaomidi:
Apologies for taking a while to review this. Is it okay for me to review this after the other two stacked changes are merged?

BranLwyd:
> Apologies for taking a while to review this. Is it okay for me to review this after the other two stacked changes are merged?

No worries -- I went ahead and merged the stacked PRs. (and sorry for taking so long to do so, I was assuming you'd want to review w/o merging!)

key-rotator/key/key.go
```diff
@@ -12,17 +12,51 @@ import (
...
+var _ json.Unmarshaler = &Key{}
 
 // FromVersions creates a new key comprised of the given key versions.
-func FromVersions(versions ...Version) (Key, error) {
-	vs := make([]Version, len(versions))
-	copy(vs, versions)
+func FromVersions(primaryVersion Version, otherVersions ...Version) (Key, error) {
+	vs := make([]Version, 1+len(otherVersions))
+	vs[0] = primaryVersion
+	copy(vs[1:], otherVersions)
```
aaomidi, line 30: This is cool. I've always used append and had never considered this. TIL!

key-rotator/key/key.go
```diff
@@ -98,52 +145,51 @@ func (cfg RotationConfig) Validate() error {
...
-		kvs = append(kvs, v)
+		vs = append(vs, v)
 	}
-	sort.Slice(kvs, func(i, j int) bool { return kvs[i].CreationTime.Before(kvs[j].CreationTime) })
+	sort.Slice(vs, func(i, j int) bool { return vs[i].CreationTimestamp < vs[j].CreationTimestamp })
 
 	// Policy: if no key versions exist, or if the youngest key version is
 	// older than `create_min_age`, create a new key version.
-	if len(kvs) == 0 || age(kvs[len(kvs)-1]) > cfg.CreateMinAge {
+	if len(vs) == 0 || age(vs[len(vs)-1]) > cfg.CreateMinAge {
```
aaomidi, line 177: Nit:

```
age(vs[len(vs)-1])
```

To me this is a bit hard to follow what's happening there.

aaomidi (approved):
So I like this, quite a lot. Handles a really complex problem quite well. This is probably too late to address now, and honestly probably shouldn't be addressed and we should trust the automation, but I would **love** a dry run capability to show what it's about to do with all the secrets.

key-rotator/key/key.go
BranLwyd: I added some clarifying comments, and pulled out the inner indexing expression onto another line.

PR #1135, opened 2021-11-15
Add validations to manifest's UpdateKeys function

BranLwyd:
Specifically (from https://github.com/abetterinternet/prio-server/issues/24):
* When being used to update keys in a manifest, keys must have at least one version.
* Post-update, manifests must have exactly one packet encryption key version.
* Post-update, manifests must have at least one batch signing key version.
* Post-update, manifests' non-key data must match the pre-update manifest data exactly.
* Post-update, manifests' key data for key versions that exist both pre- & post-update must match exactly.

This PR does **not** implement one final check from https://github.com/abetterinternet/prio-server/issues/24: "When being used to update keys in a manifest, existing key versions' public portions should match with the public keys written to the manifest, matching by creation time/key ID (both of which would provide the same matching)." This is easy enough to implement, but is going to break a lot of tests, so I'm going to pursue it in a separate PR.

Stacked on https://github.com/abetterinternet/prio-server/pull/1128.

### DIFF
--- a/key-rotator/manifest/manifest.go
+++ b/key-rotator/manifest/manifest.go
@@ -1,6 +1,7 @@
 package manifest
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/abetterinternet/prio-server/key-rotator/key"
@@ -56,7 +57,22 @@ type UpdateKeysConfig struct {
 	PacketEncryptionKeyCSRFQDN  string  // the FQDN to specify for packet encryption key CSRs
 }
 
+func (cfg UpdateKeysConfig) Validate() error {
+	if cfg.BatchSigningKey.IsEmpty() {
+		return errors.New("batch signing key has no key versions")
+	}
+	if cfg.PacketEncryptionKey.IsEmpty() {
+		return errors.New("packet encryption key has no key versions")
+	}
+	return nil
+}
+
 func (m DataShareProcessorSpecificManifest) UpdateKeys(cfg UpdateKeysConfig) (DataShareProcessorSpecificManifest, error) {
+	// Validate parameters.
+	if err := cfg.Validate(); err != nil {
+		return DataShareProcessorSpecificManifest{}, fmt.Errorf("invalid update config: %w", err)
+	}
+
 	// Copy the current manifest, clearing any existing batch signing/packet encryption keys.
 	newM := m
 	newM.BatchSigningPublicKeys, newM.PacketEncryptionKeyCSRs = BatchSigningPublicKeys{}, PacketEncryptionKeyCSRs{}
--- a/key-rotator/manifest/manifest.go
+++ b/key-rotator/manifest/manifest.go
@@ -137,7 +137,7 @@ func validateUpdatedManifest(m, oldM DataShareProcessorSpecificManifest) error {
 
 	// Post-update, manifests must have exactly one packet encryption key version.
 	if len(m.PacketEncryptionKeyCSRs) != 1 {
-		return fmt.Errorf("expected only one packet encryption public key (had %d)", len(m.PacketEncryptionKeyCSRs))
+		return fmt.Errorf("expected exactly one packet encryption public key (had %d)", len(m.PacketEncryptionKeyCSRs))
 	}
 
 	// Post-update, manifests' non-key data must match pre-update manifest data exactly.
--- a/key-rotator/manifest/manifest.go
+++ b/key-rotator/manifest/manifest.go
@@ -37,12 +37,16 @@ type DataShareProcessorSpecificManifest struct {
 	PacketEncryptionKeyCSRs PacketEncryptionKeyCSRs `json:"packet-encryption-keys"`
 }
 
-func (m DataShareProcessorSpecificManifest) Equal(o DataShareProcessorSpecificManifest) bool {
+func (m DataShareProcessorSpecificManifest) equalModuloKeys(o DataShareProcessorSpecificManifest) bool {
 	return m.Format == o.Format &&
 		m.IngestionIdentity == o.IngestionIdentity &&
 		m.IngestionBucket == o.IngestionBucket &&
 		m.PeerValidationIdentity == o.PeerValidationIdentity &&
-		m.PeerValidationBucket == o.PeerValidationBucket &&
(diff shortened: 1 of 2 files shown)


key-rotator/manifest/manifest.go
```diff
@@ -36,12 +37,16 @@ type DataShareProcessorSpecificManifest struct {
...
+func (m DataShareProcessorSpecificManifest) equalModuloKeys(o DataShareProcessorSpecificManifest) bool {
 	return m.Format == o.Format &&
 		m.IngestionIdentity == o.IngestionIdentity &&
 		m.IngestionBucket == o.IngestionBucket &&
 		m.PeerValidationIdentity == o.PeerValidationIdentity &&
-		m.PeerValidationBucket == o.PeerValidationBucket &&
+		m.PeerValidationBucket == o.PeerValidationBucket
+}
+
+func (m DataShareProcessorSpecificManifest) Equal(o DataShareProcessorSpecificManifest) bool {
```
aaomidi, line 48: nit: Just keep in mind that this is a pretty large struct that we're going to be copying each time we want to call Equal. Not a big deal and premature optimization at best.

key-rotator/manifest/manifest.go
```diff
@@ -56,7 +61,22 @@ type UpdateKeysConfig struct {
 	PacketEncryptionKeyCSRFQDN  string  // the FQDN to specify for packet encryption key CSRs
 }
 
+func (cfg UpdateKeysConfig) Validate() error {
+	if cfg.BatchSigningKey.IsEmpty() {
+		return errors.New("batch signing key has no key versions")
```
aaomidi, line 66: Nit: These errors are being referenced by the test. I'd make them a var so you can just directly grab them instead of having to do string matching. Also gives flexibility in case you need to do something when you see these errors when this actually runs __for realz__

key-rotator/manifest/manifest_test.go
```diff
@@ -227,6 +228,339 @@ func TestUpdateKeys(t *testing.T) {
...
+		{
+			name: "empty batch signing key",
+			cfg: UpdateKeysConfig{
+				BatchSigningKey:             key.Key{},
+				BatchSigningKeyIDPrefix:     bskPrefix,
+				PacketEncryptionKey:         k(kv(0, k0)),
+				PacketEncryptionKeyIDPrefix: pekPrefix,
+				PacketEncryptionKeyCSRFQDN:  fqdn,
+			},
+			wantErrStr: "batch signing key has no key versions",
```
aaomidi, line 247: Nit: Generally these, it seems like it'd be a bit nicer to reference them directly.

key-rotator/manifest/manifest.go
```diff
@@ -103,9 +123,48 @@ func (m DataShareProcessorSpecificManifest) UpdateKeys(cfg UpdateKeysConfig) (Da
...
+	}
+
+	// Post-update, manifests' non-key data must match pre-update manifest data exactly.
+	if !m.equalModuloKeys(oldM) {
+		return fmt.Errorf("non-key data modified")
+	}
+
+	// Post-update, manifests' key data for key versions that exist both pre- &
+	// post-update must match exactly.
+	for kid, key := range m.BatchSigningPublicKeys {
```
aaomidi, line 150: Correct me if im wrong here, but I think there should be one more check of "these `kid`s should definitely exist on the new key version and error if it doesn't (?)

aaomidi (approved):

key-rotator/manifest/manifest.go
BranLwyd: Agreed: this was added as part of the follow-on PR https://github.com/abetterinternet/prio-server/pull/1136. (I didn't put them in this PR to keep PR sizes at least somewhat manageable, and because the validations in https://github.com/abetterinternet/prio-server/pull/1136 required refactoring a large number of tests in several packages.)

key-rotator/manifest/manifest_test.go
BranLwyd: Acknowledged; see my response to the other comment.

key-rotator/manifest/manifest.go
BranLwyd: I'd argue that using string-matching in tests is the right path, for now.

These error conditions aren't part of the public API of the package. That is, we don't expose these errors in a way that would allow other packages to discriminate between the different error cases/types/kinds -- errors are currently just a human-readable descriptive string. We don't expose error kinds because we don't need to, and I don't implement functionality that I don't need for YAGNI reasons. (If things changed and we did start exposing error types as a public part of the package, i.e. because some other package needed to discriminate between the different ways that `Validate`/`UpdateKeys` can fail, it would be absolutely correct to test against the public API.)

While I don't want to expand the package's public API, I could define package-private errors and check against them via `errors.Is`. When the error string is a constant string, as on this line, this would be easy enough (`var batchSigningKeyHasNoVersions = errors.New("...")`). But several errors are not constant strings (e.g. L140, or any other error coming from `fmt.Errorf`). To test against these in a type-safe way, I'd need to implement a new error type for each error condition as a private implementation detail, then use `errors.Is` against those new types. But ... testing against a private error type is IMO no better[1] than testing against a private error string, and it is more code to implement.

WDYT?

[1] Modulo, I suppose, a private error type could reword the human-readable string returned from `Error()` without needing to update the relevant test. I'm OK with that tradeoff.

key-rotator/manifest/manifest.go
BranLwyd: Yup -- I like to pass things by-value, until there's a good reason not to (e.g. performance, necessary mutability, ...). IMO, it's somewhat less error-prone to use values rather than pointers-to-values because you can be sure passing your value to some function won't mutate it; though admittedly in this case, there are several map fields in the struct so this goal is not fully realized. (and keep in mind that what is copied for e.g. a string is the pointer-and-length, not the underlying byte array, so these copies are of 11 64-bit words per struct copy, rather than an arbitrary amount of data)

key-rotator/manifest/manifest.go
aaomidi: I think this is fine, honestly this test actually ensures that if we're breaking changes on error messages we REALLY need to be aware of it. LGTM

PR #1136, opened 2021-11-17
Add further validations to manifest's UpdateKeys function

BranLwyd:
Specifically (from https://github.com/abetterinternet/prio-server/issues/24):
* When being used to update keys in a manifest, existing key versions' public portions should match with the public keys written to the manifest, matching by creation time/key ID (both of which would provide the same matching).
* Detection for: "if a batch signing key got so far ahead of its related manifests that the primary key was not included in the manifest, we would be trying to sign data with a key that we do not recognize as valid for verification."
* Detection for: "if a packet encryption key got so far ahead of its related manifests that the key did not include the version advertised as primary in the manifest, we would be asking clients to encrypt with a key we no longer consider valid for decryption."

Stacked on https://github.com/abetterinternet/prio-server/pull/1135.

### DIFF
--- a/key-rotator/manifest/manifest.go
+++ b/key-rotator/manifest/manifest.go
@@ -1,6 +1,9 @@
 package manifest
 
 import (
+	"crypto/ecdsa"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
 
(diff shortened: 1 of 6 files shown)


aaomidi (approved):
Wow this is beautifully done.

key-rotator/manifest/manifest.go
```diff
@@ -165,6 +247,52 @@ func validateUpdatedManifest(m, oldM DataShareProcessorSpecificManifest) error {
...
+	// Verify batch signing keys.
+	if err := cfg.BatchSigningKey.Versions(func(v key.Version) error {
+		kid := cfg.batchSigningKeyID(v.CreationTimestamp)
+		bsk, ok := m.BatchSigningPublicKeys[kid]
+		if !ok {
+			return nil // key version does not exist in manifest
+		}
+		manifestPubkey, err := bsk.toPublicKey()
+		if err != nil {
+			return fmt.Errorf("couldn't parse batch signing key version %q from manifest: %v", kid, err)
```
aaomidi, line 264: Nit; was the use of %v on purpose here? They behave a __tiny__ bit differently https://pkg.go.dev/fmt#Errorf

aaomidi: (same with a bunch of other areas)

BranLwyd: Whoops, nope, not intentional. Changed `%v` to `%w` for any errors being passed to `fmt.Errorf`.

PR #1138, opened 2021-11-22
Terraform: add `key-rotator` cron job

BranLwyd:
This currently hard-codes `--dry-run`, so it will never actually rotate any keys. But it deploys & runs successfully on my dev deployment, so it works to validate the configuration modulo write ACLs.

Credit to the recently-deleted `locality_kubernetes` which my `kubernetes_locality` was closely based on.

PR #1147, opened 2021-11-23
Key Rotator: terraform to allow controlling which localities have their keys rotated

BranLwyd:
Also, fixup a few small issues:
  * Fix secret definitions so they don't try to modify _any_ secret
    data, rather than only leaving secret data at "secret_key" alone.
    (this is necessary because the new key rotation infra will also set
    the "key_versions" secret data)
  * Tweak key rotator AWS permissions to actually grant the necessary
    permissions on the Manifest bucket's objects.

Stacked on https://github.com/abetterinternet/prio-server/pull/1138.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -249,6 +249,11 @@ variable "packet_encryption_key_rotation_policy" {
   }
 }
 
+variable "enable_key_rotation_localities" {
+  type    = list(string)
+  default = []
+}
+
 terraform {
   backend "gcs" {}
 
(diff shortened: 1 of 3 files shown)


BranLwyd:
(with this PR, I was able to actually perform a key rotation in my dev environment. it looks like the facilitator is having trouble with the new key format, but I think I can investigate that orthogonally to sending this for review.)

aaomidi (approved):

PR #1148, opened 2021-11-24
Key rotator: fix bug with packet encryption key deserialization.

BranLwyd:
Previously, we weren't checking the number of bytes returned from
`Decode`; in practice, this meant packet encryption keys that
round-tripped through serialization would end up with a `D` value that
was a few bytes too long. This didn't actually break anything for the
key rotator itself, since it doesn't use the key, but as one might
imagine the facilitator was none-too-happy about receiving these
malformed keys.

Sadly, this wasn't caught by tests... because the test that would have
caught it was set up to use a "bad" key. :cry:


aaomidi:
Oh my so that's what was happening. I had ran into this exact same issue and I don't even remember what the solution ended up being. Good find on the problem.

aaomidi (approved):

PR #1151, opened 2021-11-24
Add (bogus) expiry date to manifests' batch signing public key.

BranLwyd:
This field is evidently not respected by any parties in the current ENPA
deployment (most/all of the expiration dates in the production
deployment are now in the past). But per discussion with ENPA team
members, we want to continue populating this with *something* in case
one of the parties is parsing this value & then ignoring it.

I also switch to specifying the dates as RFC 3339 rather than ISO 8601.
The populated data may very well fit both specs, but our current tooling
writes in RFC3339 format, so update the comment to be sure:
https://github.com/abetterinternet/prio-server/blob/f8e9707bb94471e6d9d7abcb9afda3216e652adc/deploy-tool/main.go#L186-L195

### DIFF
--- a/key-rotator/manifest/manifest.go
+++ b/key-rotator/manifest/manifest.go
@@ -357,7 +357,7 @@ type BatchSigningPublicKey struct {
 	// PublicKey is the PEM armored base64 encoding of the ASN.1 encoding of the
 	// PKIX SubjectPublicKeyInfo structure. It must be an ECDSA P256 key.
 	PublicKey string `json:"public-key"`
-	// Expiration is the RFC3339-encoded UTC date at which this key expires.
+	// Expiration is the ISO 8601 encoded UTC date at which this key expires.
 	Expiration string `json:"expiration"`
 }
 
--- a/key-rotator/manifest/manifest.go
+++ b/key-rotator/manifest/manifest.go
@@ -6,6 +6,7 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"time"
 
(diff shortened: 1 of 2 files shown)


key-rotator/manifest/manifest.go
```diff
@@ -352,7 +357,7 @@ type BatchSigningPublicKey struct {
 	// PublicKey is the PEM armored base64 encoding of the ASN.1 encoding of the
 	// PKIX SubjectPublicKeyInfo structure. It must be an ECDSA P256 key.
 	PublicKey string `json:"public-key"`
-	// Expiration is the ISO 8601 encoded UTC date at which this key expires.
+	// Expiration is the RFC3339-encoded UTC date at which this key expires.
```
tgeoghegan, line 360: hair-splitting-nit: [in the design doc](https://docs.google.com/document/d/1MdfM3QT63ISU70l63bwzTrxr93Z7Tv7EDjLfammzo6Q/edit#bookmark=kix.yx44j8g5efpv), we do define these dates as ISO 8601, which I chose because [W3C says it's the "right" JSON date format](https://www.w3docs.com/snippets/javascript/the-right-json-date-format.html). AFAICT, Go's `time.RFC3339` is a valid ISO 8601 date, since it doesn't use a signed offset, which [Wikipedia says is the one way RFC 3339 deviates from ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#RFCs).

So I think documenting that we use ISO 8601 is most correct, because that's the contract with other servers. That we use Go's `time.RFC3339` is an implementation detail.

tgeoghegan (approved):

PR #1152, opened 2021-11-25
Key rotator: add Key.Diff, use it to describe why we are writing keys.

BranLwyd:
I think this is useful for debugging/observability. (I plan to add
similar functionality to manifests in an upcoming PR.)

### DIFF
--- a/key-rotator/key/key.go
+++ b/key-rotator/key/key.go
@@ -79,12 +79,12 @@ func (k Key) Equal(o Key) bool {
 // string if and only if the two keys are equal.
 func (k Key) Diff(o Key) string {
 	// Build up structures allowing easy generation of diffs.
-	var newPKTS, oldPKTS *int64
+	var newPrimaryKeyTS, oldPrimaryKeyTS *int64
 	infos := map[int64]struct{ oldMat, newMat *Material }{}
 	for i, v := range k.v {
 		v := v
 		if i == 0 {
-			newPKTS = &v.CreationTimestamp
+			newPrimaryKeyTS = &v.CreationTimestamp
 		}
 		info := infos[v.CreationTimestamp]
 		info.newMat = &v.KeyMaterial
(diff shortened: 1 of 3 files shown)


key-rotator/key/key.go
```diff
@@ -73,6 +74,65 @@ func (k Key) Equal(o Key) bool {
 	return true
 }
 
+// Diff returns a human-readable string describing the differences from the
+// given `o` key to this key, suitable for logging. Diff returns the empty
+// string if and only if the two keys are equal.
+func (k Key) Diff(o Key) string {
+	// Build up structures allowing easy generation of diffs.
+	var newPKTS, oldPKTS *int64
```
tgeoghegan, line 82: It took me a minute but I eventually figured out that `PKTS` expands to `PrimaryKeyTimeStamp`. Since the type is `int64` and not something like `time.Time` that makes it more obvious, and since `PK` could also easily mean "public key" or "private key", could we expand the acronym for readability?

key-rotator/key/key.go
```diff
@@ -73,6 +74,65 @@ func (k Key) Equal(o Key) bool {
...
+	for i, v := range o.v {
+		v := v
+		if i == 0 {
+			oldPKTS = &v.CreationTimestamp
+		}
+		info := infos[v.CreationTimestamp]
+		info.oldMat = &v.KeyMaterial
+		infos[v.CreationTimestamp] = info
+	}
+	tss := make([]int64, 0, len(infos))
```
tgeoghegan, line 102: nit: I think this is more readable as `timestamps` than `tss`

tgeoghegan (approved):

key-rotator/main.go
```diff
@@ -336,8 +336,10 @@ func writeKeys(ctx context.Context, keyStore storage.Key, locality string,
 	// Write packet encryption key.
 	eg.Go(func() error {
 		if oldPacketEncryptionKey.Equal(newPacketEncryptionKey) {
+			log.Debug().Msgf("Skipping write for packet encryption key for %q: key unchanged", locality)
```
tgeoghegan, line 339: I think instead of putting the locality into the log message, you could instead make it a field in the structured log message (I think with `log.Debug().String("locality", locality).Msg(...)` or thereabouts), and then it will be easier to search for messages.

tgeoghegan: Same observation applies to the log statements on lines 342, 354, 357.

tgeoghegan (changes requested):

key-rotator/main.go
BranLwyd: I include the relevant information in both the structured and unstructured messages, since the default log view for GKE will not show the structured data at *all* -- all you get is the content of `Msgf` (though thankfully the GKE viewer does include a link to the normal GCP log viewer that does include structured data).

tgeoghegan (approved):

PR #1153, opened 2021-11-26
Key rotator: describe diffs that cause manifest writes.

BranLwyd:
I think this is useful for debugging/observability.


key-rotator/main.go
```diff
@@ -370,11 +370,12 @@ func writeManifests(
 		ingestor, oldManifest, newManifest := ingestor, oldManifest, newManifestByIngestor[ingestor]
 		eg.Go(func() error {
 			if oldManifest.Equal(newManifest) {
+				log.Debug().Msgf("Skipping write for manifest for (%q, %q): key unchanged", locality, ingestor)
```
tgeoghegan, line 373: As in #1152, could you put locality and ingestor in separate string fields on the structured message for ease of searching?

key-rotator/main.go
```diff
@@ -370,11 +370,12 @@ func writeManifests(
...
 				return nil
 			}
-			dspName := dspName(locality, ingestor)
-			if err := manifestStore.PutDataShareProcessorSpecificManifest(ctx, dspName, newManifest); err != nil {
-				return fmt.Errorf("couldn't write manifest for %q: %w", dspName, err)
+			log.Info().Msgf("Writing manifest for (%q, %q): %s", locality, ingestor, newManifest.Diff(oldManifest))
```
tgeoghegan, line 376: Move `locality` and `ingestor` to structured log fields.

tgeoghegan (changes requested):

key-rotator/main.go
BranLwyd: Done -- as in that PR, I also keep `locality`/`ingestor` in the unstructured log message, because at least some interfaces do not surface the structured portion of the log.

tgeoghegan (approved):

PR #1154, opened 2021-11-29
Key rotator: add basic metrics.

BranLwyd:
This should be enough to provide observatbility around key rotator
success/failure & when keys/manifests are actually written.


key-rotator/main.go
```diff
@@ -44,16 +47,42 @@ var (
...
-	kubeconfig = flag.String("kubeconfig", "", "The `path` to user's kubeconfig file; if unspecified, assumed to be running in-cluster") // typical value is $HOME/.kube/config
+	dryRun      = flag.Bool("dry-run", true, "If set, do not actually write any keys or manifests back (only report what would have changed)")
+	timeout     = flag.Duration("timeout", 10*time.Minute, "The `deadline` before key-rotator terminates. Set to 0 to disable timeout")
+	awsRegion   = flag.String("aws-region", "", "If specified, the AWS `region` to use for manifest storage")
+	pushGateway = flag.String("push-gateway", "", "Set this to the gateway to use with prometheus. If left empty, metrics will not be pushed to prometheus.")
+	kubeconfig  = flag.String("kubeconfig", "", "The `path` to user's kubeconfig file; if unspecified, assumed to be running in-cluster") // typical value is $HOME/.kube/config
+
+	// Metrics.
+	pusher      *push.Pusher // populated only if --push-gateway is specified.
+	keysWritten = promauto.NewCounter(prometheus.CounterOpts{
```
tgeoghegan, line 58: Unfortunately, using a `Counter` with the Prometheus `pushgateway` isn't a great idea (see [the README](https://github.com/prometheus/pushgateway/blob/master/README.md#non-goals) which explains that pushgateway is not a distributed counter). The reason is that the in-memory counter value will get reset to 0 every time the cronjob runs, and then when that value gets pushed to the pushgateway, it will override whatever counter value was there before, which frustrates a `Counter`'s expectation of [always increasing monotonically](https://prometheus.io/docs/concepts/metric_types/#counter). A `Gauge` would be more appropriate.

tgeoghegan (changes requested):

BranLwyd:
I also added a commit configuring the key rotator to specify `--push-gateway`; it looks good in my dev env, at least in the sense that the metrics are getting exported & are visible in grafana.

tgeoghegan (approved):

PR #1158, opened 2021-11-29
Docker CI fixes

tgeoghegan:
This fixes some issues with Docker image CI actions that I noticed while catching up on recent changes. See the commit messages for details.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -92,6 +92,15 @@ jobs:
       with:
         file: ./facilitator/Dockerfile
 
+  workflow-manager-docker:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Docker build
+      uses: docker/setup-buildx-action@v1
+    - name: build
+      run: ./workflow-manager/build.sh
+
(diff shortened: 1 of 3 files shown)


BranLwyd:
Thanks for the key-rotator fix -- it was on my TODO list to take a look but you got there first.

I think the workflow-manager error is probably because the context of the docker build is the root of the repo (`.`), while the Dockerfile currently expects to have its context be the workflow manager directory (`./workflow-manager/`). You **may** be able to fix this by specifying the `context` parameter to the CI build step: https://github.com/docker/build-push-action#inputs. (haven't tested)

If not, the Dockerfile can be updated for the newly-expected context. i.e. the COPY commands need to copy `workflow-manager/go.{mod,sum}` rather than `go.{mod,sum}`. (probably other changes need to be made, too, to both the `Dockerfile` and `build.sh`) A good way to test this is from the root of the `prio-server` repo, running `docker build -f workflow-manager/Dockerfile .`.

tgeoghegan:
I opted instead to invoke `workflow-manager/build.sh`, which better exercises what happens when we release container images from `main` or a release tag.

BranLwyd (approved):

PR #1166, opened 2021-11-30
Remove task-replayer.

BranLwyd:
This utility read a task (i.e. an SQS/Pubsub message, representing a unit of intake or aggregation work) from a file on disk, then enqueued it into a specified SQS/Pubsub topic.

I think we can do the same by using the [AWS CLI](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/sqs/send-message.html) or the [GCP CLI](https://cloud.google.com/pubsub/docs/publisher#gcloud). Since we can use other (not-maintained-by-us) tools to perform the work that task-replayer performs, and task-replayer is causing [minor confusion for dependabot](https://github.com/abetterinternet/prio-server/runs/4360243961?check_suite_focus=true), just remove it.


tgeoghegan (approved):
You're right, this is a maintenance burden, and we've used `workflow-manager` to reschedule tasks the last few times anyway.

PR #1171, opened 2021-12-01
prod-intl: replace `mn-mn` with `mn`

### LINKED ISSUE #1080: Redeploy Mongolia as `mn`
Per discussion in Slack, we want to use ISO-3166-1 country code for national deployments, e.g., `mn` for Mongolia instead of `mn-mn`. The locality identifiers are opaque to prio-server, so all this means is that we need to tear down the `mn-mn` infrastructure and deploy plain old `mn` instead.

tgeoghegan:
Re-deploy Mongolia with correct ISO code for nationwide deployment.

Resolves #1080

### DIFF
--- a/terraform/variables/prod-intl.tfvars
+++ b/terraform/variables/prod-intl.tfvars
@@ -1,14 +1,14 @@
 environment     = "prod-intl"
 gcp_region      = "us-west1"
 gcp_project     = "prio-intl-prod"
-localities      = ["mn-mn", "mx-coa", "mx-jal", "mx-pue", "mx-yuc"]
+localities      = ["mn", "mx-coa", "mx-jal", "mx-pue", "mx-yuc"]
 aws_region      = "us-west-2"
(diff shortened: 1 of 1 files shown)


terraform/variables/prod-intl.tfvars
```diff
@@ -57,9 +57,9 @@ ingestors = {
   }
 }
 cluster_settings = {
-  initial_node_count = 3
-  min_node_count     = 3
-  max_node_count     = 3
+  initial_node_count = 9
```
BranLwyd, line 60: I'm fine with this change, just calling it out in case it was not intended to be included in the PR.

BranLwyd (approved):

terraform/variables/prod-intl.tfvars
tgeoghegan: It was intended and necessary so that we won't shrink the cluster back to 3 nodes when we deploy this change.

PR #1173, opened 2021-12-02
Use correct key ID for batch signing private key.

BranLwyd:
Previously, the facilitator would always write the key identifier for
the "original" (e.g. old-style) key. This commit creates the necessary
plumbing for the correct key identifier to make it into the facilitator,
while still allowing fallback for "old-style" keys that have not yet
been rotated.

I'm still experimenting with/testing this, but I think it's ready for
review.


tgeoghegan (approved):

BranLwyd:
My dev environment finally processed through all the "bad" batches, so I can confirm that this clears up the error -- key IDs in generated batches now matches the key used to sign the batch.

PR #1174, opened 2021-12-05
Key storage: add implementations writing to AWS & GCP.

BranLwyd:
These aren't yet wired into the main program logic; but I need to do some research/thinking on the best way to authorize the rotator to write secrets into its backup GCP projects/AWS accounts, so this feels like a good place to cut a PR.


key-rotator/storage/key_aws.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/abetterinternet/prio-server/key-rotator/key"
+)
+
+// NewAWSKey returns a Key implementation using the AWS secret manager for
+// backing storage. This key store writes keys in a way that is suitable for
+// backup; keys written by this store cannot be read by other components of the
+// Prio system (e.g. the facilitator).
+func NewAWSKey(sm *secretsmanager.SecretsManager, prioEnv string) Key { return awsKey{sm, prioEnv} }
+
+type awsKey struct {
```
tgeoghegan, line 23: Should there be a `var _ awsKey = Key` to verify that this satisfies the `Key` interface?

tgeoghegan (changes requested):
This set of changes handles backing up a key at the time of creation, but I think we will also need to be able to back up existing keys (e.g. in prod intl where we have five, unbackedup localities deployed). That may be in a different PR though.

BranLwyd:
I think for backing up existing keys, we want something like an `--always-write-keys` that writes keys even if no changes are detected. Then we just run one instance of the key rotator with that flag specified. I'll send a PR shortly.

tgeoghegan (approved):

PR #1186, opened 2021-12-06
Update handling of serialized private key values.

BranLwyd:
Previously, they might be serialized as less than 32 bytes if the most-significant bytes of the key turned out to be zero. This would cause problems, AFAICT, in both key-rotator parsing logic as well as facilitator key-handling logic.

### DIFF
--- a/deploy-tool/main.go
+++ b/deploy-tool/main.go
@@ -87,8 +87,19 @@ type privateKeyMarshaler func(*ecdsa.PrivateKey) ([]byte, error)
 // expected by libprio-rs encrypt::PrivateKey, which is the X9.62 uncompressed
 // public key concatenated with the secret scalar.
 func marshalX962UncompressedPrivateKey(ecdsaKey *ecdsa.PrivateKey) ([]byte, error) {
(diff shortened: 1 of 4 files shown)


BranLwyd:
To close the loop, it looks like `ecdsa.GenerateKey` can generate these "short" private keys (whose most-significant byte is 0). I ran the following program for a few seconds and just under 1/256 of the keys generated had "short" private keys:

```golang
package main

import (
	"crypto/ecdsa"
	"crypto/elliptic"
	"crypto/rand"
	"fmt"
)

func main() {
	var keys, shortKeys uint64
	for {
		k, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
		if err != nil {
			panic(fmt.Sprintf("Couldn't generate key: %v", err))
		}

		keys++
		if len(k.D.Bytes()) < 32 {
			shortKeys++
		}
		if keys%10000 == 0 {
			fmt.Printf("Generated %d keys, got %d short private keys\n", keys, shortKeys)
		}
	}
}
```

tgeoghegan (approved):

PR #1187, opened 2021-12-06
Enable key rotator in staging-facil environment.

BranLwyd:
I set it up at a rapid pace: the rotator runs once per ten minutes
rather than once per day, and will allow a new batch signing key version
to become primary after 6 hours rather than a week.

Haven't applied yet, but similar changes work for my dev environment. Will apply once this passes review.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -249,6 +249,11 @@ variable "packet_encryption_key_rotation_policy" {
   }
 }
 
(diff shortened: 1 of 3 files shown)


tgeoghegan (approved):

tgeoghegan:
This LGTM. However, you'll run into trouble because `staging-pha` won't be updated to periodically check for new keys.

PR #1191, opened 2021-12-08
terraform: configure cluster-autoscaler in EKS

tgeoghegan:
In order to autoscale an EKS cluster (i.e., the worker nodes), one must
configure and deploy a `cluster-autoscaler` deployment. This commit does
that in Terraform, adapting the instructions from Amazon[1].

[1]: https://docs.aws.amazon.com/eks/latest/userguide/autoscaling.html#cluster-autoscaler

### DIFF
--- a/terraform/modules/account_mapping/account_mapping.tf
+++ b/terraform/modules/account_mapping/account_mapping.tf
@@ -28,6 +28,11 @@ variable "kubernetes_service_account_name" {
   type = string
 }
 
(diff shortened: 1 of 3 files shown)


tgeoghegan:
I forcibly scaled my dev EKS cluster up to 6 nodes and then deployed this, and observed it scaling back down to 3 nodes over about an hour.

BranLwyd (approved):

PR #1195, opened 2021-12-09
terraform: configure on demand node group for EKS

### LINKED ISSUE #1192: EKS: guarantee nodes in all availability zones
On December 8, 2021, we found that our 9 node Kubernetes cluster suddenly had all its nodes scheduled in `us-west-2c`, in spite of the EKS node group and underlying EC2 autoscaling group being configured to allocate nodes across `us-west-2{a,b,c}`. Based on the ASG activity logs, it seems this is because EC2 couldn't satisfy our spot instance requests, perhaps because of unusually high demand for VMs in `us-west-2`.

This is bad in general: we want capacity in multiple AZs for resilience. More particularly, it makes our `prometheus-server` deployment unschedulable, because it needs a persistent volume backed by an EBS volume that is in `us-west-2a`. Thus we want to guarantee that our Kubernetes cluster has at least one node in each configured AZ.

tgeoghegan:
This is a pretty hacky fix for the issue encountered in #1192, where
we couldn't allocate spot instances in AWS and found ourselves with 9
nodes in a single AZ, leaving `prometheus-server` unschedulable because
it happened to not be the AZ where its EBS volume is.

To guarantee that we will have at least one node in each AZ, we
configure two node groups: one with cheaper but less reliable spot
instances that can be autoscaled by the `cluster-autoscaler` and another
with what AWS calls "on demand" instances, containing as many nodes as
we have subnets in the cluster (and hence AZs).

Resolves #1192


tgeoghegan:
I'm not a huge fan of this solution because I worry that there might be unintended consequences to using multiple node groups. The other most straightforward solution would be to change the existing node group to use `capacity_type = "ON_DEMAND"`, but the downside there is that we'd be paying more for the same VMs.

BranLwyd (approved):

PR #1216, opened 2021-12-14
Facilitator: add `PrioValidityBatch` message with serialization support

BranLwyd:
`PrioValidityBatch` is comprised of a `PrioBatchSignature`, a `PrioValidityHeader`, and a sequence of `PrioValidityPacket`s (i.e. the three messages comprising a validation batch). The latter two messages are included as serialized bytes to allow signature verification, and then hash integrity checking, to be done over the same bytes that the signature/hash was generated against.

I also added some read tests that attempt to read "golden" serialized data and match it against what we expect it to deserialize to; this is to protect against the possibility of accidentally-backwards-incompatible changes (i.e. protecting against a change that still allows roundtripping through serialization/deserialization, but does not allow deserialization of data serialized with a previous version of the code). The golden serialized data was generated from the code as it stood before I started making the changes in this PR.

### DIFF
--- a/facilitator/src/idl.rs
+++ b/facilitator/src/idl.rs
@@ -41,7 +41,7 @@ pub trait Packet: Sized {
     /// that unlike other structures, this does not take a primitive
     /// std::io::Read, because we do not want to create a new Avro schema and
     /// reader for each packet. The Reader must have been created with the
-    //// schema returned from Packet::schema.
+    /// schema returned from Packet::schema.
     fn read<R: Read>(reader: &mut Reader<R>) -> Result<Self, Error>;
 
     /// Serializes and writes a single Packet to the provided avro_rs::Writer.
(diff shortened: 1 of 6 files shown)


facilitator/src/idl.rs
```diff
@@ -1087,35 +1094,139 @@ impl Packet for InvalidPacket {
...
 
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct ValidationBatch {
+    pub sig: BatchSignature,
+    pub header: Vec<u8>,
+    pub packets: Vec<u8>,
+}
+
+impl ValidationBatch {
+    /// Reads and parses one Header from the provided std::io::Read instance.
```
tgeoghegan, line 1105:
```suggestion
    /// Reads and parses one ValidationBatch from the provided std::io::Read instance.
```

facilitator/src/idl.rs
```diff
@@ -1087,35 +1094,139 @@ impl Packet for InvalidPacket {
...
+    pub fn read<R: Read>(reader: R) -> Result<Self, Error> {
+        let mut reader = Reader::with_schema(*VALIDATION_BATCH_SCHEMA, reader)
+            .map_err(|e| Error::AvroError("failed to create Avro reader".to_owned(), e))?;
+        // We expect exactly one record and for it to be an ingestion signature
```
tgeoghegan, line 1109:
```suggestion
        // We expect exactly one record and for it to be a validation batch
```

facilitator/src/idl.rs
```diff
@@ -1087,35 +1094,139 @@ impl Packet for InvalidPacket {
...
+        let mut sig = None;
+        let mut header = None;
+        let mut packets = None;
+
+        for (field_name, field_value) in fields {
+            match (field_name.as_str(), field_value) {
+                ("sig", field_value) => sig = Some(BatchSignature::try_from(field_value)?),
+                ("header", Value::Bytes(buf)) => header = Some(buf),
+                ("packets", Value::Bytes(buf)) => packets = Some(buf),
+                _ => (),
```
tgeoghegan, line 1165: I think we should error out if we encounter any unexpected fields.

tgeoghegan (changes requested):

facilitator/src/idl.rs
```diff
@@ -112,65 +162,38 @@ impl BatchSignature {
         let mut batch_header_signature = None;
         let mut key_identifier = None;
 
-        for tuple in record {
-            match (tuple.0.as_str(), tuple.1) {
+        for (field_name, field_value) in fields {
+            match (field_name.as_str(), field_value) {
                 ("batch_header_signature", Value::Bytes(v)) => batch_header_signature = Some(v),
                 ("key_identifier", Value::String(v)) => key_identifier = Some(v),
-                (f, _) => {
```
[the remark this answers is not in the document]
BranLwyd: Hmm... I made this change purposefully. Otherwise, we can't introduce a new field without it being a breaking change, i.e. we'd have to update every facilitator that might read this message type simultaneously. (or introduce the field as a two-release process: in the first release, allow reading the new field; in the second release, start writing the new field. that's a lot of pain just for a new field!)

I should've called this out beforehand as a purposeful behavior change (apologies)--if my argument is not convincing, I'll revert this part of the change.

facilitator/src/idl.rs
BranLwyd: See response to the other instance of this comment--I think this is what we want, but I'll revert if you disagree even w/ my reasoning.

facilitator/src/idl.rs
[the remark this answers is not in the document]
BranLwyd: (not that this should be taken as iron-clad evidence, but "allowing but ignoring unknown fields" is how e.g. protobuf is implemented, for pretty much the reason I described above)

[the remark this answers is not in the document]
tgeoghegan: No, this is fine. TBH I don't expect these Avro messages to ever change anyway.

tgeoghegan (approved):

Issue #1196, opened 2021-12-09
Update Grafana Helm chart


tgeoghegan:
There is a [spicy Grafana CVE](https://grafana.com/blog/2021/12/08/an-update-on-0day-cve-2021-43798-grafana-directory-traversal/). It's not a big concern for prio-server because none of our Grafanas are publicly accessible, but we should still update the deployed Grafana Helm chart to the latest since it's quite easy to do.

tgeoghegan:
We should update the other Helm charts (Prometheus, metrics adapters) while we're at it.

tgeoghegan:
Updating the Prometheus chart past 15.0 requires [one manual step](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus#to-150). We just have to do `kubectl -n monitoring delete deployments/prometheus-kube-state-metrics` before `make apply` in each env. That specific deployment is stateless and can be recreated by deploying the new Helm chart.

tgeoghegan:
Deployed to `staging-facil` and `staging-intl` earlier this afternoon. We will deploy to production probably next week.

tgeoghegan:
This was deployed to prod-us yesterday but is not yet in prod-intl.

BranLwyd:
This was deployed to `prod-intl` yesterday afternoon, closing.

Issue #1169, opened 2021-12-01
Enable worker autoscaling in production environments


BranLwyd:
Currently, the ability to auto-scale workers based on them falling behind on work is implemented, but the default configuration is that the minimum & maximum number of workers is the same, so no autoscaling actually occurs.

We should update both ISRG & NIH deployments to allow worker autoscaling to actually occur.

tgeoghegan:
Out of an abundance of caution, we should enable this in staging environments first (crucially `staging-facil` on GCP and `staging-intl` on AWS) and roll it out to prod once we see it working well in staging.

tgeoghegan:
Deployed to `staging-facil` and `staging-pha` in #1194.

tgeoghegan:
This got deployed to prod-us yesterday, but is not yet in prod-intl, so reopening.

BranLwyd:
Deployed to `prod-intl` as of yesterday afternoon; closing.

PR #1219, opened 2021-12-15
Deploy #1196, #1169 to production

tgeoghegan:
Catch production up with changes recently deployed in staging. See individual commit messages for details.


terraform/variables/prod-us.tfvars
```diff
@@ -86,66 +116,96 @@ ingestors = {
...
+        min_intake_worker_count    = 1
+        max_intake_worker_count    = 6
+        min_aggregate_worker_count = 1
+        max_aggregate_worker_count = 3
       }
       us-ca = {
-        intake_worker_count    = 15
-        aggregate_worker_count = 3
+        min_intake_worker_count    = 1
+        max_intake_worker_count    = 10
```
BranLwyd, line 152: nit: I'm almost certain this is intentional, but this shrinks the maximum number of intake workers for us-ca from 15 to 10. I'm OK with it if you are (I suspect you know better than me what numbers are reasonable here), just want to make sure it's intentional.

BranLwyd: (and same observation for the Apple us-ca deployment)

BranLwyd (approved):

terraform/variables/prod-us.tfvars
tgeoghegan: Yes: my thinking here is that we have been overprovisioned in CA for a while. Fewer than 25 or 15 workers for Apple or Google respectively should be enough to stay on top of the backlog. If it's not, we can easily crank up the `max_intake_worker_count`.

PR #1220, opened 2021-12-16
Key rotator: allow automatic turnup of new localities.

BranLwyd:
This change allows the `key-rotator` to automatically handle provisioning new keys, without the need for an operator to manually run `deploy-tool`. Once this is deployed to production, I think we can delete `deploy-tool` entirely.

The main change in this PR is to wire the `key-rotator` to have "manifest templates" so that it knows all the right non-key data to include in a new manifest, and update the logic to use the templates if no manifest has been written to the appropriate cloud storage bucket.

I also had to make a small change to key-rotation: we decided to disable rotation of the BSK, but we still need to "rotate" the initial empty key to provision a first version for new localities. I think this is small enough to sneak into this PR.

Other than unit tests, I deployed a new locality to my dev environment and observed that the key-rotator generated keys/manifests appropriately. As of writing I am still waiting for data to flow all the way through the system, but I think this is ready for review.

See individual commit messages for further details.

### DIFF
--- a/key-rotator/main.go
+++ b/key-rotator/main.go
@@ -40,14 +40,14 @@ var (
 	csrFQDN           = flag.String("csr-fqdn", "", "Required. FQDN to use as common name in generated CSRs")
 
 	// Rotation configuration.
-	batchSigningKeyEnableRotation = flag.Bool("batch-signing-key-enable-rotation", true, "Determines if batch signing keys are rotated")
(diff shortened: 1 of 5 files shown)


BranLwyd:
My dev environment is outputting sum parts for my new locality; I think that's about as good of an end-to-end test of automated key provisioning as I can do pre-merge.

tgeoghegan (approved):
LGTM; thanks for breaking this up into digestible commits.

PR #1221, opened 2021-12-16
GKE `facilitator` debugging

tgeoghegan:
Configures firewall rules to allow SSH to GKE worker nodes. I don't yet know how to do this for EKS. Second commit adds a document with basic instructions on SSHing into GKE nodes as well as obtaining and symbolicating `facilitator` coredumps.

### DIFF
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -1,0 +1,64 @@
+# Debugging
+
+## SSH into GKE worker node
+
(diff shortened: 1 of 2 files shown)


BranLwyd (approved):
Thanks for the docs! I didn't try out the steps, but they read well/are easy to follow.

PR #1238, opened 2021-12-20
Specify flags for facilitator-based jobs directly.

BranLwyd:
The previous configmap-based indirection would not trigger a job restart
on configmap change. This has led to a few instances (in preproduction
environments, thankfully) where the operator believed the facilitator
was running with different config than it was actually running.

Instead, now we specify flags directly in the deployment. This will
trigger terraform to cause an update whenever the configuration changes.
This also happens to cause the facilitator to log its config (at least, the
parts that were moved to flags), since it already logs its arguments on
startup.

Tested: this is running in my dev environment; jobs came up without complaint.

### DIFF
--- a/terraform/modules/kubernetes/kubernetes.tf
+++ b/terraform/modules/kubernetes/kubernetes.tf
@@ -341,6 +341,7 @@ resource "kubernetes_deployment" "intake_batch" {
           name  = "facile-container"
           image = "${var.container_registry}/${var.facilitator_image}:${var.facilitator_version}"
           args = [
+            "--pushgateway=${var.pushgateway}",
             "intake-batch-worker",
             "--is-first=${var.is_first ? "true" : "false"}",
             "--batch-signing-private-key-default-identifier=${kubernetes_secret.batch_signing_key.metadata[0].name}",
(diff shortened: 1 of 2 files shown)


terraform/modules/kubernetes/kubernetes.tf
```diff
@@ -408,7 +340,24 @@ resource "kubernetes_deployment" "intake_batch" {
         container {
           name  = "facile-container"
           image = "${var.container_registry}/${var.facilitator_image}:${var.facilitator_version}"
-          args  = ["intake-batch-worker"]
+          args = [
+            "--pushgateway=${var.pushgateway}",
```
tgeoghegan, line 344: `facilitator` doesn't use the Prometheus pushgateway. I think this has just been there since before `facilitator` ran as a persistent deployment pulling tasks from queues (in the long ago, the dark ages, just about a year ago). We should be able to just remove this argument. I notice that `facilitator/src/bin/facilitator.rs` still has this argument defined, which is a shame, because it doesn't do anything. I'd say we should remove that, too, but I'm worried it might break existing deployments if `clap` rejects unexpected arguments.

tgeoghegan (changes requested):

terraform/modules/kubernetes/kubernetes.tf
BranLwyd: I think it would almost certainly be OK removing the flag entirely, since it has only been specified as an environment variable in the past. The one wrinkle is the Google deployment, where I'm not sure if flags are specified as flags or environment variables.

For now I deprecate, once this change has gone out long enough I suggest we remove. Perhaps by then I'll have peeked into Google's configuration & will know how they specify flags.

tgeoghegan (approved):

PR #1241, opened 2021-12-22
During aggregation, skip batches with undecryptable data share packets.

BranLwyd:
Previously, if we received an undecryptable packet, we would fail out
the entire aggregation. Now, we silently skip ingestion batches that
contain undecryptable packets; our peer will also skip such ingestion
batches because we won't have generated a peer validation batch.

This avoids the unfortunate scenario of a single bad client packet
causing an entire aggregation batch to be dropped; instead, effectively
only the ingestion batch is dropped. This restores the behavior from
before we stopped writing own-validation batches.

### DIFF
--- a/facilitator/src/aggregation.rs
+++ b/facilitator/src/aggregation.rs
@@ -16,7 +16,7 @@ use prio::{
     field::FieldPriov2,
     server::{Server, VerificationMessage},
 };
-use slog::{info, o, Logger};
+use slog::{info, o, warn, Logger};
 use std::{
     collections::{HashMap, HashSet},
     convert::TryFrom,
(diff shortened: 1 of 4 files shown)


facilitator/src/aggregation.rs
```diff
@@ -289,44 +289,64 @@ impl<'a> BatchAggregator<'a> {
...
+                Ok(p) => ingestion_and_own_validation_packets.push((ingestion_packet, p)),
+                Err(e) => {
+                    if let Some(Error::PacketDecryptionError(packet_uuid)) = e.downcast_ref() {
+                        // This ingestion batch contains a packet that can't be
+                        // decrypted. Ignore the entire ingestion batch, on the
+                        // assumption that our peer will do the same, since we
+                        // also wouldn't have been able to produce a peer
+                        // validation batch from this ingestion batch.
+                        warn!(self.logger, "Skipping ingestion batch due to undecryptable packet";
+                            event::PACKET_UUID => packet_uuid.to_string());
```
tgeoghegan, line 314: Besides the log, we could emit a Prometheus metric for undecryptable packets, and maybe put an alert on it someday. I can live without it, though. We'll at least be able to search logs to tell that this has happened.

tgeoghegan (approved):

facilitator/src/aggregation.rs
BranLwyd: For now, I'll leave it be -- there are a few other conditions that are warning-only right now.

BranLwyd:
Note to self/FTR: letting this bake for a while in my dev environment before I commit.

Issue #1170, opened 2021-12-01
Restrict operator privileges


tgeoghegan:
Our opsec could be improved considerably. Specifically, we should be deploying to production from some secure and ideally automated environment like a Github Action, and operators should have to opt-into dangerous cloud platform credentials by assuming an AWS IAM role or a GCP service account (i.e., our human users should have readonly privileges at best).

tgeoghegan:
I have been tinkering with this in EKS/AWS and have concocted a Scheme.

This ticket is about a few different things. What I've worked out this afternoon is a way to have operators use read-only credentials for AWS accounts and EKS clusters by default, and use separate credentials and authentication contexts to assume privileges that allow modification or destruction of resources. This scheme should also address #1031. It won't address #1165, but it does introduce an IAM indirection that should make it easier to do.

What this scheme does _not_ do is eliminate insider threats. The aim here is to minimize the risk of _accidental_ damage to deployed resources by requiring one extra hurdle to be leaped before dangerous credentials are available in an operator's shell. This is also AWS and EKS specific; something analogous will have to be worked out for GCP/GKE.

## AWS credentials

Currently, our operators have an IAM user in each AWS account with full admin privileges, via membership in an IAM group that is granted the FullAdmin policy. I propose to:

- remove all permissions and group memberships from IAM users (see note about the magic `system:masters` member below)
- create role `read-only-access` with policy `arn:aws:iam::aws:policy/ReadOnlyAccess`
- create role `admin-access` with policy `arn:aws:iam::aws:policy/AdministratorAccess`, and a role assumption policy requiring MFA
- attach policies to each IAM user allowing assumption of roles `read-only-access` and `admin-access`, the latter requiring MFA

Then, we have each operator configure their `~/.aws/config` with entries like:

```
[profile my-iam-user]
region = us-west-2

[profile my-iam-user-ro]
role_arn = arn:aws:iam::<ACCOUNT#>:role/read-only-access
source_profile = my-iam-user
role_session_name = <useful string for audit logs>

[profile my-iam-user-admin]
role_arn = arn:aws:iam::<ACCOUNT#>:role/admin-access
source_profile = my-iam-user
role_session_name = <useful string for audit logs>
mfa_serial = arn:aws:iam::<ACCOUNT#>:mfa/<mfa-device-name>
```

[...]

PR #1242, opened 2021-12-22
Remove `deploy-tool`

tgeoghegan:
This series of commits removes `deploy-tool` and moves its last responsibility (generation of keys and manifests for sample data generator) into Terraform. This allows us to remove most of the outputs from the top-level Terraform module, allowing us to later expunge the notion of `aws_profile` from Terraform, which paves the way for the MFA authentication flow that will be introduced in #1170.


BranLwyd (approved):
🎉 You love to see it.

Keep in mind we haven't seen `key-rotator` provision keys for a new environment in prod yet -- only my dev environment. Up to you if you'd like to hold off on merging or not; I suppose the worst we'll have to do is revert a few merges to get back to a working `deploy-tool` if something does go wrong.

tgeoghegan:
I definitely don't want to merge until next year. All of this is part of the work for #1170, which will significantly change how we access production, and I don't want to take on that disruption on the last day before holiday break.

Issue #1031, opened 2021-10-11
Manage EKS `system:masters` members somehow


tgeoghegan:
@BranLwyd and I learned last week that the AWS IAM user that creates an EKS cluster is automatically and invisibly added to a `system:masters` Kubernetes group, allowing management of Kubernetes cluster resources. However, other IAM users get no privileges, even if they have administrator privileges at the AWS layer, and must be explicitly added to the `kube-system/aws-auth` config map in an appropriate `mapUsers` section[1].

We should work out a way to manage this access via Terraform.

[1]: https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html

tgeoghegan:
I discussed what I want to do about this [here](https://github.com/abetterinternet/prio-server/issues/1170#issuecomment-991392687), but I've run into snags while trying to implement it. Introducing readonly and admin IAM roles and using them in the `aws-auth` config map works, but the problem is doing that via Terraform. The `aws-auth` map gets created automatically by EKS, and the Kubernetes provider for Terraform doesn't provide a way to add values to an existing config map, so the only way to do this in Terraform would be to do something complicated with a `local-exec` provisioner, which I would really rather not do. AWS have [an open issue about this](https://github.com/aws/containers-roadmap/issues/185), or rather the general problem of programmatic management of EKS cluster access. Once they provide API for that, and it gets exposed through the `aws` TF provider, we should adopt it. In the meantime, given how rarely we provision clusters, I think we will just document the need to manually fix up the `kube-system/aws-auth` config map, though we can still automate the creation of the AWS IAM roles and the k8s `ClusterRole` and `ClusterRoleBinding`.

tgeoghegan:
The [`prio-server` onboarding docs](https://github.com/abetterinternet/docs/blob/main/prio-server/onboarding.md#eks-aws-auth-configmap) discuss managing the config map. As I noted, I don't think we can do much better until there is EKS API plumbed up through Terraform for us. In the meantime, I'm taking this out of the current tech debt milestone.

PR #1243, opened 2021-12-22
terraform: module for Kubernetes cluster bootstrap 

### LINKED ISSUE #1046: `terraform`: don't configure `kubernetes` provider with output from `gke`/`eks` modules
Our Terraform modules aim to manage an entire env with a single `terraform apply` command. Accordingly, we have modules `eks` and `gke` which manage Elastic Kubernetes Service and Google Kubernetes Engine clusters, respectively, and we use outputs from those modules to configure provider `kubernetes` (e.g., API endpoint, server certificate, that sort of thing). According to [issues on the Terraform project](https://github.com/hashicorp/terraform-provider-kubernetes/issues/1102) and [TF documentation](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#stacking-with-managed-kubernetes-cluster-resources), this is an antipattern, is not supported by Terraform, and may break in future releases.

We should understand what Terraform intends to support in the long term and do whatever refactoring of our modules is necessary to colour inside of the lines Hashicorp is providing. That might mean managing two Terraform states per environment (one to bring up a k8s cluster, one for every thing else), or it might mean more thoroughly dividing our `apply` from `apply-bootstrap` to rule out the possibility of any individual `terraform plan/apply` invocation trying to modify too many resources at once.

tgeoghegan:
Introduces a new module in `terraform/cluster_boostrap` which is
responsible for provisioning a managed GKE or EKS Kubernetes cluster
into a GCP project or AWS account provisioned by `isrg-cloud-bootstrap`.

The bulk of the new module consists of the existing `eks` and `gke`
modules, moved to a new location. We do this in a separate TF state to
avoid configuring a TF provider with output from TF managed resources
(see #1046 for details).

The following functionality is new:
 - creation of an AWS ECR or Google Artifact Registry repository for
   container images via Terraform
 - creation of a ClusterRoleBinding for the read-only group in EKS

This commit removes from `terraform/Makefile` management of the GCS
bucket used to store the Terraform remote state. We now expect the
bucket to have been created manually by the operator. This reduces the
risk of bugs in `Makefile` targets accidentally destroying Terraform
state.

Finally, we align documentation in `terraform/README.md` with docs
introduced in isrg-cloud-boostrap [1] and abetterinternet/docs [2].

[1] abetterinternet/isrg-cloud-bootstrap#1
[2] abetterinternet/docs#10

Resolves #1046
Part of #1170
Part of #1031


tgeoghegan:
Stacked on #1242!

tgeoghegan:
Downgrading to a draft for now -- I still want to tinker with this more before review.

tgeoghegan:
I think the TF stuff is done here, but I need to do a whole bunch of README cleanup, and then make sure that the documentation in `prio-server` is consistent with what's in https://github.com/abetterinternet/isrg-cloud-bootstrap/pull/1 and with https://github.com/abetterinternet/docs/pull/10.

BranLwyd (approved):

PR #1252, opened 2022-01-03
Release 0.6.21, enable key rotation in all prod-us localities.

BranLwyd:


tgeoghegan (approved):
Any reason not to turn this on in `prod-intl` at the same time?

BranLwyd:
> Any reason not to turn this on in `prod-intl` at the same time?

No good technical reason; I'm just trying to touch as few environments at once as possible (& `prod-us` needs to go first due to the expiry issue), and I hadn't thought through who I'd need to coordinate with to rotate the `-intl` environments. I'll plan to do the `-intl` environments once we're happy with `prod-us`.

PR #1260, opened 2022-01-04
Refactor BatchWriter to use a single BatchWriter::write method.

BranLwyd:
This replaces BatchWriter::{packet_file_writer, put_header,
put_signature}. This change has no (intended) functional changes; it
will enable BatchWriter::write to optionally write the batch in the new
single-object, atomic style in an upcoming PR.

I drop the --use-bogus-packet-file-digest because I had to make changes
due to the larger refactoring, and per discussion this flag is not
necessary/desirable for facilitator servers.


BranLwyd:
BTW, I couldn't put the bogus-packet-file-digest functionality behind `#[cfg(test)]` or similar because the integration tests use it, and AFAICT they are compiled in a release configuration(!). Oh well. I only discovered this after the flag was gone, but I think removing the flag is still worthwhile.

facilitator/src/batch.rs
```diff
@@ -299,84 +300,97 @@ impl<'a, H: Header, P: Packet> BatchWriter<'a, H, P> {
...
-    /// some Err() otherwise. packet_file_writer returns the digest of all the
-    /// content written by the operation.
-    pub fn packet_file_writer<F>(&self, operation: F) -> Result<Digest>
-    where
-        F: FnOnce(&mut Writer<&mut DigestWriter<&mut Box<dyn TransportWriter>>>) -> Result<()>,
-    {
+    pub fn write(
+        &self,
+        key: &BatchSigningKey,
+        mut header: H,
```
tgeoghegan, line 321: Does this need to take the header by value, or can it be `&mut header: H`?

facilitator/src/batch.rs
```diff
@@ -299,84 +300,97 @@ impl<'a, H: Header, P: Packet> BatchWriter<'a, H, P> {
...
+        packets: impl Iterator<Item = Result<P>>,
```
tgeoghegan, line 322: I think using [`IntoIterator`](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) would make this a bit more ergonomic. Something like:

```
fn write<I: IntoTerator<Item = Result<P>>(&self, key, header, packets: I) -> Result<()> {
    let packets = packets.into_iter();
    for packet in packets { ... }
}
```

Now the caller can pass in anything that implements `IntoIterator`, so anything from a `Vec[Result<P>]` to `&[Result<P>]` to a hashmap to etc., and we hide the call to `into_iter` here in this method instead of making every caller do it.

[Example of this pattern from `libprio`](https://github.com/abetterinternet/libprio-rs/blob/10e4d168c04546256916d940cde8b4ba9ac432a5/src/vdaf/hits.rs#L142)

I also found some discussion of the tradeoff between making a function generic in `T: Trait` vs. using `impl Trait` [in the Rust book](https://doc.rust-lang.org/reference/types/impl-trait.html#anonymous-type-parameters)

facilitator/src/idl.rs
```diff
@@ -58,6 +58,8 @@ lazy_static! {
 }
 
 pub trait Header: Sized {
+    /// Sets the SHA256 digest of the packet file this header describes.
+    fn set_packet_file_digest(&mut self, digest: impl Into<Vec<u8>>);
```
tgeoghegan, line 62: This is slick, and I like that you're not forcing the caller to heap-allocate a `Vec` here. As elsewhere, I think making this `fn set_packet_file_digest<D: Into<Vec<u8>>(&mut self, digest: D)` is slightly better.

facilitator/src/batch.rs
tgeoghegan: Separately, I'm not sure how I feel about `Item = Result<P>`. It seems like this forces most of the callers of `BatchWriter:write` to wrap their packets in `Ok()` to accommodate one callsite where the packet construction can fail. I'm OK with it, but I did want to point you to [this neat feature of `Result`](https://doc.rust-lang.org/rust-by-example/error/iter_result.html). Instead of taking `Iterator<Item = Result<P>>`, you could take `Iterator<Item = P>`, and expect/require callers that have an iterator of `Result<P>` to use a `.collect()` call to map those into `P` while still handling errors. The other advantage is that errors that occur in the calling function no longer get handled here in `write()`.

That being said, I think what you have here works and is good enough for an internal interface, so if you prefer this to putting a call to `collect` with a turbofish in `intake.rs`, that's fine.

tgeoghegan (changes requested):

facilitator/src/batch.rs
BranLwyd: It could be a mut ref, but I prefer to pass ownership since `write` needs to modify the header in what is (from the caller's POV) a somewhat arbitrary fashion, & it's not clear that there are "normal" (i.e. production) uses where we want to keep the header around afterwards.

For tests & other places that want/need to keep the header around, all headers can be cloned. WDYT?

facilitator/src/batch.rs
BranLwyd: I like switching to `IntoIter` & switching to `T: Trait` over `impl Trait`. Done.

I slightly preferred using an iterator of `Result<P>` only because it allows us to potentially only parse a single packet at a time, instead of needing to keep all in memory (as is necessary if we collect to a `Vec`). But we don't actually make use of that, and your suggestion is certainly easier to read, so done. (I did a little digging at the realistic packet sizes -- for ingestion packets in `us-ca`, the largest packet file I could find was a few megabytes total -- so I suspect any potential memory savings is unnecessary at the moment.)

facilitator/src/batch.rs
BranLwyd: It could be a mut ref, but I have a slight preference for passing ownership: the header that's passed is effectively a "template" since `write` needs to modify it in a way that is strange/arbitrary to the caller (i.e. `write` needs to add a digest). By passing ownership, the details of how the header is modified aren't included in the contract for `write` since it's `write`'s header anyway.

[Also, no "normal" (i.e. production) uses of `write` need to keep the header around after calling `write`, so we don't lose anything by consuming the header. For tests that want to keep the header around to e.g. check that it is round-tripped appropriately, all headers can be cloned.]

tgeoghegan: I'm OK with this if the intent is to consume the header.

tgeoghegan (approved):

PR #1262, opened 2022-01-05
terraform: use `kubectl_manifest`

### LINKED ISSUE #1088: Can't plan or apply `kubernetes_manifest` resources unless Kubernetes namespace already exists
in `terraform/modules/kubernetes/kubernetes.tf`, we create a couple of `kubernetes_manifest` resources. Those manifests get deployed into a Kubernetes namespace. `terraform plan`-ning the creation of those resources evidently entails (the equivalent of) `kubectl apply --dry-run`. If the namespace in question doesn't exist, then the manifest application dry run fails because the Kubernetes cluster doesn't know about the namespace we propose to deploy the manifest into.

This breaks deploying new localities, because that operation involves creating a namespace and creating a `kubernetes_manifest` in one apply operation. Curiously, `make apply-bootstrap` doesn't work around this: in spite of `apply-bootstrap` threatening to create namespaces during its planning phase, it doesn't actually do it. Presumably this is because of the weird behavior around `-target` that `terraform` warns us about.

tgeoghegan:
The `kubernetes_manifest` resource in provider `hashicorp/kubernetes`
has a known issue[1] where resources created in a manifest can't depend
on other resources that don't exist yet. To work around this, we instead
use `gavinbunney/kubectl`'s `kubectl_manifest` resource, which does not
have this problem because it uses a different mechanism for planning.

[1] https://github.com/hashicorp/terraform-provider-kubernetes/issues/1380

Resolves #1088

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -323,6 +323,10 @@ terraform {
       source  = "hashicorp/tls"
       version = "~> 3.1.0"
     }
+    kubectl = {
+      source  = "gavinbunney/kubectl"
+      version = "~> 1.13.1"
+    }
   }
 }
 
@@ -366,9 +370,6 @@ provider "kubernetes" {
   host                   = local.kubernetes_cluster.endpoint
   cluster_ca_certificate = base64decode(local.kubernetes_cluster.certificate_authority_data)
   token                  = local.kubernetes_cluster.token
-  experiments {
-    manifest_resource = true
-  }
 }
 
 provider "helm" {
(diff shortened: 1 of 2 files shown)


tgeoghegan:
Pulling in another TF provider just to work around a bug in the Hashicorp provider isn't ideal, but the alternative would have been to introduce a multi-stage Terraform process so that we could guarantee that all Kubernetes namespaces already exist before we try to deploy a manifest, so I think this is better.

BranLwyd (approved):

PR #1263, opened 2022-01-05
Workflow manager: consider batch complete if signature exists.

BranLwyd:
I think this is safe for both new-style (single-object) & old-style
(multi-object) batches. Single-object batches are safe because they
really are just a single object, so checking for that one object is the
correct thing to do. Multi-object batches are safe because the signature
object was always written last (& only if writing the header & packet
objects succeeded), and both AWS & GCP offer strong read-after-write &
list consistency [1][2], so if the signature object was written we know
the batch & packet objects will also be written.

[1] https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html#ConsistencyModel
[2] https://cloud.google.com/storage/docs/consistency

### DIFF
--- a/workflow-manager/batchpath/batchpath.go
+++ b/workflow-manager/batchpath/batchpath.go
@@ -20,9 +20,8 @@ type BatchPath struct {
 	dateComponents []string
 	ID             string
 	Time           time.Time
-	metadata       bool
-	avro           bool
-	sig            bool
+
+	sigExists bool
 }
 
 // List is a type alias for a slice of BatchPath pointers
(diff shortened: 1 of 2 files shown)


BranLwyd:
I set my dev environment up to have one locality use multi-object validation batches everywhere, one locality use single-object validation batches everywhere, and one locality use a mixture of single-object & multi-object validation batches. With this change, I can see all three localities generating sum part data.

BranLwyd:
Hmm. I suppose we don't know if the ingestion servers write their batches signature-last. I'm going to make this functionality change only for validation batches.

BranLwyd:
OK, this should be ready for review once again; apologies for the churn.

tgeoghegan (approved):

PR #1275, opened 2022-01-11
Terraform: config to allow single-object validation batches to be used.

BranLwyd:
Defaults to disabled (as needed currently, until all production
deployments are released to a recent-enough version to be able to parse
single-object validation batches).

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -267,6 +267,16 @@ variable "stackdriver_exporter_helm_chart_version" {
   default = ""
 }
 
+variable "single_object_validation_batch_localities" {
+  type        = list(string)
+  default     = []
+  description = <<DESCRIPTION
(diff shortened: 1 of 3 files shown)


terraform/main.tf
```diff
@@ -267,6 +267,16 @@ variable "stackdriver_exporter_helm_chart_version" {
   default = ""
 }
 
+variable "single_object_validation_batch_localities" {
```
tgeoghegan, line 270: nit: Expressing this in this manner opens the possibility of typos, where we might have `ingestors.ingestor-1.localities.gondor`, but set `single_object_validation_batch_localities["gnodor"]`. We could instead add a boolean valued field to the `ingestors.localities` object. However what you've got here is consistent with how we did the key rotation localities so either is fine with me.

tgeoghegan (approved):

PR #1276, opened 2022-01-12
Fix monitoring, update Prometheus/Grafana to latest available versions.

BranLwyd:
The ultimate issue with the monitoring was that the prometheus URL was
specified as "prometheus-server" rather than "http://prometheus-server".
I'm not sure why it used to work, but I suppose some update added
additional validation requiring the scheme to be specified.


tgeoghegan (approved):
Good find, I'm quite pleased that we root caused the problem.

Issue #1274, opened 2022-01-11
Migrate staging and production environments to new Terraform schema


tgeoghegan:
https://github.com/abetterinternet/prio-server/pull/1243 & https://github.com/abetterinternet/docs/pull/10 & https://github.com/abetterinternet/isrg-cloud-bootstrap introduce new Terraform modules to address a number of issues. This issue tracks preparing and executing an implementation plan for migrating staging and production environments to the new stuff. At a high level, this entails:

- importing GCP projects and AWS accounts into `isrg-cloud-bootstrap`
- Fixing or creating up AWS IAM roles in member accounts as needed
- importing existing GKE/EKS resources into each env's `cluster_bootstrap` module state
- removing GKE/EKS resources from each env's existing TF state
- ensuring that each EKS cluster's `kube-system/aws-auth` config map is as expected
- scrubbing GCP project IAM lists to excise sre@letsencrypt.org (wrap up #1056)

tgeoghegan:
First step for me is to write a detailed implementation plan based on my experience doing this for our dev envs.

tgeoghegan:
# Staging migration deployment plan

## Project git commit SHAs

The steps documented here will be performed with the following commits:

- [`prio-server` 21b25cd567d757ca52ec79fe5fc1f23ed0ed9ef0](https://github.com/abetterinternet/prio-server/commit/21b25cd567d757ca52ec79fe5fc1f23ed0ed9ef0)
- [`isrg-cloud-boostrap` d1f068d418bcbed318383b2fcd3624c5eb393cc1](https://github.com/abetterinternet/isrg-cloud-bootstrap/commit/d1f068d418bcbed318383b2fcd3624c5eb393cc1)

## Import staging AWS accounts and GCP projects into `isrg-cloud-bootstrap`

The following GCP projects need to be migrated:

- `prio-staging`
- `prio-staging-key-backup`
- `prio-intl-staging`

`staging-intl` is our only staging EKS environment and it is hosted in the management AWS account, so it does not need to be migrated.

* Activate a shell with MFA credentials for management account admin profile
    * `./aws-mfa --profile enpa-intl-staging --role-arn arn:aws:iam::024759592502:role/admin-access --role-session-name timg --mfa-device-arn arn:aws:iam::024759592502:mfa/timg --token-code XXXXXX bash`
* Make a copy of existing `isrg-cloud-boostrap/enpa` Terraform state
  * `cd /path/to/isrg-cloud-bootstrap`
[...]

PR #1277, opened 2022-01-12
terraform: speed up prep targets in Makefile

tgeoghegan:
Historically, we blow away `.terraform` (or
`cluster_bootstrap/.terraform`, depending on which `prep` target is
invoked) before initializing `terraform`. The idea is to avoid the risk
of one environment's `.tfstate` bleeding into another, which `terraform
init` can sometimes sneakily do to you.

While working on #1274, I found myself switching between remote and
local Terraform backends a lot. My implementation plans had a lot of `rm
-rf .terraform` in them, in accordance with the `prep` and
`prep-cluster-bootstrap` targets in `terraform/Makefile`. The downside
of destroying the entire directory is that you have to reinstall the
Terraform providers every time you `plan` or `apply`. I learned that
what's really important is the file `.terraform/terraform.tfstate`, and
that if we destroy just that file in the `prep` targets, we avoid the
risk of cross-contamination of backends but can speed up `prep`
considerably by not re-downloading providers.

### DIFF
--- a/terraform/Makefile
+++ b/terraform/Makefile
@@ -31,7 +31,7 @@ help:
 
 .PHONY: prep
 prep: ## Prepare a new workspace (environment) if needed, configure the tfstate backend, update any modules, and switch to the workspace
-	@rm -rf .terraform/
+	@rm .terraform/terraform.tfstate
 	@if [ ! -f "$(VARS)" ]; then \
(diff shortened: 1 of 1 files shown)


BranLwyd (approved):
Thank you so much for this PR, this will dramatically shorten the loop when working with Terraform config.

Issue #226, opened 2020-11-24
Automated integration/correctness testing


tgeoghegan:
We should be able to automatically run end to end integration tests against instances of the application running in a cloud. The ideal would be the ability to run such a test on any PR, but I'd settle for a nightly test run against a staging environment.

BranLwyd:
Per chat with @tgeoghegan today, these tests should include correctness testing; i.e. check that the (generated) ingestion data matches the aggregated output in the environment.

tgeoghegan:
Solving this would entail creating a new deployment whose job it is to reassemble sum parts and verify that the outputs are as expected. Beyond that, we would have to either create a channel between the sample generation task and the sum verification task to allow the former to transmit expected sums to the latter, OR somehow make sample generation deterministic so that the sum verifier can figure out the expected sums without explicit communication.

PR #1283, opened 2022-01-14
Generate sample data deterministically from packet UUID.

BranLwyd:
The UUID is itself random, so sample data is still random; the upside of
this is that we can now determine a given sample ingestion packet's data
without reading & decrypting the data shares. This will be handy for a
validator which wants to make sure the sums we are emitting match up
with the ingestion packets included in those sums.

[Note that since even "random" v4 UUIDs contain a few deterministic
bytes, part of the data will be deterministic; I think that's OK.]

### DIFF
--- a/facilitator/src/sample.rs
+++ b/facilitator/src/sample.rs
@@ -13,6 +13,7 @@ use prio::{
     encrypt::PublicKey,
     field::{FieldElement, FieldPriov2},
 };
+use ring::digest;
 use slog::{info, o, Logger};
 use uuid::Uuid;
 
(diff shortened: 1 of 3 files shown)


BranLwyd:
I'm doing this to address #226, at least as far as validating that the sums we output are equal to the sum of the ingestion packets that go into them. My basic implementation strategy is something like:

1. Generate sample data in a way that allows the data to be determined from the packet itself [this PR].
2. Write a validator job that:
  a. Reads sum parts, computes actual sum.
  b. Reads all validation batches included in the sum, based on the validation batch IDs.
  c. Computes expected sum by computing from ingestion UUIDs in validation batches.
  d. Yells if the expected & actual sums don't match up.

The biggest failure case this would miss would be something of the form "we somehow completely ignored an ingestion packet" -- I haven't thought that part through too hard, but we could potentially later extend the validator job to also look for ingestion packets that "should" be included in the validation batches but aren't, and yell about them too.

Note this also ignores the various "bug-simulation" features of the sample generator (short packet, dropped packets) -- since the sample generator job doesn't use them, that's OK for now.

tgeoghegan (changes requested):
This is quite clever, but I'm not sure it'll quite work. You are deriving inputs from _packet_ UUIDs, but the portal server/collector/sum verifier you'll eventually write will have the list of _batch_ UUIDs, so you won't be able to rederive the inputs on that end. However, I think you could instead make the derivation a deterministic function of `(batch UUID, packet index)`. Then, on the sum verifier end, if you get an aggregation with 100 contributions over 10 batches, you can assume that each batch contained 10 inputs and re-derive the inputs.

BranLwyd:
Ugh, you're correct of course. The sequence of values is now deterministic based on `batch_id` (and `dimension`, for the length). I also pulled this logic out to a separate function, since I'll need to be able to call it from elsewhere soon anyway.

facilitator/src/sample.rs
```diff
@@ -339,6 +335,43 @@ impl<'a> SampleGenerator<'a> {
     }
 }
 
+/// Returns an iterator over sample datas; each will be a vector of length
```
tgeoghegan, line 338:
```suggestion
/// Returns an iterator over sample data; each datum will be a vector of length
```

tgeoghegan (approved):
Rather elegant use of an iterator! I'd say that we should use an HMAC instead of the bespoke SHA256 construction but the output isn't security critical.

BranLwyd:
> Rather elegant use of an iterator! I'd say that we should use an HMAC instead of the bespoke SHA256 construction but the output isn't security critical.

BTW -- this doesn't matter much/at all, but I'd argue that the existing hash-based construction is preferable to a MAC-based construction. I'm trying to implement a collision-resistant function over (batch_uuid, index). I presume you're suggesting `MAC(K = batch_uuid, M = index)` [or possibly flipping key/message, it's arbitrary]. I prefer the current implementation of `H((batch_uuid, index))`, with the tuple serialized as a concatenation of the bytes of its parts (which works since both parts are of constant size). Both constructions produce a collision-resistant value based on (batch_uuid, index); the advantage of the hash-based construction over the MAC-based construction is that the hash-based construction is slightly simpler.

Issue #1285, opened 2022-01-14
`facilitator`: Occasional hang when accessing S3 from GKE


tgeoghegan:
We have occasionally observed that `facilitator` `intake-batch` workers get stuck, inasmuch as they no longer do work, but are still running.

Our current hypothesis is that there is a bug in our code that manages GCP access and/or identity tokens causing a deadlock.

This issue tracks finding and fixing the problem. I'm using it to compile observations and data about the bug.

tgeoghegan:
@hostirosti has observed this in the Google instance used for international ENPA. ISRG has observed this in the instance we run for US ENPA. When we detect this problem, we see that the last thing logged by a stuck facilitator is usually (always?) that it was doing a PUT to S3. We also observe that the stuck facilitator can still respond to metrics scrapes after the thread trying to write to S3 gets stuck, suggesting the process is still running, but the thread writing to S3 can't make progress.

We haven't caught an occurrence of this in an EKS cluster, but we can't be sure it never happens there.

tgeoghegan:
We obtained some core dumps from stuck `facilitator` processes. I've attached the backtraces from all threads to this comment. What we observe is thread #1 stuck under `<facilitator::transport::s3::S3Transport as facilitator::transport::Transport>::put`, which in turn is stuck on `tokio::runtime::Handle::block_on` (slightly different stack in the case of Robert's core, which is from a different `facilitator` version).

The other threads seem to be Tokio workers waiting for work to do, but I'm not sure I'm reading the backtraces correctly. My feeble understanding of Tokio runtimes is that we expect there to be two threadpools in a Tokio runtime, one dedicated to well-behaved tasks that yield regularly (via `.await`) and another dedicated to blocking operations ([details](https://docs.rs/tokio/latest/tokio/task/index.html). So it might be that threads 3 and 4 are the blocking and non-blocking Tokio thread pools.

There is a further thread with some `crossbeam` frames in it. `Cargo.lock` suggests that only `slog-async` depends on `crossbeam` so this could be a dedicated thread for logging.

[stuck-rkubis-backtrace.txt](https://github.com/abetterinternet/prio-server/files/7872386/stuck-rkubis-backtrace.txt)
[stuck-us-ca-backtrace.txt](https://github.com/abetterinternet/prio-server/files/7872387/stuck-us-ca-backtrace.txt)
[...]

PR #1286, opened 2022-01-14
S3 transport put: add 10-minute timeout to starting multipart upload.

BranLwyd:
I spent a lot of today looking at this code -- in #1285, all three stack
traces are getting stuck in S3Transport::put, and more specifically in
the block_on call modified by this commit. I didn't find a smoking gun,
but I'd like to add a timeout to this request.

I suppose there are two possibilities for this deadlock:
 1) The deadlock is a true "thread" deadlock, e.g. waiting on a mutex
    that the current thread already owns. In this case, I'm not sure the
    new timeout will help.
 2) The deadlock is due to some future in create_multipart_upload not
    correctly implementing the future interface, e.g. not arranging for
    the wakeup to occur when progress can be made. In this case, the
    timeout will effectively sidestep the bug (in the sense that we'll
    now give up after waiting ten minutes, rather than waiting forever),
    and we'll have a new error in the logs confirming that this is what
    is happening.

That is, this commit may provide a workaround for the issue and even if
it doesn't, won't put us in a worse position; and either way, we have
more information. I suspect that we're in situation #2 since, from how
I'm interpreting the stack traces, we don't appear to be waiting in
application code (but async is complex enough that I might reasonably be
incorrect about that reasoning.)

Why ten minutes? That's how long retry_request is hardwired to try
before giving up, so it seemed like an OK choice.


tgeoghegan (changes requested):
I think this change is a good idea even if it doesn't wind up fixing the deadlock. However, I think we should apply this to all API calls we make via `rusoto`, since apparently [none of them are subject to any timeout by default](https://github.com/rusoto/rusoto/issues/1795). What do you think about putting the timeout logic into `aws_credentials::retry_request`, since that function is only ever used for AWS API requests?

BranLwyd:
> I think this change is a good idea even if it doesn't wind up fixing the deadlock. However, I think we should apply this to all API calls we make via `rusoto`, since apparently [none of them are subject to any timeout by default](https://github.com/rusoto/rusoto/issues/1795). What do you think about putting the timeout logic into `aws_credentials::retry_request`, since that function is only ever used for AWS API requests?

I like the idea of expanding the timeout to apply to all retried rusoto operations -- I had no idea that rusoto didn't have timeouts on its IO operations by default, that's may very well be the root cause!

tgeoghegan (approved):

PR #1295, opened 2022-01-18
Adding randomized part to KMS Keyring name

hostirosti:
Google KMS Keyrings cannot be deleted, just disabled.
This creates a conflict with terraform in a destroy / create cycle
Adding a randomized string to the KMS Keyring name avoids this conflict

### DIFF
--- a/terraform/cluster_bootstrap/modules/gke/gke.tf
+++ b/terraform/cluster_bootstrap/modules/gke/gke.tf
@@ -128,10 +128,17 @@ resource "google_container_node_pool" "worker_nodes" {
 
   depends_on = [google_project_service.compute]
 }
+      
+resource "random_string" "kms_id" {
+  length  = 8
+  upper   = false
+  number  = false
+  special = false
+}
 
 # KMS keyring to store etcd encryption key
 resource "google_kms_key_ring" "keyring" {
-  name = "${var.resource_prefix}-kms-keyring"
+  name = "${var.resource_prefix}-${random_string.kms_id.result}-kms-keyring"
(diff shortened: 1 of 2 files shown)


hostirosti:
ran the main.tf through `terraform fmt` and can't find any changes. Can we rerun the CI tests?

tgeoghegan:
It's complaining about `terraform/cluster_bootstrap/modules/gke/gke.tf` which has an extra tab above `resource "random_string" "kms_id"`. `terraform fmt --recursive` should fix it.

I think this change is a good idea, but the difficulty is that existing deployments already have key rings without the random string component and so applying this change would cause Terraform to make new keyrings, and then new `etcd` and GCS encryption keys. I don't think there's a way to rename a GCP keyring such that we could import existing keyrings into this naming scheme. Can we copy keys from one keyring to another?

hostirosti:
@tgeoghegan, indeed backward compatibility is an issue here. We can abandon this PR or I'll change it into a comment for future reference if people come across this looking for guidance. This issue only surfaces if you `apply` - `destroy` - `apply` in the same project with the same tfvars (resource_prefix).

tgeoghegan:
That's what I was afraid of. This is still a good idea, though, and a good change to take for the bringup of future clusters. I wonder whether there's a good way to tell Terraform to use the random string for creating new keyring resources, but not to try changing the name of existing ones. I think a [`lifecycle` block](https://www.terraform.io/language/meta-arguments/lifecycle) on the `google_kms_key_ring` resource with `ignore_changes = [name]` might do it but I'm not sure if there would be side effects.

divergentdave:
Closing as this formed the basis of #1364

PR #1299, opened 2022-01-19
Key rotator: add flags to skip manifest pre/post-update validations.

BranLwyd:
These are unsafe to use in general--the pre-update validations serve as
a check that the key-rotator is running against a consistent set of
(manifest, k8s secret), the post-update validations serve as a check
that our update logic didn't do something very unexpected. But they may
be necessary if the key-rotator needs to run to fix an environment that
is already out-of-sync (and therefore triggering validation failures).

### DIFF
--- a/key-rotator/manifest/manifest_test.go
+++ b/key-rotator/manifest/manifest_test.go
@@ -332,6 +332,8 @@ func TestUpdateKeys(t *testing.T) {
 func TestPostUpdateKeysValidations(t *testing.T) {
 	t.Parallel()
 
+	now := time.Now()
+
 	// We check only validation check failures here, and we do so by directly
 	// calling `validateUpdatedManifest` rather than calling `UpdateKeys`
 	// because there is no/should be no way to trigger these checks via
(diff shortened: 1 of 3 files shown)


key-rotator/manifest/manifest_test.go
```diff
@@ -775,10 +796,18 @@ func pek(tss ...int64) key.Key {
 }
 
 // manifestBSK creates a manifest BatchSigningPublicKeys with the given
-// timetamps. Order does not matter. Key material is arbitrary, but will match
-// that of other batch signing keys at the same timestamp, and will very likely
-// not match other key materials.
+// timetamps, expiring at the current time. Order does not matter. Key material
```
tgeoghegan, line 799:
```suggestion
// timestamps, expiring at the current time. Order does not matter. Key material
```

key-rotator/manifest/manifest_test.go
```diff
@@ -775,10 +796,18 @@ func pek(tss ...int64) key.Key {
...
+// is arbitrary, but will match that of other batch signing keys at the same
+// timestamp, and will very likely not match other key materials.
 func manifestBSK(tss ...int64) BatchSigningPublicKeys {
+	return manifestBSKWithExpiration(time.Now(), tss...)
+}
+
+// manifestBSKWithExpiration creates a manifest BatchSigningPublicKeys with the
+// given timetamps, expiring at the given timestamp. Order does not matter. Key
```
tgeoghegan, line 807:
```suggestion
// given timestamps, expiring at the given timestamp. Order does not matter. Key
```

tgeoghegan (approved):

PR #1315, opened 2022-01-24
Pass key directly to integration-tester worker.

BranLwyd:
Previously, the key was looked up by interfacing with the k8s cluster.
Making this change allows us to drop our dependencies on the kube &
k8s-openapi crates, as well as making the facilitator k8s-agnostic once
again. This also makes the key-distribution to the integration tester
job uniform with all other jobs that receive private keys.

Currently running in my dev environment, where it is successfully
generating samples.

This fell out of a dependabot PR trying & failing to update the
k8s-openapi crate -- I went in to investigate/fix and then realized we
shouldn't be using this crate at all.


tgeoghegan (approved):
-223 lines in `Cargo.lock` is my fave kind of change.

PR #1316, opened 2022-01-25
terraform: opt into spot VMs on GKE

tgeoghegan:
@hostirosti pointed out a recent GKE feature that lets us opt into
spot VMs in our GKE cluster. Much like AWS spot instances, spot VMs are
drawn from spare Google Compute Engine capacity and cost a lot less, in
exchange for lesser availability, which is acceptable for a batch
processing system like ours.

https://cloud.google.com/kubernetes-engine/docs/concepts/spot-vms

### DIFF
--- a/terraform/cluster_bootstrap/modules/gke/gke.tf
+++ b/terraform/cluster_bootstrap/modules/gke/gke.tf
@@ -97,6 +97,9 @@ resource "google_container_cluster" "cluster" {
 }
 
 resource "google_container_node_pool" "worker_nodes" {
+  # Beta provider is required for the spot VMs feature
+  provider = google-beta
+
   name               = "${var.resource_prefix}-node-pool"
   location           = var.gcp_region
   cluster            = google_container_cluster.cluster.name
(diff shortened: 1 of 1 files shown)


BranLwyd (approved):

tgeoghegan:
I've deployed this to my GCP devenv and I'm going to let it bake there for a bit before merging, but TBH I'm not sure what validation we might do beyond verifying that we still get any VMs.

divergentdave:
Given that we have one node pool, does `spot = true` imply all our instances will be spot instances? I'm concerned that this could lead to issues if all our instances get terminated. Should we add a second node pool in the same cluster with on-demand nodes, and update the relevant autoscaling parameters?

tgeoghegan:
Maybe you're right. We did run into something like that on the EKS side, where we were able to get as many spot instances as we'd asked for, but not in all the AZs we asked for, and so `prometheus-server` couldn't be scheduled because we didn't have any instances in the same AZ as the EBS volume it uses for metrics.

I'm going to revert this while we ponder that question. I don't think I'll get back around to looking at this in the immediate future though, so @divergentdave feel free to take this PR and run with it.

PR #1317, opened 2022-01-28
Add sample validator job.

### LINKED ISSUE #226: Automated integration/correctness testing

BranLwyd:
Not that I doubted this would be the case, but I breathed a sigh of
relief to see that the validations pass. I'm seeing some
object-retrieval errors in my own dev environment; I'm still
investigating if these indicate a bug in this code or in my dev
environments, but I think this is ready for review.

A few notes:
 1) The terraform is perhaps questionable--please pay close attention to
    it in review. Specifically, the fake_server_resources module now needs
    to refer to resources created by the *peer's* deployment, and
    effectively does so by constructing the proper names by hand. This
    works, but is fragile -- if we ever rename one of these resources,
    we now have to remember to do so in fake_server_resources too.

    Separately, I think fake_server_resources could probably use a
    refactoring pass (I left a TODO). The problem is that it refers to
    "own" and "peer" resources, but it's not logically part of either
    deployment. Instead, it should refer to "facilitator" and "pha", I
    suppose.

    The validator also needs to run in GCS: it takes advantage of GCS'
    feature allowing multiple separate subscriptions to a single pubsub
    topic.

 2) I had to fix a bug in BatchReader/TransportReader to allow us to
    read SumPart batches without any invalid packets (since
    TransportWriter will not create 0-byte objects due to a bug). See
    the relevant comment for details, but in short I fixed it this way
    because fixing the bug in TransportWriter would constitute a
    behavior change visible to other parties (specifically MITRE, who
    read the sum-part buckets in production).

 3) I left a few other TODOs for some bugs I spotted that did not have
    to be fixed in this commit -- the commit is large enough as it is.

Closes #226.


BranLwyd:
OK, my dev-env testing looks good now & I think this is ready for review -- it had failed to generate a few sum parts due to an unknown but no-longer-occurring issue during development, which led to failed message handling; since there wasn't a DLQ specified previously, it would retry forever.

Separately, the IAM bindings to allow the validator to take on the facilitator & PHA sum-part service accounts got disconnected(?). I updated the logic, but I'm not sure if this is a fix or if I just needed to deploy to my dev environment again.

terraform/modules/fake_server_resources/fake_server_resources.tf
```diff
@@ -224,6 +277,133 @@ resource "kubernetes_deployment" "integration-tester" {
   }
 }
 
+# Dead letter topic to which undeliverable intake batch messages are sent
```
tgeoghegan, line 280: I'm guessing that you didn't re-use the `pubsub/pubsub.tf` module because in this case, you don't want to create a topic since you're adding an extra subscription to the existing aggregate topic? And I suppose that refactoring the `pubsub` module to only conditionally create a topic would cause a lot of existing `google_pubsub_topic*` resources to get renamed, which invalidates existing Terraform state?

I have to admit that I can't think of a way out of this that doesn't incur the pain of moving TF resources or avoids duplicating this handful of resources *somewhere*. It almost makes me wish Terraform had a simple preprocessor with an `#include` directive so we could at least delegate the copy-pasting to a program.

terraform/modules/fake_server_resources/fake_server_resources.tf
```diff
@@ -224,6 +277,133 @@ resource "kubernetes_deployment" "integration-tester" {
...
+resource "google_pubsub_topic" "validate_dead_letter" {
+  for_each = var.ingestor_pairs
+  name     = "${var.environment}-${each.key}-validate-dead-letter"
+}
+
+locals {
+  # GCP PubSub creates a service account for each project used to move
+  # undeliverable messages from subscriptons to a dead letter topic
```
tgeoghegan, line 288:
```suggestion
  # undeliverable messages from subscriptions to a dead letter topic
```
The typo also exists in the original module

tgeoghegan (changes requested):
This is great, thank you for taking the time do to this! Besides the inline comments, some meta-thoughts:

- You observe that you had to make assumptions about the layout of the peer environment to make this work and it's a little gross. If you want to do better, you could formalize this a bit by using [Terraform's remote state feature](https://www.terraform.io/language/state/remote-state-data): instead of assuming some service account or bucket will exist in the remote env, you could fetch the peer env's TF state and consult output variables. Then again, it's a bunch more work to achieve the same result, and it's not clear that assuming that some output variable will exist in the peer TF state is that much better than assuming that some service account and GCS bucket will exist.
- This test is driven by aggregation tasks enqueued by the facilitator environment's `workflow-manager`. That means that we won't catch scenarios where the PHA ran an aggregation but the facilitator didn't schedule it for some reason. I think a reasonably lightweight way to handle that would be to have the validator job do something similar to what the MITRE bot does: scan the sum bucket contents and compare them against each other, and raise an error if it ever sees an unbalanced sum part. I'd be happy to log an issue for that enhancement and move forward with this change which is already a huge leap beyond what we had.

BranLwyd (commented):
> * You observe that you had to make assumptions about the layout of the peer environment to make this work and it's a little gross. If you want to do better, you could formalize this a bit by using [Terraform's remote state feature](https://www.terraform.io/language/state/remote-state-data): instead of assuming some service account or bucket will exist in the remote env, you could fetch the peer env's TF state and consult output variables. Then again, it's a bunch more work to achieve the same result, and it's not clear that assuming that some output variable will exist in the peer TF state is that much better than assuming that some service account and GCS bucket will exist.

That would be much cleaner. But I think it would also introduce an ordering requirement on deployment (deployment w/o test before deployment w/ test), as well as some additional complexity in the "production" terraform in support of the "test" terraform. For now, I think I'd find it acceptable to construct the names; if this causes problems, I suppose there's nothing stopping us from refactoring to use remote state.

> * This test is driven by aggregation tasks enqueued by the facilitator environment's `workflow-manager`. That means that we won't catch scenarios where the PHA ran an aggregation but the facilitator didn't schedule it for some reason. I think a reasonably lightweight way to handle that would be to have the validator job do something similar to what the MITRE bot does: scan the sum bucket contents and compare them against each other, and raise an error if it ever sees an unbalanced sum part. I'd be happy to log an issue for that enhancement and move forward with this change which is already a huge leap beyond what we had.

Agreed -- I had some high-level thoughts about this that were basically "look at the files written and figure out if we didn't include a packet somewhere???". I think we should log an issue for future enhancement.

facilitator/src/transport/gcs.rs
```diff
@@ -134,6 +134,14 @@ impl Transport for GcsTransport {
         let response = self
             .agent
             .call(&logger, &request, "get")
+            .map_err(|err| {
+                if let Error::HttpError(ref ureq_err) = err {
+                    if matches!(ureq_err, ureq::Error::Status(404, _)) {
```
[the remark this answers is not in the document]
tgeoghegan: My worry was that GCS might return 403 for objects that don't exist to stop an attacker from distinguishing between 403 and 404 to learn that an object exists. However, my reading of [the API docs](https://cloud.google.com/storage/docs/json_api/v1/status-codes#403_Forbidden) is that if we should consistently get 404 if we have appropriate permissions on the bucket and objects therein, so I find it hard to believe we would get burned by this in any configuration that isn't kinda busted anyway.

terraform/modules/fake_server_resources/fake_server_resources.tf
tgeoghegan: Let's stick with this. All the "better" options would incur some operational pain.

tgeoghegan (approved):

Issue #659, opened 2021-05-21
Rebuild `prio-server` staging 


tgeoghegan:
Currently, our staging environment is two instances of `prio-server`, both running on GCP/GKE, though one of them has some storage buckets in S3 to simulate the exchange of data with a peer running in AWS. Once we support running the whole application in EKS, we should rebuild our existing staging environment so that it exchanges data between one instance fully running in GCP/GKE and one fully running in AWS/EKS.

tgeoghegan:
I want to have one staging environment that exercises both the "old style" (prod-us) and "new style" (international) manifest schemas. To achieve this, we'll need some more Terraform work so that one env can have global manifests with format version 0 and 1, and then localities that use either format version 1 or 2 for their specific manifests.

BranLwyd:
With the two changes above to coalesce all versions of the DSP gobal/locality-specific manifests into a single version, we no longer need to do the work to deploy multiple versions of these manifests to various different localities for testing.

PR #1330, opened 2022-02-01
Coalesce DSP-specific manifests versions 1 & 2 into a single version.

BranLwyd:
I call the new version "V1or2" -- naming assistance very welcome.
AFAICT, the only difference between the two versions it that V2 added an
optional "peer_validation_identity" field which defaulted to None.
Generic code that attempted to extract a peer_validation_identity from a
V1 manifest would get None, as well. Since this isn't a
backwards-incompatible change, we might as well simplify things by going
back to only a single version.

This simplifies #659 in that we will no longer need to represent both
DSP manifest formats for testing since there is now only a single
manifest format.

### DIFF
--- a/facilitator/src/manifest.rs
+++ b/facilitator/src/manifest.rs
@@ -200,8 +200,7 @@ pub struct DataShareProcessorGlobalManifestV1 {
 pub enum DataShareProcessorSpecificManifest {
     // format version 0 was used in Narnia integration tests and is no longer
     // supported
-    V1(DataShareProcessorSpecificManifestV1),
-    V2(DataShareProcessorSpecificManifestV2),
+    V1or2(DataShareProcessorSpecificManifestV1or2),
 }
 
 impl DataShareProcessorSpecificManifest {
@@ -230,13 +229,9 @@ impl DataShareProcessorSpecificManifest {
             .context("format value in manifest has wrong type")?;
 
         let manifest = match format {
-            1 => Self::V1(
+            1 | 2 => Self::V1or2(
                 serde_json::from_slice(json)
                     .context("failed to decode v1 specific manifest from JSON")?,
-            ),
-            2 => Self::V2(
-                serde_json::from_slice(json)
-                    .context("failed to decode v2 specific manifest from JSON")?,
             ),
             _ => return Err(anyhow!("unsupported manifest format {}", format).into()),
         };
@@ -250,8 +245,7 @@ impl DataShareProcessorSpecificManifest {
     /// identifier to the public keys on success, or an error otherwise.
     pub fn batch_signing_public_keys(&self) -> Result<BatchSigningPublicKeys> {
         let pem_keys = match self {
-            Self::V1(manifest) => &manifest.batch_signing_public_keys,
-            Self::V2(manifest) => &manifest.batch_signing_public_keys,
+            Self::V1or2(manifest) => &manifest.batch_signing_public_keys,
         };
         let mut keys = HashMap::new();
         for (identifier, public_key) in pem_keys.iter() {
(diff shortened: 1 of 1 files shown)


facilitator/src/manifest.rs
```diff
@@ -200,8 +200,7 @@ pub struct DataShareProcessorGlobalManifestV1 {
 pub enum DataShareProcessorSpecificManifest {
```
tgeoghegan, line 200: The only thing this enum was for was dispatching method calls to either `struct DataShareProcessorSpecificManifestV1` or `struct DataShareProcessorSpecficManifestV2`. Now that there's only one struct, I think you can rename it to `struct DataShareProcessorSpecificManifest`, move the methods from the enum to the struct (eliminating the now-pointless `match` blocks) and delete the enum.

tgeoghegan (changes requested):
I don't mind simplifying the enum/struct hierarchy, but I'm concerned we will lose some strictness. We should reject a manifest that has `"format": 1` and some value for key `"peer-validation-identity"` (we don't currently have a test that exercises exactly this but the test case labeled "Unexpected top-level field" in `invalid_specific_manifest` does test this), but I think this change will accept it, so I think you'd need some manual validation after deserializing from JSON to check that a v1 manifest value conforms to the v1 schema.

facilitator/src/manifest.rs
BranLwyd: Sure -- I left the enum originally to not wipe out the versioning, but YAGNI very much applies here. Done.

BranLwyd (commented):
> I don't mind simplifying the enum/struct hierarchy, but I'm concerned we will lose some strictness. We should reject a manifest that has `"format": 1` and some value for key `"peer-validation-identity"` (we don't currently have a test that exercises exactly this but the test case labeled "Unexpected top-level field" in `invalid_specific_manifest` does test this), but I think this change will accept it, so I think you'd need some manual validation after deserializing from JSON to check that a v1 manifest value conforms to the v1 schema.

OK, done. Note that we're still a little less strict than before: we would accept a V1 manifest that specifies a "None" value for its peer_validation_identity, which previously would have been rejected. I'm OK with that.

BranLwyd:
> > I don't mind simplifying the enum/struct hierarchy, but I'm concerned we will lose some strictness. We should reject a manifest that has `"format": 1` and some value for key `"peer-validation-identity"` (we don't currently have a test that exercises exactly this but the test case labeled "Unexpected top-level field" in `invalid_specific_manifest` does test this), but I think this change will accept it, so I think you'd need some manual validation after deserializing from JSON to check that a v1 manifest value conforms to the v1 schema.
> 
> OK, done. Note that we're still a little less strict than before: we would accept a V1 manifest that specifies a "None" value for its peer_validation_identity, which previously would have been rejected. I'm OK with that.

BTW, this is a very pedantic/semantic point, but to make sure we're on the same page: the intent of this PR isn't just to merge some internal structs/enums. It's to recognize that both versions of the DSP-specific manifest could/should actually just be _one_ version, and express that in the code. This will save me from needing to test both versions in the Staging rebuild.

The reason that a new version was not required is because the only difference between the versions is the introduction of an optional field, where the prior-to-field-existing behavior is the same as the default value of the new field. The complexity of a new version number is only needed when a backwards-incompatible change must be introduced; it's preferable to introduce backwards-compatible changes to existing message versions when possible in order to avoid unnecessary complexity.

While this is somewhat more commentary than necessary for this small PR, I want to make sure you don't object to my way of thinking about things -- I'm hoping to simplify the Staging rebuild using this reasoning, so if you object, please do so now. :)

tgeoghegan:
I don't think I disagree with you philosophically about backward compatibility. The reason I put in a version key and bumped it for a minor change like this was so that I wouldn't have to think through the backward compatibility cases, at the cost of writing some extra code whose correctness I felt was easier to verify through tests.

tgeoghegan (approved):

PR #1331, opened 2022-02-01
Coalesce DSP global manifest versions 0 & 1.

BranLwyd:
Like the DSP locality-specific manifest versions, we don't actually need
two different message versions here -- we can define a single message
that can accept both versions. The one downside is that aws_account_id
is now always optional (previously it was required for V0 messages, and
disallowed for V1 messages) -- this means that our parsing is slightly
less strict. But we no longer use aws_account_id anywhere, so IMO it is
unlikely that this will matter.

This simplifies the implementation of #659.


tgeoghegan (commented):
>The one downside is that aws_account_id
>is now always optional (previously it was required for V0 messages, and
>disallowed for V1 messages) -- this means that our parsing is slightly
>less strict. But we no longer use aws_account_id anywhere, so IMO it is
>unlikely that this will matter.

Our [global manifest in prod-us](https://prod-us.manifests.isrg-prio.org/global-manifest.json) advertises `aws-account-id`:
```
{
    "format": 0,
    "server-identity": {
        "aws-account-id": 338276578713,
        "gcp-service-account-email": "prio-prod-us-sum-writer@prio-prod-us.iam.gserviceaccount.com"
    }
}
```

It's consumed by NCI's Terraform to define S3 bucket ACLs. I believe you're right that we don't consult their global manifest's `aws-account-id` anywhere, though. The exception is the `lint-manifest` subcommand on `facilitator`, which I think should continue to reject invalid v0 DSP global manifests.

That said, I see that you manually implemented a check that v0 manifest must have `aws-account-id` and v1 manifests should never have it, so I don't think we have lost any strictness in parsing.

BranLwyd:
> > The one downside is that aws_account_id
> > is now always optional (previously it was required for V0 messages, and
> > disallowed for V1 messages) -- this means that our parsing is slightly
> > less strict. But we no longer use aws_account_id anywhere, so IMO it is
> > unlikely that this will matter.
> 
> Our [global manifest in prod-us](https://prod-us.manifests.isrg-prio.org/global-manifest.json) advertises `aws-account-id`:
> 
> ```
> {
>     "format": 0,
>     "server-identity": {
>         "aws-account-id": 338276578713,
>         "gcp-service-account-email": "prio-prod-us-sum-writer@prio-prod-us.iam.gserviceaccount.com"
>     }
> }
> ```
> 
> It's consumed by NCI's Terraform to define S3 bucket ACLs. I believe you're right that we don't consult their global manifest's `aws-account-id` anywhere, though. The exception is the `lint-manifest` subcommand on `facilitator`, which I think should continue to reject invalid v0 DSP global manifests.
> 
> That said, I see that you manually implemented a check that v0 manifest must have `aws-account-id` and v1 manifests should never have it, so I don't think we have lost any strictness in parsing.

So, I'd say that this reveals that we are in an unfortunate situation: the production deployments have lagged so far behind what is committed to our repository that we can no longer create a test environment that matches the configuration of the production environment without doing the moral equivalent of rolling back whatever change got rid of us caring about `aws-account-id`, which I suppose we don't want to do.

That is fairly bad, but unless we want to go back to caring about `aws-account-id` in *our* TF, the best move at this point might be to work with NCI to update their terraform to also stop caring about `aws-account-id`.

BranLwyd:
I upated this with the result of the discussion in https://github.com/abetterinternet/prio-server/pull/1332. To address your concerns around validation, I included validation of the form "at most one of aws_account_id & gcp_service_account_id may be set at once" -- I think this is equivalent to the validation provided by the versioning mechanism without requiring explicit versioning.

tgeoghegan (approved):

PR #1332, opened 2022-02-01
Update staging terraform to match desired end state.

BranLwyd:
I _think_ what you're asking for in the initial comment to #659 is for one environment to have `use_aws = false / pure_gcp = true`, and the other to have `use_aws = true / pure_gcp = false`. Please correct me if I'm wrong about that.

### DIFF
--- a/terraform/variables/staging-facil.tfvars
+++ b/terraform/variables/staging-facil.tfvars
@@ -71,9 +71,10 @@ test_peer_environment = {
 }
 is_first                 = false
 use_aws                  = false
-workflow_manager_version = "0.6.21"
-facilitator_version      = "0.6.21"
-key_rotator_version      = "0.6.21"
+pure_gcp                 = true
+workflow_manager_version = "0.6.23"
+facilitator_version      = "0.6.23"
+key_rotator_version      = "0.6.23"
 victorops_routing_key    = "prio-staging"
 
 intake_max_age             = "75m"
--- a/terraform/variables/staging-pha.tfvars
+++ b/terraform/variables/staging-pha.tfvars
@@ -71,8 +71,9 @@ test_peer_environment = {
 }
 is_first                 = true
 use_aws                  = true
-workflow_manager_version = "0.6.18"
-facilitator_version      = "0.6.18"
+workflow_manager_version = "0.6.23"
+facilitator_version      = "0.6.23"
+key_rotator_version      = "0.6.23"
 victorops_routing_key    = "prio-staging"
 
 intake_max_age             = "75m"
@@ -80,3 +81,17 @@ default_aggregation_period = "30m"
 
 default_peer_share_processor_manifest_base_url = "storage.googleapis.com/prio-staging-facil-manifests"
 default_portal_server_manifest_base_url        = "storage.googleapis.com/prio-staging-pha-manifests/portal-server"
+
+key_rotator_schedule           = "*/10 * * * *" // once per ten minutes
+enable_key_rotation_localities = ["*"]
+batch_signing_key_rotation_policy = {
+  create_min_age   = "6480h" // 6480 hours = 9 months (w/ 30-day months, 24-hour days)
+  primary_min_age  = "6h"    // 6 hours
+  delete_min_age   = "9360h" // 9360 hours = 13 months (w/ 30-day months, 24-hour days)
+  delete_min_count = 2
+}
+
+prometheus_helm_chart_version           = "15.0.2"
+grafana_helm_chart_version              = "6.20.5"
+cloudwatch_exporter_helm_chart_version  = "0.17.2"
+stackdriver_exporter_helm_chart_version = "1.12.0"


tgeoghegan:
Not exactly. What I think we want is to be able to set `pure_gcp` on a per-locality basis. Consider that in `prod-us`, our env is `use_aws = false / pure_gcp = false`, and NCI's is `use_aws = true / pure_gcp = false` (not exactly, because NCI doesn't use our TF, but conceptually). In `prod-intl`, our env is `use_aws = true / pure_gcp = true`, and Google's is `use_aws = false / pure_gcp = true`.

I'd like us to have a single pair of staging envs (call them `staging-facil` on GCP and `staging-pha` on AWS) with localities `gondor` and `narnia`. `gondor` would be `pure_gcp = false` in both `staging-facil` and `staging-pha` and `narnia` would be `pure_gcp = true` in both `staging-facil` and `staging-pha`.

Currently, the `pure_gcp` flag gets applied to an entire env. What's missing is the ability to have a `pure_gcp = true` locality and a `pure_gcp = false` locality co-exist in the same env.

BranLwyd:
> Not exactly. What I think we want is to be able to set `pure_gcp` on a per-locality basis. Consider that in `prod-us`, our env is `use_aws = false / pure_gcp = false`, and NCI's is `use_aws = true / pure_gcp = false` (not exactly, because NCI doesn't use our TF, but conceptually). In `prod-intl`, our env is `use_aws = true / pure_gcp = true`, and Google's is `use_aws = false / pure_gcp = true`.
> 
> I'd like us to have a single pair of staging envs (call them `staging-facil` on GCP and `staging-pha` on AWS) with localities `gondor` and `narnia`. `gondor` would be `pure_gcp = false` in both `staging-facil` and `staging-pha` and `narnia` would be `pure_gcp = true` in both `staging-facil` and `staging-pha`.
> 
> Currently, the `pure_gcp` flag gets applied to an entire env. What's missing is the ability to have a `pure_gcp = true` locality and a `pure_gcp = false` locality co-exist in the same env.

I've mostly updated the terraform the way you suggest, but the global manifest is based on the global pure_gcp value (and it's not locality/ingestor-specific, so it's unclear how to respect the new locality/ingestor-specific setting). Pushed this change for visibility, I'll talk about it in the upcoming meeting if I get a chance.

tgeoghegan:
My idea for all this was each env in the staging or devenv pair would deploy two sets of manifests at different relative paths in the manifest bucket. Then, we would use a different peer data share processor manifest base URL in each locality. So let's say you have envs `facil` and `pha` that have envs `gondor` and `narnia` whose manifest buckets both contain:

```
pure_gcp/global-manifest.json // v1 global manifest
pure_gcp/gondor-ingestor-1-manifest.json // v2 specific manifest
pure_gcp/gondor-ingestor-2-manifest.json // v2 specific manifest
impure_gcp/global-manifest.json // v0 global manifest
impure_gcp/narnia-ingestor-1-manifest.json // v1 specific manifest
impure_gcp/narnia-ingestor-2-manifest.json // v1 specific manifest
```

Then, each env's `gondor` locality gets configured with `peer_share_processor_manifest_base_url` = `storage.googleapis.com/<bucket-name>/pure_gcp` and each env's `narnia` locality gets `peer_share_processor_manifest_base_url` = `storage.googleapis.com/<bucket-name>/impure_gcp`.

BranLwyd:
So I think that could work, but it will add complexity to our already-very-complex terraform to support a test-only case. (I say it adds complexity because there currently is no support for deploying to different prefixes based on the `pure_gcp` variable, nor for deploying multiple global manifests for a single environment, and we'd need to build this new functionality out such that we could trigger the functionality for the staging-us environment but definitely not our production environments or staging-intl & not our dev envs by default.)

Instead of adding complexity, what do you think about this alternative solution which IMO reduces complexity instead:
1) Merge the two versions of the global & local manifests. [we almost got there with the recent PRs, but reintroducing version-specific validation broke this]
2) Perhaps reintroduce configuration that cares about the global manifest's `aws-account-id`, if we want to test that configuration.

This solution wouldn't require multiple global manifests because there is only one version of the global manifest. (2) is required if we want to test the configuration used in our prod-us deployment -- right now it doesn't matter if `aws-account-id` is populated or not in our test environments, since we don't actually read it anyhwere. (That's true whether we merge versions or not -- even with separate versions, testing with a v0 manifest doesn't effectively test a configuration matching the `prod-us` deployment since we'll never look to `aws-account-id` with our current terraform.)

I updated the PR with example code in an attempt to make my suggestion clear. I didn't update the global manifest because that PR hasn't been merged yet, but I would make the same change I made to the locality-specific manifest: drop the version-specific validation, only validating that the version is either 0 or 1. I haven't yet tried deploying this anywhere yet because I want to make sure we're on the same page re the end goal before I start modifying environments.

terraform/modules/data_share_processor/data_share_processor.tf
```diff
@@ -449,28 +449,11 @@ output "aggregate_queue" {
 }
 
 output "specific_manifest" {
```
tgeoghegan, line 451: What do we use this output for at all anymore? So that we can provide the default specific manifest template to `key-rotator`?

tgeoghegan (commented):
I agree with your desire to minimize complexity. I think with this approach, we'll get continuous testing of all the cross-cloud authentication flows we care about, though our validation of different manifest formats will be a hair less rigorous than it could be. I think I can live with that.

BranLwyd:
I updated this PR with the outcome of the discussion from https://github.com/abetterinternet/prio-server/pull/1332. In an attempt to address your concerns around validation, I keep validation of the form "at most one of aws_account_id & gcp_service_account_id can be set" -- I think this is equivalent to the pre-existing validation, without pegging things to a specific version number.

tgeoghegan (approved):

terraform/modules/data_share_processor/data_share_processor.tf
BranLwyd: Yep: `key-rotator` uses this as the "template" from which it can generate full manifests for (locality, ingestor) pairs which do not already have one. [I did this to make sure the generated manifest matches as closely as possible to what was being generated by `deploy-tool`.]

BranLwyd:
I've deployed this change in my dev environment, I'm going to let it bake at least overnight before I move forward.

I checked NCI's TF config and I'm confident the changes to the global manifest will not cause problems with their deploy. Unfortunately I can't check the int'l deployment. :-/

BranLwyd:
@abetterinternet/ppm-committers The new staging is currently deployed based on this branch, and appears to be working. I am going to wait to merge and roll the Go 1.17.7 update into this -- if there are any final issues to be resolved, please look soon.

PR #1333, opened 2022-02-02
Fix bug around task-deadline extension.

### LINKED ISSUE #319: facilitator: deadline being extended too often
The deadline extension logic in #309 is a little over eager: it waits until 10 minutes have elapsed to extend the ack deadline or visibility timeout (depending on cloud platform) but will then do so at every callback, which in the case of aggregation tasks (the only ones that actually take >10 minutes to run) means after every aggregated batch. This is basically harmless, as all it does is push the deadline out a little farther, but it causes unnecessary traffic to PubSub. We should fix this so that it only extends the deadline 10 minutes after the previous deadline extension.

BranLwyd:
Previously, after the first time we extended a task's deadline, we would
attempt to extend the deadline again every time we checked if the
deadline should be updated. Now, every time we extend a task's deadline,
we "reset" the starting point so that we only attempt to reset the
deadline again once enough time has passed.

I had to switch to "stable" channel to get a version of Rust (1.58) recent enough to have const duration-handling functions.

Closes #319.

### DIFF
--- a/facilitator/rust-toolchain.toml
+++ b/facilitator/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.56.1"
+channel = "stable"
 components = [ "rustc", "rustfmt", "clippy" ]
--- a/facilitator/src/task.rs
+++ b/facilitator/src/task.rs
@@ -9,7 +9,7 @@ use slog::{Key, Record, Serializer, Value};
 use std::{
     fmt,
     fmt::{Debug, Display},
-    time::Duration,
+    time::{Duration, Instant},
 };
 use uuid::Uuid;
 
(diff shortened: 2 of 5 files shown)


facilitator/rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.56.1"
+channel = "stable"
```
tgeoghegan, line 2: I think we should revisit whether `rust-toolchain.toml` is a good idea. We should be running tests in GitHub CI with the same Rust toolchain version used to build containers. The version here is used for the former, the container image in the `prio-facilitator` Dockerfile sets the latter. The difficulty is that dependabot keeps our Dockerfile up to date, but doesn't appear to know about `rust-toolchain.toml`. We need to research what the current best practices are for specifying a minimum required Rust version that also plays nice with CI, building containers and dependabot.

Anyway, that's not really relevant to this PR, and using stable is strictly better than pinning to 1.56.1, I think.

facilitator/src/task.rs
```diff
@@ -43,17 +43,30 @@ pub trait TaskQueue<T: Task>: Debug + DynClone + Send + Sync {
...
+        // call from the if statement below to the declaration here.
+        const DURATION_BEFORE_REFRESH: Option<Duration> =
+            DEADLINE_INCREMENT.saturating_mul(9).checked_div(10);
+
+        if last_refresh.elapsed() >= DURATION_BEFORE_REFRESH.unwrap() {
+            return self
+                .extend_task_deadline(handle, &DEADLINE_INCREMENT)
+                .map(|_| {
+                    *last_refresh = Instant::now();
+                });
```
tgeoghegan, line 69: I'm comfortable with this, but I wonder if it becomes a little tidier if you return `Result<Instant>` from this method instead of mutating `last_refresh`.

tgeoghegan (approved):

facilitator/rust-toolchain.toml
BranLwyd: Agreed, though I'd bet a strategy of "pin everything to stable, until that causes a problem" would work out OK.

facilitator/src/task.rs
BranLwyd: It pushes the (extremely minor) complexity of updating `last_refresh` out to the caller, but it also gets rid of an out-paramemter, so on balance it's probably a good change.

I guess if we cared a lot, we could create a little `TaskRefresher` struct and put the state & the logic inside it, but I think that might be overkill for how simple this functionality is.

facilitator/rust-toolchain.toml
divergentdave: A lot of projects specify a "minimum supported Rust version", or MSRV, colloquially. (This is largely a convention, but RFC 2495 will add a Rust version field to Cargo.toml) Some libraries go further, and make guarantees about for how many months/years their MSRV will trail the stable channel. There are differing opinions on whether bumping the MSRV implies different degrees of SemVer incompatibility. (As such, a dependency update may necessitate an MSRV bump)

In our situation, with the facilitator server being a binary, and libprio-rs presumably having few external consumers, we can probably afford to be more liberal with changing our MSRV. To hold ourselves to MSRV compatibility, we would want to at least run a check build in CI on both our specified MSRV and the latest stable channel. (and perhaps nightly) The main downside of attempting to support a wide range of Rust versions is that conflicts between lints or Clippy lints in the new version versus available APIs in the old version mean having to carry `#[allow()]` annotations in problem areas, or TODO comments where we would like to use a newer API, until the next MSRV bump.

tgeoghegan (approved):

PR #1342, opened 2022-02-07
Add `CODEOWNERS` file

tgeoghegan:
Once we have set `CODEOWNERS`, we can configure the GitHub repo to require review from code owners, which IIUC should automatically tag all members of [ppm-committers](https://github.com/orgs/abetterinternet/teams/ppm-committers/members) on PRs to this repo.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @abetterinternet/ppm-committers


tgeoghegan:
My plan is to ask everyone currently in that group whether they want to get emailed every time we post a PR. If they don't, then I argue they don't belong in a group called `ppm-committers`, and we can remove them.

divergentdave (approved):
LGTM. We could move the file to `.github/CODEOWNERS` to keep the root clean.

PR #1346, opened 2022-02-09
Terraform fixes

divergentdave:
This fixes a few issues I encountered while setting up my development environment.

- The `google-beta` provider needs to be configured in addition to the `google` provider. Without this, the two resources that use it were failing for want of a region on my computer.
- The bootstrap make targets needs its `-target` flags to be scoped down further, so that we can get the first manifests created, without touching anything that depends on reading them via the `http` provider.
- I added resources to flip on two GCP services, and added dependencies to GCP services in two cases where I saw error messages on the first run.

### DIFF
--- a/terraform/Makefile
+++ b/terraform/Makefile
@@ -104,7 +104,7 @@ plan-bootstrap: prep ## Show what terraform thinks it will do
 		-var-file="$(VARS)" \
 		-target=module.manifest_gcp \
 		-target=module.manifest_aws \
-		-target=module.fake_server_resources \
+		-target=module.fake_server_resources[0].google_storage_bucket_object.global_manifest \
 		-target=module.portal_server_resources \
 		-target=module.gke \
 		-target=module.eks \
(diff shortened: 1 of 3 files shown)


BranLwyd (approved):
I'm not totally sure why the `google-beta` provider is necessary / why Tim or I didn't need to enable it, but I don't think enabling it has much chance to cause harm, so LGTM.

divergentdave:
It's necessary for [google_iam_workload_identity_pool](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/iam_workload_identity_pool) resources, as well as [google_iam_workload_identity_pool_provider](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/iam_workload_identity_pool_provider). (The cluster_bootstrap directory needs it for [google_artifact_registry_repository](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/artifact_registry_repository), but it's already got the same provider configuration stanzas for both.)

As for why you didn't get the error, I think it's that I opted to not configure a default region during `gcloud init`. This lives in configuration parameter `compute/region`.

PR #1348, opened 2022-02-09
terraform: Use aws provider's allowed_account_ids

divergentdave:
When this new variable is set, the `aws` provider will refuse to take any actions if it is given credentials to a mismatched account.

For now, I have added a commented-out declaration in each .tfvars file, with an example account ID, assuming we don't want to check in any actual account IDs.

### DIFF
--- a/terraform/cluster_bootstrap/main.tf
+++ b/terraform/cluster_bootstrap/main.tf
@@ -26,6 +26,19 @@ The AWS region in which resources should be created.
 DESCRIPTION
 }
 
+variable "allowed_aws_account_ids" {
+  type        = list(string)
+  default     = null
+  description = <<DESCRIPTION
+A list of AWS account IDs that are allowed to be used. Set this in an
+environment's variables to prevent accidentally using the environment with
+the wrong AWS configuration profile. This variable may be omitted, in which
+case no AWS account ID check is performed. If a list of account IDs is
+given, the provider will check the account of the active credentials, and
+return an error if it is not listed here.
+DESCRIPTION
+}
+
 variable "use_aws" {
   type        = bool
   description = <<DESCRIPTION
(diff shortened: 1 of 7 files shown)


tgeoghegan:
I'm wondering whether there is any risk to disclosing our AWS account numbers in a public repository like this. Admittedly, we already expose them in the manifests which are publicly available so it's no worse, and the safety from `allowed_aws_account_ids` is really nice.

BranLwyd:
> I'm wondering whether there is any risk to disclosing our AWS account numbers in a public repository like this. Admittedly, we already expose them in the manifests which are publicly available so it's no worse, and the safety from `allowed_aws_account_ids` is really nice.

IMO, that ship has sailed due to the manifests: looking to the NIH's config, their `prod-us.tfvars` shows that the ISRG's manifest base URL is `prod-us.manifests.isrg-prio.org`. Navigating to that URL provides a listing of all hosted manifests, so an attacker can enumerate manifests from there. I can't seem to list other parties' manifest buckets, but they could pretty easily figure out the manifest names by reading the terraform and/or experimenting.

divergentdave:
That makes sense. I'll crank in the IDs and revise the instructions.

tgeoghegan:
> IMO, that ship has sailed due to the manifests: looking to the NIH's config, their `prod-us.tfvars` shows that the ISRG's manifest base URL is `prod-us.manifests.isrg-prio.org`. Navigating to that URL provides a listing of all hosted manifests, so an attacker can enumerate manifests from there. I can't seem to list other parties' manifest buckets, but they could pretty easily figure out the manifest names by reading the terraform and/or experimenting.

Yes, I agree, all these parameters are public anyway. This is worth it for the safety. Though in future projects, we should think about keeping .tfvars files (which may contain sensitive values like team member names or email addresses) in a private repo separate from the Terraform modules.

PR #1350, opened 2022-02-10
Facilitator: main function refactor

divergentdave:
This makes two neighboring changes to `main()`. First, I pulled out construction of the `clap::App` into a separate function. This will be the first step of migrating to the next major version of `clap`, but meanwhile, it's convenient to be able to fold it. Second, I noticed that when certain jobs bail on a fatal 404 error, they panic on the way out because the global logger gets torn down before a `ureq::Stream` gets dropped, and attempts to log. I fixed this by splitting `main()` into two functions and reporting errors before the outer one returns. (You can see this behavior by running `facilitator validate-ingestion-sample-worker` with `--pha-manifest-base-url=https://httpbin.org/status/404`)

### DIFF
--- a/facilitator/src/bin/facilitator.rs
+++ b/facilitator/src/bin/facilitator.rs
@@ -970,19 +970,7 @@ fn app() -> App<'static, 'static> {
         )
 }
 
-fn main() -> Result<(), anyhow::Error> {
-    let matches = app().get_matches();
-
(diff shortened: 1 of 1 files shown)


BranLwyd (approved):
Thanks! This is a nice change -- `facilitator.rs` has grown quite large so any refactoring/cleanup is very welcome.

PR #1385, opened 2022-02-15
facilitator: structured error handling

divergentdave:
This introduces enum-based error handling in the IDL module. It eliminates all but one use of anyhow in `src/idl.rs`, and effectively moves several variants of the crate-level `Error` to this module.

This doesn't introduce any behavior changes for packet decryption failures, etc. yet. I figured this module would be good to start with, as it has the highest density of `anyhow::Context::context()` calls. In the interest of eating an elephant one bite at a time, I'm putting this up for early feedback, and we can treat it as a draft if we like.

### DIFF
--- a/facilitator/src/aggregation.rs
+++ b/facilitator/src/aggregation.rs
@@ -127,7 +127,8 @@ impl<'a> BatchAggregator<'a> {
                 self.trace_id,
                 &self.logger,
             )
-            .read(&self.ingestion_transport.transport.batch_signing_public_keys)?;
+            .read(&self.ingestion_transport.transport.batch_signing_public_keys)
+            .map_err(|e| Error::AnyhowError(e.into()))?;
             let mut peer_validation_batch_reader = BatchReader::new(
                 Batch::new_validation(self.aggregation_name, batch_id, batch_date, !self.is_first),
                 &*self.peer_validation_transport.transport,
@@ -140,7 +141,8 @@ impl<'a> BatchAggregator<'a> {
                     .set_metrics_collector(&collector.peer_validation_batches_reader_metrics);
             }
             let (peer_validation_hdr, peer_validation_packets) = peer_validation_batch_reader
-                .read(&self.peer_validation_transport.batch_signing_public_keys)?;
+                .read(&self.peer_validation_transport.batch_signing_public_keys)
+                .map_err(|e| Error::AnyhowError(e.into()))?;
 
             if servers.is_none() {
                 // Ideally, we would use the encryption_key_id in the ingestion packet
@@ -241,7 +243,7 @@ impl<'a> BatchAggregator<'a> {
                 },
                 invalid_uuids.into_iter().map(|u| InvalidPacket { uuid: u }),
             )
-            .map_err(Error::AnyhowError)
+            .map_err(|e| Error::AnyhowError(e.into()))
     }
 
     /// Aggregate the batch for the provided batch_id into the provided server.
--- a/facilitator/src/aggregation.rs
+++ b/facilitator/src/aggregation.rs
@@ -5,7 +5,7 @@ use crate::{
     logging::event,
     metrics::AggregateMetricsCollector,
     transport::{SignableTransport, VerifiableAndDecryptableTransport, VerifiableTransport},
-    BatchSigningKey,
+    BatchSigningKey, ErrorClassification,
 };
 use chrono::NaiveDateTime;
 use prio::{
@@ -42,6 +42,29 @@ pub enum AggError {
     Intake(IntakeError),
     #[error("invalid verification message: {0}")]
     InvalidVerification(TryFromIntError),
+}
+
+impl ErrorClassification for AggError {
+    fn is_retryable(&self) -> bool {
+        match self {
+            // If the batches and dates in this task are empty, retrying the task won't help.
+            AggError::EmptyBatchesAndDates => false,
+            // These indicate an issue with the peer server's validation message. Retries are not
+            // necessary in this case.
+            AggError::HeaderMismatch(_) | AggError::InvalidVerification(_) => false,
+            // libprio-rs errors may be due to getrandom failures.
+            AggError::PrioSetup(_) | AggError::PrioAccumulate(_) | AggError::Validation(_, _) => {
+                true
+            }
+            // This error is only reachable if we try to aggregate with zero servers, which
+            // indicates a configuration error with this process. Retry again later.
+            AggError::ValidationUnknown(_) => true,
+            // Dispatch to wrapped error types.
+            AggError::BatchRead(e) => e.is_retryable(),
+            AggError::BatchWrite(e) => e.is_retryable(),
+            AggError::Intake(e) => e.is_retryable(),
+        }
+    }
 }
 
 pub struct BatchAggregator<'a> {
--- a/facilitator/src/aggregation.rs
+++ b/facilitator/src/aggregation.rs
@@ -285,7 +285,7 @@ impl<'a> BatchAggregator<'a> {
             match ingestion_packet.generate_validation_packet(servers) {
                 Ok(p) => ingestion_and_own_validation_packets.push((ingestion_packet, p)),
                 Err(e) => {
-                    if let IntakeError::PacketDecryptionError(packet_uuid) = e {
+                    if let IntakeError::PacketDecryption(packet_uuid) = e {
                         // This ingestion batch contains a packet that can't be
                         // decrypted. Ignore the entire ingestion batch, on the
                         // assumption that our peer will do the same, since we
--- a/facilitator/src/aggregation.rs
+++ b/facilitator/src/aggregation.rs
@@ -1,6 +1,9 @@
 use crate::{
     batch::{Batch, BatchReadError, BatchReader, BatchWriteError, BatchWriter},
-    idl::{IngestionDataSharePacket, IngestionHeader, InvalidPacket, SumPart, ValidationPacket},
+    idl::{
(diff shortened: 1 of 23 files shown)


facilitator/src/intake.rs
```diff
@@ -18,6 +20,18 @@ use slog::{debug, info, o, Logger};
 use std::{collections::HashMap, iter::Iterator};
 use uuid::Uuid;
 
+#[derive(Debug, thiserror::Error)]
+pub enum IntakeError {
+    #[error(transparent)]
+    Idl(IdlError),
+    #[error("packet decryption failure for packet {0}")]
+    PacketDecryptionError(Uuid),
```
tgeoghegan, line 28: nit: I prefer that variants on an error enum not end in `Error`. Otherwise you get `IntakeError::PacketDecryptionError` which unnecessarily repeats `Error` (stutters, as some Go guidelines put it).

tgeoghegan (commented):
I think the approach of introducing an error enum in each module, then putting variants on `lib::Error` that wrap each module's error is good. But I'm not sure if it's a good idea for functions/methods in each module to all return `lib::Error`. My rationale is that the error type returned by a function is also implicitly documentation of what can go wrong with a method, and thus what cases a caller should be prepared to handle. If I call `BatchIntaker::generate_validation_share` and I get an error `Err(lib::Error)`, then it looks like I need to be ready to handle any of the variants on that enum, but in fact I will only ever encounter `Err(Error::Intake(IntakeError))` and only need to worry about variants of `IntakeError`. So I favor functions in a module returning that module's error enum, and then using variants tagged with `#[from]` to make it easy to convert between error types.

I'll also observe that what you've got here has a slightly awkward discrepancy between how this is done in module `idl` (functions return `Result<T, IdlError>` and modules `intake` and `aggregate` (functions return `Result<T, Error>`).

That said, the changes you've got achieve the actual objective (structured error handling that lets us avoid unnecessary retries) and for want of a clear best practice in Rust for error handling, my observations are really just rooted in my opinion, so I think as the person making the changes, you have the prerogative of doing this whichever way you prefer.

facilitator/src/intake.rs
divergentdave: Ah thanks, forgot this one when I moved it.

divergentdave:
I agree, the `crate::Error` in return types in the intake/aggregate modules is mainly there for the purposes of migration. I'd like to see all those specialized, but I'm not sure how much will be done in this go-around. From eyeballing the `anyhow!` and `context` calls, the error types in these modules will need a good number more cases added, then these error types can be replaced. This second commit gets me to the point where the packet decryption error can pass all the way up to the queue ACK/NACK, without being wrapped in an `anyhow` error.

divergentdave:
This is complete and ready for review now.

BranLwyd (commented):
A couple of meta-comments:

1) In general, prefer smaller PRs over larger PRs; this makes it easier on your reviewers & generally leads to higher-quality code review. For this PR, I think the dead-letter forwarding functionality, the terraform changes, and each module's error-structuring could have been separate PRs; though, natch, some of the module error-structuring PRs would have been dependent on one another. (I don't recommend trying to factor into separate PRs now, but good to keep in mind for future changes.)

2) I think this PR may have arguably gone too far: the purpose of structuring errors in this case is to allow an "is-error-retryable" signal to be propagated. But several of the structured error types have all error types retryable/not-retryable, and several others have many more types than is required. Again, I don't necessarily recommend going back and trying to change it now, but I think the same effect could have been reached with less effort/churn, without (IMO) leaving the code messy/unmaintainable.

terraform/modules/pubsub/pubsub.tf
```diff
@@ -98,5 +99,6 @@ output "queue" {
     topic             = google_pubsub_topic.task.name
     subscription_kind = "gcp-pubsub"
     subscription      = google_pubsub_subscription.task.name
+    dead_letter_topic = google_pubsub_topic.dead_letter.name
```
tgeoghegan, line 102: IIUC, this means we will publish unretryable errors to the existing PubSub dead letter topic (or dead letter SNS topic on the AWS side). I don't think this is what we want, because those dead letters will trigger our existing alerting rules for dead letters. I think dead letters from unretryable failures like packet decryption failure should go to a different dead letter topic and queue so that we can be aware of them but not alert.

tgeoghegan (changes requested):
I think the error enums you've introduced will do the job, but I also think they are overkill. Notice that while you've added dozens of error variants, we really only care about a single new bit of information: is this error retryable or not. Besides that, we don't behave any differently in the face of `AggError::HeaderMismatch` vs. `AggError::InvalidVerification` or anything else.

Previously, we had `anyhow::Error`, which we can think of as wrapping a string and a `Box<dyn 'static + Error>` (the underlying problem). I suspect we could have achieved our goal here with an error type that wraps a string, `Box<dyn 'static + Error>` and a boolean indicating whether it's retryable.

It's telling that there aren't dozens of new tests to go with the dozens of new error variants. That means that we can't guarantee that we return the right error object in any situation. If this stuff were in an API surface we were exporting, I'd argue that we should delete it, because our API would be suggesting contracts that our tests don't prove we uphold.

All that being said: now that you've expended the effort, there's no downside to having such pleasantly descriptive and structured errors, and since there are no clients outside of `prio-server` depending on the structure of the errors, it's OK if we change them unpredictably in the future. I just wanted to provide a perspective.

facilitator/src/aggregation.rs
```diff
@@ -16,9 +16,57 @@ use slog::{info, o, warn, Logger};
...
+    #[error("ingestion header does not match peer validation header. {0}")]
+    HeaderMismatch(String),
+    #[error("error setting up Prio server: {0}")]
+    PrioSetup(prio::server::ServerError),
+    #[error("error accumulating shares: {0}")]
+    PrioAccumulate(prio::server::ServerError),
+    #[error("trace id {1} failed to validate packets: {0}")]
+    Validation(prio::server::ServerError, Uuid),
+    #[error("trace id {0} failed to validate packets: unknown validation error")]
+    ValidationUnknown(Uuid),
```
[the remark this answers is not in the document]
divergentdave: That's correct, `aggregate_worker()` does log these errors with the trace ID in a key-value pair. I'll take it out, since it's redundant.

facilitator/src/aws_credentials.rs
```diff
@@ -683,28 +709,27 @@ fn get_caller_identity_token_at_time(
...
+        host.parse().map_err(|_| StsError::BadHeaderHost)?,
     );
     headers.insert(
         HeaderName::from_static(AMAZON_DATE_HEADER),
         request_time
             .format(LONG_DATETIME)
             .to_string()
             .parse()
-            .context("failed to parse formatted date as header value")?,
+            .map_err(|_| StsError::BadHeaderDate)?,
```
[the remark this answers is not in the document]
divergentdave: Currently, `InvalidHeaderValue` from the http crate unconditionally prints "failed to parse header value" on these errors. I think I'll add it to the enum variants just so it's not different anymore, and in case of future improvements from http.

facilitator/src/transport.rs
```diff
@@ -15,9 +14,28 @@ use std::{
...
+    #[error(transparent)]
+    Local(#[from] local::FileError),
+    #[error(transparent)]
+    Gcs(#[from] GcsError),
+    #[error(transparent)]
+    S3(#[from] S3Error),
+    #[error("an error occurred while cancelling after another error: {original}, {cancellation}")]
+    Cancellation {
+        original: Box<TransportError>,
+        cancellation: Box<TransportError>,
```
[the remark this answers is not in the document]
divergentdave: I was getting a Clippy warning about a large difference between sizes of variants elsewhere, (i.e., `BatchReadError::Transport`) but here in particular I need the boxes because a `TransportError` can't contain another `TransportError` in-line.

terraform/modules/pubsub/pubsub.tf
divergentdave: I'll work on the new topic/queue next. How about "...-rejected" as a name?

tgeoghegan: That sounds good. I think we should also put some alerts on these new queues, but we can discuss that separately from this PR.

BranLwyd: +1 to putting this in another PR, but I'd ideally like to see alerts before we release this to production: I totally agree that dropping the occasional packet decryption error is fine, but if we see a large volume of these errors I'd like to know (since this might effectively mean we are currently dropping data on the floor in an unrecoverable way).

tgeoghegan (approved):
Changes LGTM, but I agree with @BranLwyd that we should devise a plan for alerting on non-retryable/client errors before deploying.

terraform/modules/pubsub/pubsub.tf
BranLwyd: I thought about this a little more; I'm still worried about the risk of us forgetting that `is_retryable = false` also means that we don't alert on the failed work item (unless it is high volume). But this is fairly hypothetical--I definitely don't want to complicate an already-large PR with this.

Would you mind adding to the `is_retryable` doc string that this also effectively also changes DLQ routing/alerting? This is sort of a "layering violation" but I think it's important enough that it should be documented, and it may not be immediately obvious to someone who is "just adding a new error case" in the future.

BranLwyd (approved):
LGTM to me too, but please update the `ErrorClassification` / `is_retryable` doc strings to note the DLQ change. IMO, `is_retryable` might be a too-generic name vs the doc string ("This trait captures whether a given error is due to corruption in client-provided data") -- I think either the doc string should be made more generic, or the `is_retryable` name should be made much more specific. And either way IMO we should capture that this will have an effect on alerting.

PR #1387, opened 2022-02-16
Fix flapping of fake-ingestor identity in sum-part-writer bucket.

BranLwyd:
This was due to main.tf's
data_share_processors_to_sum_part_writer_token_creator interfering with
the fake_server_resources.tf's
integration_test_identity_to_my_sum_part_bucket_writer_token_creator. I
fix this by specifying the membership in
data_share_processors_to_sum_part_writer_token_creator directly; the
downside is this "leaks" some integration-test configuration outside of
fake_server_resources.tf.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -712,7 +712,10 @@ resource "google_service_account_iam_binding" "data_share_processors_to_sum_part
   count              = var.use_aws ? 0 : 1
   service_account_id = google_service_account.sum_part_bucket_writer.name
   role               = "roles/iam.serviceAccountTokenCreator"
-  members            = [for v in module.data_share_processors : "serviceAccount:${v.gcp_service_account_email}"]
+  members = concat(
+    [for v in module.data_share_processors : "serviceAccount:${v.gcp_service_account_email}"],
+    local.is_env_with_ingestor ? ["serviceAccount:${module.fake_server_resources[0].gcp_service_account_email}"] : []
+  )
 }
 
 # GCP services we must enable to use Workload Identity Pool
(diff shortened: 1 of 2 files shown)


tgeoghegan (approved):
Good find! What you've got solves the problem, but if you want to avoid the minor issue of "leaking" information from `fake_server_resources`, there's another way. The issue here is that [`google_service_account_iam_binding` is authoritative](https://registry.terraform.io/providers/hashicorp/google/3.33.0/docs/resources/google_service_account_iam) for a role, so it overwrites the changes made by the `google_service_account_iam_member` resource in `fake_server_resources`. We could change `resource "google_service_account_iam_binding" "data_share_processors_to_sum_part_bucket_writer_token_creator"` to be a `google_service_account_iam_member` with a `for_each` over `module.data_share_processors`. Then each of those `google_service_account_iam_member` resources would co-exist happily with the `google_service_account_iam_member` in `fake_server_resources`.

tgeoghegan (approved):

BranLwyd:
> Good find! What you've got solves the problem, but if you want to avoid the minor issue of "leaking" information from `fake_server_resources`, there's another way. The issue here is that [`google_service_account_iam_binding` is authoritative](https://registry.terraform.io/providers/hashicorp/google/3.33.0/docs/resources/google_service_account_iam) for a role, so it overwrites the changes made by the `google_service_account_iam_member` resource in `fake_server_resources`. We could change `resource "google_service_account_iam_binding" "data_share_processors_to_sum_part_bucket_writer_token_creator"` to be a `google_service_account_iam_member` with a `for_each` over `module.data_share_processors`. Then each of those `google_service_account_iam_member` resources would co-exist happily with the `google_service_account_iam_member` in `fake_server_resources`.

I was wary of this approach because I believed that a lack of an authoritative resource would lead to us leaving permissions around unnecessarily--i.e. if we ever removed one of these `google_service_account_iam_member` resources from our TF, the membership it created would hang around. But a small experiment shows that `google_service_account_iam_member` cleans up after itself (i.e. deletes the membership it creates when the resource is deleted). I think this solution is better all-around.

PR #1420, opened 2022-02-25
Fix failure rate alerting, new alert for rejecting

divergentdave:
This fixes an issue with the expressions used for failure rate alerting. It also introduces a new value of "rejected" for the label "status", on the "tasks_finished" metrics, for tasks that result in a non-retryable error. We apply the "rejected" value when we divert the task to a separate rejected queue, rather than let it reach the dead letter queue. A new alert is introduced to monitor the fraction of tasks being rejected, and alert if it is too high.

~~Edit: deploy note - this will require deleting the secret monitoring/prometheus-alertmanager-config and running Terraform to re-populate it~~

### DIFF
--- a/facilitator/src/bin/facilitator.rs
+++ b/facilitator/src/bin/facilitator.rs
@@ -1732,14 +1732,18 @@ where
     let result = batch_intaker.generate_validation_share(callback);
 
     if let Some(collector) = metrics_collector {
-        match result {
+        match &result {
             Ok(()) => collector
                 .intake_tasks_finished
                 .with_label_values(&["success", aggregation_id])
                 .inc(),
-            Err(_) => collector
+            Err(e) if e.is_retryable() => collector
                 .intake_tasks_finished
                 .with_label_values(&["error", aggregation_id])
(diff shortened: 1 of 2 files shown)


BranLwyd (approved):
> Edit: deploy note - this will require deleting the secret monitoring/prometheus-alertmanager-config and running Terraform to re-populate it

For my own edification, why is this necessary? When I added the sum-validation-failure alert I don't think this was required; is it because we're modifying the expression for an existing alert?

tgeoghegan (approved):

divergentdave:
I read that in docs/monitoring.md, but looking at it closer, that was referring to the alertmanager configuration file, not the alert rules, which go into the prometheus Helm release's values. Nevertheless, applying a change without deleting the secret didn't make an alert change show up right away, but doing both did. Perhaps it works as a side-effect of kicking some configuration reloader, or maybe I wasn't patient enough.

divergentdave:
Tested it again, and waiting longer is sufficient to update alerting rules, no secret deletion required.

PR #1421, opened 2022-02-25
During aggregation, require ingestion header parameters to match.

BranLwyd:
Previously, we didn't check, and would happily aggregate even if some of
the parameters didn't match--but at a higher level, data with different
parameters can't be meaningfully aggreagated, so this produced
non-useful output. All we can do is notice & complain, but we should do
so rather than producing bad output.

### DIFF
--- a/facilitator/src/aggregation.rs
+++ b/facilitator/src/aggregation.rs
@@ -159,7 +159,7 @@ impl<'a> BatchAggregator<'a> {
         let mut included_batch_uuids = Vec::new();
 
         let mut servers = None;
-        let mut ingestion_header: Option<IngestionHeader> = None;
+        let mut last_ingestion_header: Option<IngestionHeader> = None;
         for (batch_id, batch_date) in batch_ids_and_dates {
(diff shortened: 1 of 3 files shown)


facilitator/src/aggregation.rs
```diff
@@ -157,9 +159,9 @@ impl<'a> BatchAggregator<'a> {
         let mut included_batch_uuids = Vec::new();
 
         let mut servers = None;
-        let mut ingestion_header = None;
+        let mut ingestion_header: Option<IngestionHeader> = None;
```
divergentdave, line 162: nit: could we rename this to last_ingestion_header?

divergentdave (approved):

facilitator/src/aggregation.rs
BranLwyd: Sure. (though this could also be called `arbitrary_ingestion_header` -- the implementation currently chooses the last one, but we could just as easily use *any* of the ingestion headers we've seen. the expectation is that the fields we read match for all of the ingestion headers, which is now validated with this PR)

tgeoghegan (approved):

PR #1435, opened 2022-03-01
terraform: include status of working dir. in state

divergentdave:
This adds a data source to both Terraform main modules to shell out to git in order to record the checked out commit. The commit hash will be stored in the state, and I included it in an output from the module for convenience.

Sample output:

```diff
Changes to Outputs:
+ git-describe = "52972e5-dirty"
```

What do we think, is this worth having?

(Note that terraform/.terraform.lock.hcl already contains the `external` provider, as it is already used in `data_share_processor`, without inclusion in `required_providers`)

### DIFF
--- a/terraform/cluster_bootstrap/.terraform.lock.hcl
+++ b/terraform/cluster_bootstrap/.terraform.lock.hcl
@@ -20,6 +20,26 @@ provider "registry.terraform.io/hashicorp/aws" {
   ]
 }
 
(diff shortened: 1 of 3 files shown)


BranLwyd (approved):
*shrug* I can't think of a downside, and having more info is better than having less, so LGTM.

tgeoghegan (approved):
This also means we depend on git being installed, which should be fine even if/when we run this sort of thing in a CI job.

PR #1436, opened 2022-03-02
terraform: configure log retention period

### LINKED ISSUE #1391: prio-prod-intl logs are retained forever
Our logs for `prio-prod-intl`, and likely for all of our AWS-based environments, are currently retained forever. Our log volume is not overly high so this is not a critical issue, but we should set some limit to avoid unbounded log growth.

Unfortunately, I think this will require doing a "terraform import" to bring in the fix without conflicting with the already-existing log group.

Instructions for configuring retention period for EKS log groups: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster#enabling-control-plane-logging

divergentdave:
This fixes #1391. We can hold off on merging this until we want to deal with the manual Terraform migration required, (or combine it with other such changes) but I figured I may as well get it written up and tested.

This change will require importing the log group that EKS already created, to keep Terraform from trying to create it again and finding that its name is already in use. For example:

Run at least one makefile target to make sure that cluster_bootstrap/.terraform/terraform.tfstate points to the correct backend bucket for this environment.
`$ ENV=<environment name> make plan-cluster-bootstrap`
Import the log group resource. (The easiest way to get the log group's name is to attempt to apply once, and look at the log group name in the error message)
`$ ENV=<environment name> terraform -chdir=cluster_bootstrap import -var-file=../variables/<environment name>.tfvars -input=false module.eks[0].aws_cloudwatch_log_group.cluster_logs /aws/eks/prio-<environment name>/cluster`
Finally, apply the changes. This should add a prio-env tag to the log group, and set its retention period.
`$ ENV=<environment name> make apply-cluster-bootstrap`

### DIFF
--- a/terraform/cluster_bootstrap/modules/eks/eks.tf
+++ b/terraform/cluster_bootstrap/modules/eks/eks.tf
@@ -233,6 +233,13 @@ resource "aws_eks_cluster" "cluster" {
       key_arn = aws_kms_key.etcd_secrets.arn
     }
   }
+
+  depends_on = [aws_cloudwatch_log_group.cluster_logs]
+}
+
+resource "aws_cloudwatch_log_group" "cluster_logs" {
+  name              = "/aws/eks/${var.resource_prefix}/cluster"
+  retention_in_days = 7
 }
 
 data "tls_certificate" "cluster" {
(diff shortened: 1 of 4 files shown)


divergentdave:
I added another commit to set up retention periods on the log groups created by Fluent Bit. My understanding of the docs is that this setting will only have an effect on newly created log groups. Existing log groups will have to have their retention period lowered manually, but at least there aren't any Terraform conflicts.

tgeoghegan:
@divergentdave please merge this if you'd like it to be included in today's staging deploys.

PR #1469, opened 2022-03-18
Update .terraform.lock.hcl.

BranLwyd:
These diffs are introduced just by running `make plan`, so to avoid
deploying with a "dirty" commit hash, I'd like to go ahead and
officially commit these.


divergentdave (approved):
Perhaps we should drop the "~>" operator from our constraints, and rely on dependabot surfacing patch version updates. Otherwise, this will keep happening from time to time.

PR #1498, opened 2022-03-22
Add alert rules for unhealthy containers

divergentdave:
This adds two partly overlapping alerting rules using `kube-state-metrics` data. The first checks for containers that restart repeatedly, and the second checks for containers that are waiting for one reason for an extended period of time. These will catch `CrashLoopBackOff` and `ImagePullBackOff` conditions, as well as `restartPolicy: Always` containers that exit with a success code quickly.


tgeoghegan (approved):

tgeoghegan:
This looks good for `prio-server` but this strikes me as the kind of learning we should take forward into Janus. We're not yet set up for monitoring and alerts there, but could you file an issue, maybe on `janus-ops`, noting that we should use this kind of alert in Janus so we don't forget?

PR #1611, opened 2022-04-27
Upgrade rand_core

divergentdave:
This pulls in a fix for RUSTSEC-2021-0023. `rand_core v0.6.1` was a transitive dependency of `elliptic-curve` and other crates, but we only used that crate for key parsing, not encryption or key generation.


BranLwyd (approved):

BranLwyd:
Based on your analysis, I think we are not affected by the issue with `rand_core`, and picking up this change with the next weekly release is acceptable. Please let me know if I'm misunderstanding you or you disagree with this conclusion.

PR #1660, opened 2022-05-09
terraform: deploy 0.6.37 to prod envs

tgeoghegan:


BranLwyd (approved):
Q (just curious): why doesn't this target `main` initially?

tgeoghegan:
> Q (just curious): why doesn't this target `main` initially?

The branch this PR is from contains the commit from #1659, and so if I made it target `main`, the PR would contain an extra commit and I wouldn't get a nice auto-generated PR description. Thinking about it further now, prod and staging deploy changes should merge cleanly even if merged out of order so I probably shouldn't have bothered to get clever with it.

PR #1703, opened 2022-05-27
Spot VM revamp

divergentdave:
This creates a separate GKE node pool with spot VMs, and steers different workloads to spot or non-spot nodes on both GKE and EKS. Kubernetes system services, metrics services, and CronJobs will be scheduled on non-spot nodes, while facilitator deployments may be scheduled on spot nodes. The spot nodes are assigned a NoSchedule taint, and facilitator deployments are given a toleration for that taint, plus a node affinity for spot VM/spot instance labels.

I decided against scheduling workflow-manager on spot nodes for now, as it takes several minutes to run, and we alert whenever it doesn't succeed. If we like the direction of this change, we may want to cut it into two halves to be deployed separately. Scale-in of the non-spot GKE node pool to one node per zone could be hairy if done concurrently with the other changes.


BranLwyd (approved):
I don't have a good intuitive sense for how reliable the spot instances are on different cloud providers, but I'm OK trying this out -- I suspect it'll work out just fine, and if it doesn't, rollback will be relatively painless.

divergentdave:
FYI I plan to merge this today, and roll it out this week.
Regarding node pool scale-in, our non-spot GKE node pools are starting with four nodes, and ending with three, so the controllers should be able to handle the change well.

PR #1720, opened 2022-05-31
Deploy 0.6.40 to production

tgeoghegan:


BranLwyd (approved):
Not sure why, but this PR currently contains the version bump to the staging environments as well as the prod environments. Approving anyway on the thought that the staging-only PR will be merged before this one, so it doesn't matter too much.

tgeoghegan:
Yeah, that's perplexing, but AFAICT the right sequence of commits has landed on main.

PR #1739, opened 2022-06-08
Update (indirect) generic-array dependency 0.12.3 -> 0.12.4.

branlwyd:
Dependabot reported a "high"-severity security bug in prio-server due to
this dependency: https://github.com/divviup/prio-server/security/dependabot/4

Commit generated via `cargo update -p generic-array:0.12.3`.

### DIFF
--- a/facilitator/Cargo.lock
+++ b/facilitator/Cargo.lock
@@ -245,7 +245,7 @@ dependencies = [
  "block-padding",
  "byte-tools",
  "byteorder",
- "generic-array 0.12.3",
+ "generic-array 0.12.4",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


tgeoghegan:
For posterity: @BranLwyd and I discussed this. We concluded that this dependency comes in via `warp`, which `prio-server` uses exclusively to expose a Prometheus metrics scrape endpoint, and thus no untrusted inputs are handled by the vulnerable code: the metrics scrape endpoints are only exposed as ClusterIPs and thus can't be accessed from outside the Kubernetes cluster (barring another vuln in GKE or VPC). With all that in mind, we concluded that we can release this in next week's scheduled deploy and don't need to fast-track it to prod.

PR #1817, opened 2022-07-08
workflow-manager: Avoid writing 0 to time gauges

divergentdave:
Our workflow-manager heartbeat alert has been firing after a single failed workflow-manager run, because this code path registered the gauge `workflow_manager_last_success_seconds`, never set its value, and sent the default value of 0 to prometheus-gateway. This PR fixes this issue by only registering success/failure gauges immediately before setting their values. It also fixes the `workflow_manager_last_failure_seconds` metric by pushing metrics after the gauge has been updated, and before `os.Exit()` is called.

I'm posting this as a draft early while I work on testing it.

### DIFF
--- a/terraform/modules/monitoring/prometheus_alerting_rules.yml
+++ b/terraform/modules/monitoring/prometheus_alerting_rules.yml
@@ -3,7 +3,7 @@ groups:
   rules:
   # See docs/monitoring.md for discussion of the PromQL expressions in these alerts.
   - alert: workflow_manager_heartbeat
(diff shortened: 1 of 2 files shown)


divergentdave:
I tested this out by running it in AWS and injecting failures with IAM policies (that deny s3:ListBucket with conditions on the request's prefix). The success path, fatal error path, and "Failed to schedule aggregation tasks" path all work as expected now. I revised the related alerting rule to work with the new metrics behavior.

tgeoghegan (approved):

PR #2088, opened 2022-10-07
Don't use postconditions on `data.http` resources

tgeoghegan:
Putting a `lifecycle.postcondition` on a `data.http` seems to force Terraform to defer reading that resource until the apply phase. This means that any resource that depends in turn on the `data.http` can't be planned, which leaves us unable to deploy new localities.

This commit deletes the `lifecycle` blocks, which lets us `plan` and hopefully `apply` again. We should still fail obviously if some JSON manifest can't be fetched, because parsing the JSON or accessing some field of the manifest will fail.


divergentdave (approved):
I confirmed that passing jsondecode bad data results in `Error: Error in function call` and a failure of the plan. (However, getting a nonexistent attribute from the resulting object results in `null`, not an error.) I think the results of all three data objects are covered well enough without the postconditions.

PR #2107, opened 2022-10-11
Reduce creation minimum age to 1w temporarily.

branlwyd:
This is intended to induce a key rotation. Once the key rotates once, we will revert this change.

The current situation: the key versions are (`old`, `bad`). Clients are submitting reports with both `old` & `bad` key versions. The `bad` key version cannot be used by Apple due to a technical issue. We cannot immediately delete either key version as we expect to receive reports from both. Our goal is to introduce a new key version without deleting either existing key version.

Overall plan:
* Submit this PR.
* Allow the key rotator to run. Since `bad` is more than 1w old, the key rotator will introduce a new key version `new`. As `old` was generated on 1/3 (i.e. 9 months ago), and `delete_min_age` is 13 months, no key versions will be deleted.
* Revert this PR within a week of the previous step. This will avoid the key rotator from introducing additional new versions.
* Continue operating with (`old`, `bad`, `new`) for the time being. On approximately 2/3/2023, the key rotator will remove `old` & we will return to operating with only two key versions.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -234,7 +234,7 @@ variable "packet_encryption_key_rotation_policy" {
     delete_min_count = number
   })
   default = {
-    create_min_age   = "6480h" // 6480 hours = 9 months (w/ 30-day months, 24-hour days)
+    create_min_age   = "168h" // 168 hours = 1 week (w/ 24-hour days)  // TODO: revert to 6480 hours = 9 months (w/ 30-day months, 24-hour days)
     primary_min_age  = "0"
     delete_min_age   = "9360h" // 9360 hours = 13 months (w/ 30-day months, 24-hour days)
     delete_min_count = 2


tgeoghegan (approved):
This plan in conjunction with #2108 seems good. What do you think about using `kubectl edit` in just one locality to set `-create-min-age=168h` along with `-dry-run` so that we can convince ourselves that applying this change in all localities will do what we want?

divergentdave (approved):
This plan looks good to me.

If we wanted to test with one locality first, `ta-ta` would be a good choice, as there's no data going through it. We could even drop the `--dry-run` and check that it behaves as expected running the second time.

branlwyd:
Indeed, testing is a great idea. `ta-ta` is a good testing ground for this -- per logs, looks like it rotated keys on 9/12. (I don't remember but I suspect we tested the key-rotator on this locality early.)

My testing steps were (all via GCP web UI):
* Edit `prod-us-key-rotator-ta-ta` to specify `--packet-encryption-key-create-min-age=168h`.
* Manually trigger a run.
* Verify that the run created a new key and deleted no keys.
* Manually trigger a run.
* Verify that the run did not create nor delete any keys.
* Edit `prod-us-key-rotator-ta-ta` to specify `--packet-encryption-key-create-min-age=6480h` once again.
* Manually trigger a run.
* Verify that the run did not create nor delete any keys.

Everything worked as expected. LMK if y'all have any other testing in mind, I'm going to reach out to do one final check that others are ready for the key to rotate.

PR #2110, opened 2022-10-13
Restore key creation minimum age to 6480h. (attempt #2)

branlwyd:
This is part of addressing the ENPA key rotation issue. See #2107 for discussion of the overall plan.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -234,7 +234,7 @@ variable "packet_encryption_key_rotation_policy" {
     delete_min_count = number
   })
   default = {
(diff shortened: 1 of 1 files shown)


branlwyd:
GH mismerged (?!) the previous PR (#2108). I suspect it did so since that PR had two commits, one of which was the 168h commit from #2107, followed by the 6480h commit; I purposefully set the PR up in that way to show that the second commit reverted the first. But GH apparently took this to mean that the commit was a no-op, even after #2107 was merged & #2108 was no longer a no-op.

divergentdave (approved):

PR #2111, opened 2022-10-13
Wait 23h before promoting packet encryption key to primary.

branlwyd:
Previously, we would promote a new packet encryption key to primary immediately, causing it to be advertised in our manifest as soon as it was created. Since our own jobs notice key updates on a 1h cadence, there was a small window of time where we were advertising a key that was not accepted by our jobs. We reasoned that it was very unlikely that end-user devices would receive a key update in less than one hour -- but this appears to be incorrect, based on our latest key rotation, in which a few devices appear to have started using the new key within 1h of it being introduced (leading to minor data loss).


tgeoghegan (approved):

divergentdave (approved):
LGTM. FWIW, I noticed that this will conflict with #2110.

PR #2182, opened 2022-11-07
Upgrade golangci-lint

divergentdave:
This updates golangci-lint to the latest version, reformats some comments, and swaps out some deprecated items for their non-deprecated aliases.

### DIFF
--- a/key-rotator/key/key.go
+++ b/key-rotator/key/key.go
@@ -201,14 +201,14 @@ func (cfg RotationConfig) Validate() error {
 // config, returning a new key (or the same key, if no rotation is necessary).
 //
 // Keys are rotated according to the following policy:
-//  * If no key versions exist, or if the youngest key version is older than
-//    `create_min_age`, create a new key version.
-//  * While there are more than `delete_min_key_count` keys, and the oldest key
-//    version is older than `delete_min_age`, delete the oldest key version.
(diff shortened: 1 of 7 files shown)


branlwyd (approved):

key-rotator/key/key.go
```diff
@@ -201,14 +201,14 @@ func (cfg RotationConfig) Validate() error {
...
-//      the youngest such key version as primary.
-//    * Otherwise, select the oldest key version as primary.
+//   - If no key versions exist, or if the youngest key version is older than
+//     `create_min_age`, create a new key version.
+//   - While there are more than `delete_min_key_count` keys, and the oldest key
+//     version is older than `delete_min_age`, delete the oldest key version.
+//   - Determine the current primary version:
+//   - If there is a key version not younger than `primary_min_age`, select
+//     the youngest such key version as primary.
+//   - Otherwise, select the oldest key version as primary.
```
divergentdave, line 211: This change happened because, per the latest `[doc comment documentation](https://tip.golang.org/doc/comment)`, "Gofmt reformats bullet lists to use a dash as the bullet marker, two spaces of indentation before the dash, and four spaces of indentation for continuation lines." Additionally, nested lists cannot be represented in doc comments, because "List items only contain paragraphs, not code blocks or nested lists.".

PR #2213, opened 2022-11-15
Deploy 0.6.73 to production envs

tgeoghegan:


branlwyd (approved):
(heads up that this is stacked on the staging PR, but that won't matter given the normal order of merges)

tgeoghegan:
> (heads up that this is stacked on the staging PR, but that won't matter given the normal order of merges)

Yeah, thanks for that reminder. It seems to me that you can either stack PRs as I did here and suffer the incorrect appearance of this PR including two commits or stack them by targeting PR A onto `main` and then PR B onto A, and then I inevitably forget to retarget PR B. So I choose the unaesthetic but less error-prone mode, awaiting the day that GitHub runs a managed Gerrit.

Issue #2248, opened 2022-12-02
Expedite key rotation in `prod-intl`


tgeoghegan:
Key rotation was enabled in the `prod-intl` environment by https://github.com/divviup/prio-server/pull/2246, deployed earlier today. Left as-is, `key-rotator` would run on Monday Dec 5 at 2000 UTC to generate keys, and then run again Tuesday Dec 6 at 2000 UTC to post new manifests. We've decided to expedite this process slightly in order to give ourselves a little more time to run the rest of the certificate issuance pipeline before the current certificates expire on Dec 9.

The plan is to manually run `key-rotator` today (Friday Dec 2) and then again on Sunday Dec 4. Then the MITRE cloud function will automatically pick up the new manifests.

The `key-rotator`s ran yesterday (Thursday Dec 1) in each of `prod-intl`'s six namespaces (`rw`, `mn`, `mx-coa`, `mx-pue`, `mx-yuc`, `mx-jal`) and I verified by inspecting the logs from those runs that the expected actions will be taken.

### Friday December 2: generate new packet encryption keys

Force each namespace's key rotator cron job to run immediately:
```
kubectl -n mn create job --from=cronjob/prod-intl-key-rotator-mn prod-intl-key-rotator-mn-manual-1
kubectl -n mx-coa create job --from=cronjob/prod-intl-key-rotator-mx-coa prod-intl-key-rotator-mx-coa-manual-1
kubectl -n mx-jal create job --from=cronjob/prod-intl-key-rotator-mx-jal prod-intl-key-rotator-mx-jal-manual-1
kubectl -n mx-yuc create job --from=cronjob/prod-intl-key-rotator-mx-yuc prod-intl-key-rotator-mx-yuc-manual-1
kubectl -n mx-pue create job --from=cronjob/prod-intl-key-rotator-mx-pue prod-intl-key-rotator-mx-pue-manual-1
kubectl -n rw create job --from=cronjob/prod-intl-key-rotator-rw prod-intl-key-rotator-rw-manual-1
```

Check logs of the pods that get created to verify that key rotation worked:

- batch signing keys are not generated for `mn`, `mx-coa`, `mx-pue`, `mx-yuc`, `mx-jal` (disabled by configuration)
- batch signing keys are generated for `rw` (no batch signing key exists yet)
- in each namespace, a new packet encryption keys is generated and written to Kubernetes secrets (the existing keys are sufficiently old OR do not exist in the case of `rw`)
[...]

PR #2265, opened 2022-12-05
Re-rotate keys in `mx-yuc, mx-coa, mx-pue`, part 2

tgeoghegan:
# Re-target this to main before merging!

Put key rotation configuration back to what it was:

 - enabled in all localities
 - rotating and promoting keys on default schedule

Part of #2248


terraform/variables/prod-intl.tfvars
```diff
@@ -120,14 +120,7 @@ default_peer_share_processor_manifest_base_url = "storage.googleapis.com/prio-en
 default_portal_server_manifest_base_url        = "manifest.global.enpa-pha.io"
 
 # Put all localities in dry-run mode
```
divergentdave, line 122: This comment got left behind

divergentdave (approved):
LGTM; BTW the title of this PR should be "part 3"

tgeoghegan:
> LGTM; BTW the title of this PR should be "part 3"

Yes, and I got the commit title wrong, too. Fixed in both places.

PR #2501, opened 2023-02-28
Update EKS plugins for 1.23 in prod

tgeoghegan:
This updates the EBS and VPC add-ons and the cluster autoscaler to recent ones compatible with 1.23. Because we are only doing a single minor version upgrade of the VPC CNI plugin, we don't need to perform any manual updates.


branlwyd (approved):

divergentdave (approved):

tgeoghegan:
Testing in staging uncovered one problem we need to fix before we can take this to production: https://github.com/divviup/prio-server/pull/2502

PR #2502, opened 2023-03-02
Update IAM policy for EKS cluster autoscaler

tgeoghegan:
The 1.23 Kubernetes cluster autoscaler needs one more permission in the AWS IAM policy granted to its role to function. The current documentation ([1]) also suggests restricting some of the existing permissions to only act on specific clusters, so we do that, too.

[1]: https://docs.aws.amazon.com/eks/latest/userguide/autoscaling.html


tgeoghegan:
The cluster autoscaler was unhealthy in staging EKS envs because of this problem. I manually applied a policy equivalent to what this does to staging-intl and it's now healthy.

divergentdave (approved):

PR #2607, opened 2023-03-28
Give key-rotator a toleration for spot VM taints

divergentdave:
This adds a toleration for `divviup.org/spot-vm` to the key rotation CronJob. This should hopefully help with the scheduling issues we've had over the past two days. Key rotation runs fast enough that using a spot instance should not be of concern.


branlwyd (approved):
(and even if a key rotation instance gets nuked partway through, it should be able to recover from any point -- the next (successful) run should clean things up properly.)

tgeoghegan (approved):

PR #2772, opened 2023-05-25
Add disclaimer explaining project is EOL

tgeoghegan:
Also disables dependabot since we no longer want to wrangle those PRs.


branlwyd:
(it looks like this PR doesn't disable Dependabot, but that's good -- we don't want to do that until the international environment turndown in late June.)

divergentdave (approved):

tgeoghegan:
> (it looks like this PR doesn't disable Dependabot, but that's good -- we don't want to do that until the international environment turndown in late June.)

It doesn't do that because I forgot to `git rm`! I was planning to not merge this until late June. I made the PR at this early date because I think we should loop in comms on this public communication.

tgeoghegan:
We'll merge this and also [archive the repository](https://docs.github.com/en/repositories/archiving-a-github-repository/archiving-repositories) in a few more weeks, once we stop supporting `prod-intl`.

PR #2821, opened 2023-06-06
Deploy 0.6.97 to prod-intl

divergentdave:


inahga (approved):
`prod-us` is dead, correct?

divergentdave:
Soon to be, we stopped deploying to it a week or two ago, and I'm planning on shutting things down later this week.

PR #2866, opened 2023-06-21
Bump prod to 0.6.99.

branlwyd:


terraform/variables/prod-us.tfvars
divergentdave, line 1: This environment is torn down already

divergentdave (approved):

tgeoghegan (approved):

terraform/variables/prod-us.tfvars
tgeoghegan: Yeah, let's not change this, because since the GCS bucket with TF state is now gone, this .tfvars file is the only record we have of what the final version of `prio-server` ever deployed to prod-us was (if anyone ever will care about that).

